PR #14, opened 2017-09-11
[SUCE-73]: Integrate api with front end I endpoint search result

TrebuhD:


client/elements/menus/sc-toolbar.html
```diff
@@ -149,68 +149,77 @@
...
+          _computeDiscourseLink(suttaPath, category) {
+              return (suttaPath) ? 'https://discourse.suttacentral.net/search?q="'
+                  + category + ' "' : "https://discourse.suttacentral.net/";
+          }
+
           // When looking-glass icon is clicked, determines if the searchbox is already open and if so, starts the search.
           // If not, it opens the search box and moves other elements out of the way depending on the width of the screen.
-          openSearch: function() {
+          openSearch() {
               let searchWidth;
```
gradam: Maybe define it right before usage? So in line ~230. It will be clearer

PR #20, opened 2017-09-13
[SUCE-110]: Enable endless scroll in search results view

TrebuhD:


client/elements/sc-page-search.html
```diff
@@ -273,43 +280,66 @@
...
+              };
+              this.shadowRoot.querySelector('#ajax').generateRequest();
           }
+
+          // This has some relatively expensive DOM querying, but at the moment is necessary because of this bug:
+          // https://github.com/PolymerElements/iron-list/issues/290.
+          shouldLoadMoreData() {
+              if (this.searchListNode.children.item(1)) {
+                  // update height of a single search result element (can differ by user device)
+                  this.searchResultElemHeight = this.searchListNode.children.item(1).clientHeight;
```
gradam: Can we calculate this only once? Maybe when first elements load

PR #25, opened 2017-09-18
[SUCE-148] live-reload and build process

gradam:


TrebuhD (approved):

client/package.json
```diff
@@ -0,0 +1,11 @@
+{
+  "name": "suttacentral",
+  "version": "1.0.0",
+  "scripts": {
+    "serv": "polymer serve -p 8000 --compile",
+    "dev": "npm run serv | npm run watch",
+    "test": "polymer test",
+    "build": "polymer build --preset es6-bundled --entrypoint index.html",
+    "watch": "browser-sync start --proxy 0.0.0.0:8000 --files \"**/*.*, index.html, *.js\" --no-open"
```
jhanarato, line 9: Replying to #3567:

This is the global dependency that is introduced but not used?

agilgur5: This script uses it - but it was removed and replaced in #1285. Since then, the dep was unused

jhanarato: OK, then the `watch` gets removed in 06f00a6.

PR #39, opened 2017-09-22
[SUCE-75]: Integrate API with Front End I: Endpoint - Parallel section

TrebuhD:


client/styles/sc-styles.html
```diff
@@ -39,6 +39,10 @@
       letter-spacing: 0.02em;
     };
 
+    --paper-font-common-base: {
+
+    };
```
gradam: Do we need this empty style?

PR #42, opened 2017-09-25
[SUCE-147] Generate HTMLs from PO data

gradam:


server/server/data_loader/po.py
```diff
@@ -1,60 +1,66 @@
...
     markup = []
     msgids = []
     msgstrs = []
-    
+
     po = polib.pofile(po_file)
-    
+
     for entry in po:
         ctx = entry.msgctxt
```
TrebuhD: is `ctx` used anywhere?

server/server/data_loader/po.py
```diff
@@ -194,3 +200,33 @@ def load_po_texts(change_tracker, po_dir, db):
...
                     collection = db.collection('po_strings')
                     insert_or_update(collection, doc)
+
+    generate_htmls(db)
+
+
+def generate_html(markup: str, translation: List[List[str]]) -> str:
+    translator = {string[0]: string[1] for string in translation}
+
+    pattern = re.compile('|'.join(re.escape(f'{key}">') for key in translator.keys()))
```
TrebuhD: it's not immediately obvious what's happening here, do you think we should add comments / docstrings?

server/server/migrations/migrations/add_po_htmls_collection_005.py
```diff
@@ -0,0 +1,16 @@
...
+
+class SecondMigration(Migration):
+    migration_id = 'add_po_htmls_collection_005'
+    tasks = ['create_collection']
+
+    def create_collection(self):
+        db = get_db()
+
+        po_htmls = db.create_collection('po_htmls')
+        po_htmls.add_hash_index(fields=['uid'], unique=False)
```
TrebuhD: shouldn't the uid be unique?

gradam: It is uid of sutta. we can have multiple translations of a single sutta.

TrebuhD: ah, I see.

server/server/data_loader/po.py
```diff
@@ -205,6 +203,22 @@ def load_po_texts(change_tracker, po_dir, db):
 
 
 def generate_html(markup: str, translation: List[List[str]]) -> str:
+    """ Takes HTML markup and insert translated string between sc-seg tags with id matching text uid.
```
TrebuhD: https://www.python.org/dev/peps/pep-0257/

`The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...".`

gradam: You have all that information here. It is Google's style doc-string

TrebuhD: You use both a descriptive style (`Takes`) and command style (`insert`) here. Is that Google style? :)

PR #45, opened 2017-09-26
[SUCE-50] dictionaries endpoint

gradam:


server/server/data_loader/dictionaries.py
```diff
@@ -0,0 +1,22 @@
+import json
+
+from tqdm import tqdm
+
+
+def load_dictionaries(db, dictionaries_dir):
+    print('Loading dictionaries')
+    dictionaries = dictionaries_dir.glob('*.json')
+    print(dictionaries)
```
TrebuhD: is this line necessary?

PR #46, opened 2017-09-26
Fix filtering endless scroll

TrebuhD:


client/elements/sc-page-search.html
```diff
@@ -322,19 +330,22 @@ <h2 class="search-result-title">[[item._source.heading.title]]</h2>
...
           }
 
           // Calls the input results when page is loaded
           _generateRequest() {
               if (!this.searchParams) {
                   return;
               }
+              if (this._areAllItemsLoaded()) {
+                  return;
+              }
```
gradam: can't it be one if with or ?

TrebuhD: done

PR #49, opened 2017-09-26
[SUCE-194] Suttaplex adjustments

gradam:


server/server/data_loader/dictionaries.py
```diff
@@ -12,10 +12,13 @@ def load_dictionaries(db, dictionaries_dir):
...
-            })
+            try:
+                from_lang, to_lang = dictionary.stem.split('2')
+                docs.append({
+                    'from': from_lang,
+                    'to': to_lang,
+                    'dictionary': json.load(f)
+                })
+            except ValueError:
+                pass
```
TrebuhD: I thought this was temporary?

server/server/data_loader/textdata.py
```diff
@@ -49,7 +49,7 @@ def uids_are_related(uid1, uid2, _rex=regex.compile(r'\p{alpha}*(?:-\d+)?')):
     def is_bold(self, lang, element):
         if element.tag in {'b', 'strong'}:
             return True
-        if lang in {'zh', 'lzh', 'ko', 'jp', 'tw'}:
+        if lang in {'lzh', 'lzh', 'ko', 'jp', 'tw'}:
```
TrebuhD: two `lzh`s?

server/server/data_loader/dictionaries.py
gradam: Exactly. Till the way we are supposed to hold dictionaries will be clear so it has to stay here for now

TrebuhD: Maybe at least print the `Exception`?

gradam: We could. A will add something

PR #63, opened 2017-10-05
[SUCE-225] parallel number badge

gradam:


server/server/common/queries.py
```diff
@@ -95,6 +95,12 @@
             LIMIT 1
             RETURN translation.title
     )[0]
+    
+    LET parallels_number = LENGTH(
```
TrebuhD: I'd call it either `parallel_count` or `number_of_parallels`.

PR #66, opened 2017-10-05
[SUCE-232] - Implement Redux and create frontend-builder container

TrebuhD:


client/elements/menus/language-base-menu.html
```diff
@@ -7,32 +7,33 @@
 <link rel="import" href="../../bower_components/neon-animation/web-animations.html">
 <link rel="import" href="../../bower_components/app-storage/app-localstorage/app-localstorage-document.html">
```
gradam: Those should be absolute paths. right  ?

client/redux-store.html
```diff
@@ -0,0 +1,38 @@
...
+
+<!--
+ This element creates and describes the redux store.
+ ---
+ Redux introduction: http://redux.js.org/docs/introduction/
+ Polymer-redux docs: https://tur-nr.github.io/polymer-redux/
+ -->
+<script>
+    const initialState = {
+        currentLocation: 'homepage',
```
gradam: What if the first page I visit is 'donation page' or anything other than homepage? For example from the link sent by someone to me.

client/redux-store.html
```diff
@@ -0,0 +1,38 @@
...
+        }
+    };
+
+    const store = Redux.createStore(
+        reducer,
+        initialState,
+        Redux.compose(
+            window.devToolsExtension
+                ? window.devToolsExtension()
+                : v => v
```
gradam: semicolon

client/redux-store.html
```diff
@@ -0,0 +1,38 @@
...
+        )
+    );
+
+    ReduxMixin = PolymerRedux(store);
+</script>
```
gradam: No newline at the end of the file

client/redux-store.html
TrebuhD: Nope, these are arguments

client/redux-store.html
TrebuhD: The basic process looks like this: 

the URL changes -> <sc-page-selector> detects this and updates the state -> the correct view is rendered.

For this case it's the same. The value is there to provide an idea of how the initial state tree would look, but it gets re-evaluated on every route change anyway.

client/elements/menus/language-base-menu.html
TrebuhD: correct!

PR #92, opened 2017-10-18
[SUCE-244]: Sutta View: textual information

TrebuhD:


client/elements/text/sc-simple-text.html
```diff
@@ -121,8 +129,11 @@
           // adds a class to the main container to either show or hide paragraph numbers.
           _addParagraphs() {
               const infotext = this.$.html_text_content;
-              return (infotext && !this.showParagraphs) ? infotext.classList.remove('infomode')
-                  : (infotext) ? infotext.classList.add('infomode') : '';
+              if (infotext && !this.showParagraphs) {
+                  infotext.classList.remove('infomode');
+              } else {
```
gradam: maybe don't combine if else and ?. Just use else if

server/server/api/views.py
```diff
@@ -452,6 +452,24 @@ def get(self):
...
+class Paragraphs(Resource):
+    def get(self):
+        """
+        Send list of textual information paragraphs for the sutta view
+        ---
+        responses:
+            200:
+                schema:
+                    id: paragraphs
+                    type: object
```
gradam: it is a bit poor schema.

server/server/common/queries.py
```diff
@@ -322,3 +322,8 @@
     SORT currency.name
     RETURN KEEP(currency, ['name', 'symbol', 'american_express', 'decimal'])
 '''
+
+PARAGRAPHS = '''
+FOR paragraph IN paragraphs
+    RETURN paragraph
+'''
```
gradam: no newline at end of file

server/server/api/views.py
```diff
@@ -452,6 +452,24 @@ def get(self):
...
+        """
+        db = get_db()
+
+        data = db.aql.execute(PARAGRAPHS)
+
+        return [{"uid": paragraph['uid'], "description": paragraph['description']} for paragraph in data], 200
```
gradam: it should be done in 'PARAGRAPHS' AQL

PR #103, opened 2017-10-20
Sutta view improvements

TrebuhD:


client/elements/addons/sc-badge.html
```diff
@@ -5,7 +5,7 @@
 -->
 <dom-module id="sc-badge">
   <template>
-    <style include="sc-badge-styles">
+    <style>
```
gradam: newline not needed

PR #109, opened 2017-10-23
Sutta view improvements

TrebuhD:


client/elements/text/sc-simple-text.html
```diff
@@ -91,21 +89,19 @@
...
                   },
                   loading: {
-                      type: Boolean
+                      type: Boolean,
+                      observer: '_loadingChanged'
                   },
                   error: {
                       type: Object
+                  },
+                  textViewHidden: {
```
gradam: maybe add some prefix like `isTextViewHidden`

client/elements/text/sc-simple-text.html
```diff
@@ -147,6 +164,12 @@
               this.dispatch('changeToolbarTitle', title);
           }
 
+          // returns the meta-data from the loaded sutta text
+          _computeMeta() {
+              const matches = this.sutta.text.match(/<aside id="metaarea">((.|\n)*)<\/aside>/);
+              return matches ? matches[1] : '';
```
gradam: What about the situation when we have only one item in matches?

client/elements/text/sc-text-page-selector.html
```diff
@@ -22,13 +32,18 @@
         last-error="{{error}}"
         last-response="{{responseData}}"></iron-ajax>
 
-    <template id="simple_text_container" is="dom-if" if="[[!isSegmentedText]]" restamp>
-      <sc-simple-text loading="{{loading}}" error="{{error}}" id="simple_text"></sc-simple-text>
-    </template>
+    <div class="loading-indicator" hidden$="[[!loading]]">
+      <paper-spinner-lite active="[[loading]]"></paper-spinner-lite>
```
gradam: `is` prefix for `loading`?

PR #111, opened 2017-10-23
[SUCE-245] remarks added to parallels

gradam:


client/elements/sc-donate-now-page.html
```diff
@@ -79,6 +79,7 @@
         background: var(--sc-gold-300);
         color: white;
         font-weight: bold;
+
```
TrebuhD: no reason for this line

PR #113, opened 2017-10-23
Settings dialog & paragraphs improvements

TrebuhD:


client/elements/sc-navdrawer.html
```diff
@@ -112,6 +112,11 @@
         color: var(--disabled-text-color);
       }
 
+      .menu-header {
+        @apply --paper-font-title;
+        @apply --sc-sans-font;
+        margin: 0
```
gradam: ;

client/elements/text/sc-segmented-text.html
```diff
@@ -274,42 +272,44 @@
               }
               let metaText;
               try {
-                  metaText = matches[1];
-              } catch(e) {
+                  if (matches && matches.length > 0) {
```
gradam: no need for second condition

PR #114, opened 2017-10-24
Suce 242 donations enhancements

gradam:


client/elements/sc-donate-now-page.html
```diff
@@ -215,11 +217,27 @@
       }
 
       .error-toast {
-        --paper-toast-background-color: red;
+        --paper-toast-background-color: #323232;
```
TrebuhD: we should extract this color to a mixin

client/elements/sc-donate-now-page.html
```diff
@@ -215,11 +217,27 @@
...
       }
 
       .success-toast {
-        --paper-toast-background-color: green;
+        --paper-toast-background-color: #323232;
```
TrebuhD: ditto

PR #136, opened 2017-10-31
[SUCE-268] extra padding on bottom of suttaplex

gradam:


client/elements/suttaplex/sc-suttaplex.html
```diff
@@ -72,7 +72,7 @@
       }
 
       paper-card.suttaplex .card-actions {
-        padding: 0 var(--size-md);
+        padding: 0 var(--size-md) 8px;
```
TrebuhD: 8px = `var(--size-sm)`

PR #138, opened 2017-10-31
Suce 264 dropdown should disapear

gradam:


client/elements/menus/sc-more-menu.html
```diff
@@ -99,6 +99,34 @@
...
+                e.addEventListener('click', (e) => {
+                    this._dispatchItemSelectedEvent();
+                });
+            });
+            this.$.language_menu.addEventListener('iron-select', (e) => {
+              if(this.created){
+                  this._dispatchItemSelectedEvent();
+              };
+              this.created = true;
+            });
```
TrebuhD: I'd create a separate function for adding each of these listeners.

PR #141, opened 2017-11-02
fit paragraphs on laptop screen

TrebuhD:


client/styles/sc-text-styles.html
```diff
@@ -16,7 +16,6 @@
 
       .html-text-content.side-by-side section {
         max-width: 80em;
-        padding: 32px 4% 32px 8%;
```
gradam: I liked that line.
Are you happy?

PR #142, opened 2017-11-02
[SUCE-131] css fixes

gradam:


client/elements/static/downloads-page.html
```diff
@@ -5,10 +5,6 @@
 <dom-module id="downloads-page">
   <template>
     <style is="custom-style" include="static-styles">
```
TrebuhD: `is="custom-style"` shouldn't be here

PR #147, opened 2017-11-02
[SUCE-295] vert menu bug

gradam:


client/elements/menus/more-par-menu.html
```diff
@@ -177,16 +177,26 @@
...
+                    }
+                    return 'swap-horiz';
+                case 'retelling':
+                    return 'cached';
+                case 'mention':
+                    return 'editor:format-quote';
+                default:
+                    return '';
+            }
               return '';
```
TrebuhD: is this line needed?

client/elements/menus/more-par-menu.html
```diff
@@ -177,16 +177,26 @@
               }
           }
 
-          _computeIconForType(type) {
-              if (type === 'parallels') return 'swap-horiz';
-              if (type === 'retells') return 'cached';
-              if (type === 'mentions') return 'editor:format-quote';
+          _computeIconForType(parallel) {
```
TrebuhD: I'm thinking now: is this function even needed? It's only used in `_computeIcon`, so why not put all the logic there?

PR #148, opened 2017-11-06
temporary disappearing text workaround

TrebuhD:


client/styles/sc-text-styles.html
```diff
@@ -32,6 +32,52 @@
...
+      }
+
+      .text-tooltip {
+        @apply --sc-sans-font;
+        @apply --shadow-elevation-6dp;
+        --paper-tooltip-opacity: 0.98;
+        --paper-tooltip-background: var(--paper-grey-700);
+        --paper-tooltip: {
+          @apply --sc-skolar-font-size-xs;
+          line-height: 16px;
```
gradam: Don't we have variable for that?

client/styles/sc-text-styles.html
```diff
@@ -32,6 +32,52 @@
...
+          padding: 8px var(--size-md);
```
gradam: Ditto

PR #149, opened 2017-11-06
Fix paragraphs scroll

TrebuhD:


client/elements/text/sc-segmented-text.html
```diff
@@ -311,17 +311,21 @@
...
-                  });
+                  this._addParagraphScrollListeners();
                   this.assembledMainText = this.$.segmented_text_content.innerHTML;
               }
               this._setParagraphsVisible(this.showParagraphs);
           }
 
+          _addParagraphScrollListeners() {
+              this.paragraphs.forEach((paragraph) => {
+                  const refs = this.$.segmented_text_content.querySelectorAll('.' + paragraph.uid);
```
gradam: Shouldn't we use string literals?

TrebuhD: that's correct

PR #152, opened 2017-11-08
Parallels fix: when there's no title, display acronym, don't repeat in nerdy row

TrebuhD:


client/elements/suttaplex/sc-parallel-item.html
```diff
@@ -211,7 +213,8 @@
           // returns the relevant header, either the title in the translated language,
           // otherwise in pali or otherwise just the ID.
           _computeTitle() {
-              return this.parallelItem.original_title ? this.parallelItem.original_title : this.parallelItem.uid;
+              return this.parallelItem.original_title ? this.parallelItem.original_title :
+                  this.parallelItem.acronym ? this.parallelItem.acronym : this.parallelItem.uid;
```
gradam: I am still not sure if nested inline ifs is a good idea

PR #179, opened 2017-11-20
Textual information fixes

TrebuhD:


client/elements/text/sc-segmented-text.html
```diff
@@ -432,6 +443,18 @@ <h3>Could not find the sutta text.</h3>
               }
           }
 
+          _scrollToSectionInUrl() {
+              const w = window.innerWidth;
```
gradam: Why extract that to a variable?

PR #195, opened 2017-11-27
[SUCE-317] Segmented ids: Check if all works by default

TrebuhD:


client/elements/text/sc-segmented-text.html
```diff
@@ -192,6 +192,10 @@ <h3>Could not find the sutta text.</h3>
                   segmentedIDsAdded: {
                       type: Boolean,
                       value: false
+                  },
+                  scriptChanged: {
```
gradam: The name sounds more like an event. Maybe change it to something like 'hasScriptChanged'?

client/elements/text/sc-segmented-text.html
```diff
@@ -294,18 +297,18 @@ <h3>Could not find the sutta text.</h3>
               switch (this.chosenTextView) {
                   case 'sidebyside':
```
gradam: Content of 2 first cases can be extracted to a function with argument

client/elements/text/sc-segmented-text.html
```diff
@@ -514,30 +518,46 @@ <h3>Could not find the sutta text.</h3>
...
+                  .querySelectorAll('.segmented-textual-info-paragraph.sc:not(.textual-info-paragraph-inline)');
+              let segments = this.translatedSutta ? this.shadowRoot.querySelectorAll('sc-seg.translated-text') :
+                  this.shadowRoot.querySelectorAll('sc-seg.original-text');
+              // store the segment offsets in an object to not query the elements in a loop which would force layout redraw:
+              const segmentOffsets = this.getSegmentTopOffsets(segments);
+              Array.from(segmentedParagraphs).forEach((paragraph) => {
+                  const offsets = segmentOffsets[paragraph.getAttribute('data-reference-segment-id')];
+                  let paragraphMargin = 0;
+                  if (paragraph.previousElementSibling &&
+                      paragraph.previousElementSibling.classList.contains('segmented-textual-info-paragraph')) {
```
gradam: Maybe extract it to const? So that if fits into one line.

PR #212, opened 2017-11-30
[SUC-304] updated titles to toggle if class is visible

SamaneriVimala:
Changed title attribute to toggle on and off dependent on whether Textual Information is on or off.
Other titles are on all the time if the css shows that the specific class is visible when TI is off.

### DIFF
--- a/client/elements/text/sc-simple-text.html
+++ b/client/elements/text/sc-simple-text.html
@@ -133,7 +133,7 @@
                       statePath: 'textOptions.chineseLookupTargetDictRepr',
                       observer: '_chineseLookupTargetDictReprChanged'
                   },
-                  classTitlesTI: {
+                  textualInfoClassTitles: {
                       type: Object,
                       value: {
                           "gloss":"Definition of term.",
@@ -288,9 +288,13 @@
                       const classrefs = this.$.simple_text_content.querySelectorAll('.'+key);
                       Array.from(classrefs).forEach(item => item.title = this.classTitles[key]);
                   }
-                  for (var key in this.classTitlesTI){
-                      const classrefsTI = this.$.simple_text_content.querySelectorAll('.'+key);
-                      visible ? Array.from(classrefsTI).forEach(item => item.title = this.classTitlesTI[key]) : Array.from(classrefsTI).forEach(item => item.title = "");
+                  for (var key in this.textualInfoClassTitles){
+                      const textualInfoclassrefs = this.$.simple_text_content.querySelectorAll(`.${key}`);
+                      if (visible) {
+                          Array.from(textualInfoclassrefs).forEach(item => item.title = this.textualInfoClassTitles[key]);
+                      } else {
+                          Array.from(textualInfoclassrefs).forEach(item => item.title = "");
+                      }
                   }
               }
           }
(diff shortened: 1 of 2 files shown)


client/elements/text/sc-simple-text.html
```diff
@@ -278,6 +288,10 @@ <h3>Could not find the sutta text.</h3>
                       const classrefs = this.$.simple_text_content.querySelectorAll('.'+key);
                       Array.from(classrefs).forEach(item => item.title = this.classTitles[key]);
                   }
+                  for (var key in this.classTitlesTI){
+                      const classrefsTI = this.$.simple_text_content.querySelectorAll('.'+key);
+                      visible ? Array.from(classrefsTI).forEach(item => item.title = this.classTitlesTI[key]) : Array.from(classrefsTI).forEach(item => item.title = "");
```
gradam: Line to long

client/elements/text/sc-simple-text.html
```diff
@@ -278,6 +288,10 @@ <h3>Could not find the sutta text.</h3>
                       const classrefs = this.$.simple_text_content.querySelectorAll('.'+key);
                       Array.from(classrefs).forEach(item => item.title = this.classTitles[key]);
                   }
+                  for (var key in this.classTitlesTI){
+                      const classrefsTI = this.$.simple_text_content.querySelectorAll('.'+key);
```
gradam: Please use string literals

client/elements/text/sc-simple-text.html
```diff
@@ -133,18 +133,28 @@ <h3>Could not find the sutta text.</h3>
                       statePath: 'textOptions.chineseLookupTargetDictRepr',
                       observer: '_chineseLookupTargetDictReprChanged'
                   },
-                  classTitles: {
+                  classTitlesTI: {
```
TrebuhD: I'd say don't worry about long names, it matters most if they're descriptive. I'd suggest naming this `textualInfoClassTitles` or `textualInfoClassDescriptions`.

client/elements/text/sc-simple-text.html
TrebuhD: I'd advise keeping complicated logic out of ternary operators (`? :`) - it's just harder to read. I'd rewrite this as:
```
if (visible) {
    Array.from(classrefsTI).forEach(item => item.title = this.classTitlesTI[key])
} else {
    Array.from(classrefsTI).forEach(item => item.title = "");
}
```

[Here's](https://github.com/airbnb/javascript#comparison--nested-ternaries) the relevant part of the Airbnb JS style guide.

SamaneriVimala: OK. Good. Will change that. I actually had that at first but changed it.

client/elements/text/sc-simple-text.html
SamaneriVimala: OK. Will change that.

client/elements/text/sc-simple-text.html
gradam: Please use single quotes instead of double quotes.

SamaneriVimala:
I've made the changes and pushed them to the branch. Now what?

PR #249, opened 2017-12-10
[SUCE-359] update css text classes

SamaneriVimala:


client/styles/sc-text-paragraph-num-styles.html
```diff
@@ -437,6 +437,10 @@
         display: none;
       }
 
+      #text:lang(si) a.msdiv {
```
TrebuhD: Keep in mind that complicated CSS selectors like this one can hurt performance when the browser is calculating the layout. In this case, the browser has to:
- select all `a` elements in the DOM
- filter out the ones that don't have the `msdiv` class
- for every one of them, check if they have a parent somewhere with `id="text"` and `lang="si"`.

The performance hit is small at first, but it can accumulate over time when using too many selectors like this.

BTW, is the `msdiv` class present on elements other than `a`? If not, you could eliminate the `a` selector or replace it with a more specific class selector (like `.textual-info-paragraph.msdiv`).

SamaneriVimala: The problem is that we have msdiv class on `a`-tags in several files but it has to perform differently on the files where the language is `si` than on the others. On the `si`-files it has to behave as normal paragraph (off when TI is off, showing when on), while on all other files it has to be `display:none`. 

I suppose the best thing to do then is to split it indeed into different classes in the html text-files themselves as you suggest. I will do a regex over that and add them.

TrebuhD: OK. If it's too much work, don't bother. It's your call, but I'm going to approve this PR.

SamaneriVimala: Not too much work, but my Sublime just froze after trying to make the change on 5000+ files and over 25000 entries! :-)
Will have to get back to it later because have to go now.

SamaneriVimala:
I've updated it now but ... it does not show. I also added `msdiv-si` to the parallels.json file and changed all the references in the html files. Can you tell me why this is not picked up by the system (also for future new references to be added).

PR #253, opened 2017-12-11
[SUCE-173] home page

gradam:


client/elements/static/home-page.html
```diff
@@ -1,56 +1,490 @@
...
+
+        .card-heading {
+          width: 25%;
+        }
+
+        .new-line {
+          display: unset;
+        }
+
+        .three-baskets-grid {
```
TrebuhD: I'd rename all the `three-baskets` classes so that they're more general.

PR #261, opened 2017-12-15
Suce 385 refactor for clarity around divisions

blake-sc:


server/server/common/queries.py
```diff
@@ -20,31 +20,30 @@
...
+            FILTER d_edge.type != 'text'
+            RETURN {
+                from: d_edge._from,
+                name: d.name,
+                uid: d._id,
+                num: d.num
+            }
+        )
+    
+    RETURN MERGE({uid: div._id, name: div.name, num: div.num}, {descendents: descendents, parents: parents})
```
gradam: Why do we need MERE here? Can't we just use 
`{uid: div._id, name: div.name, num: div.num, descendents: descendents, parents: parents}`?

PR #269, opened 2017-12-18
fixed homepage link (where to begin)

gradam:


client/elements/static/home-page.html
```diff
@@ -46,7 +46,7 @@
       }
 
       .card-heading {
-        @apply --sc-skolar-font-size-xxl;
+        @apply --sc-skolar-font-size-xxl;Discourses
```
TrebuhD: wut

PR #278, opened 2017-12-20
[SUCE-344] Numbering at some menu levels

TrebuhD:


server/server/api/views.py
```diff
@@ -49,6 +49,10 @@ def get(self):
 
 
 class Menu(Resource):
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self.r = re.compile(r'^[^\d]*?([\d]+)$')
```
gradam: please use more explanatory name

server/server/api/views.py
```diff
@@ -145,6 +149,8 @@ def recursive_cleanup(self, menu_entries, mapping):
         menu_entries.sort(key=self.num_sort_key)
         for menu_entry in menu_entries:
             mapping[menu_entry['uid']] = menu_entry
+            if 'id' in menu_entry and self.r.match(menu_entry['id']):
+                menu_entry['display_num'] = self.r.match(menu_entry['id'])[1]
```
gradam: DRY. Don't call `r.match` twice on the same string.

PR #279, opened 2017-12-20
[SUCE-397 ]Dictionary page

SamaneriVimala:


server/server/api/views.py
```diff
@@ -564,13 +564,28 @@ def get(self):
...
+                properties:
+                    dictname:
+                        type: string
+                    word:
+                        type: string
+                    text:
+                        type: string
 
         """
+        if (word != None):
```
gradam: `is not` should be used instead of `!=`
https://www.python.org/dev/peps/pep-0008/#programming-recommendations

SamaneriVimala: You're right. Don't ask me to do anything at 5 am ... my brain is not there and I mix up JS & python. Will change it. Thanks!

PR #280, opened 2017-12-20
Suce 296 paragraph ranges

gradam:


client/elements/text/sc-segmented-text.html
```diff
@@ -594,6 +591,11 @@ <h3>Could not find the sutta text.</h3>
           }
 
           _processLastSection(section, wrapper, idFrom, idTo, isSideBySideView, wrapperAlreadyExists) {
+            const paddingBottom = window.getComputedStyle(section).paddingBottom;
```
TrebuhD: there's something off with the indentation here.

client/elements/text/sc-simple-text.html
```diff
@@ -405,6 +401,27 @@ <h3>Could not find the sutta text.</h3>
               return firstSection;
           }
 
+          _paddingToMargin(section) {
```
TrebuhD: these function names are missing verbs. "change" would be fine.

client/elements/text/sc-simple-text.html
```diff
@@ -405,6 +401,27 @@ <h3>Could not find the sutta text.</h3>
...
+            const margin = window.getComputedStyle(section).padding;
+            requestAnimationFrame(() => {
+                section.style.margin = margin;
+                section.style.padding = '0';
+            })
+          }
+
+          _marginToPadding(section) {
+            for(let pos of ['Top', 'Bottom', 'Left', "Right"]) {
```
TrebuhD: making consecutive reads and writes to `style` inside a `for` loop is really bad practice, as the browser needs to rebuild the layout before each read. This loop only runs 4 times, so it's borderline acceptable. 

It would still be much better if this was split into two loops. In the first one you'd read the margins/paddings into an object and then write to the DOM using the values already in that object.

client/elements/text/sc-simple-text.html
```diff
@@ -381,20 +381,16 @@ <h3>Could not find the sutta text.</h3>
               }
               while (section) {
                   section.classList.add('highlight');
-                  const margin = window.getComputedStyle(section).margin;
-                  requestAnimationFrame(() => {
-                      section.style.margin = '0';
-                      section.style.padding = margin;
-                  })
+                  this._marginToPadding(section)
```
TrebuhD: I think you forgot to refactor here

client/elements/text/sc-simple-text.html
```diff
@@ -381,20 +381,16 @@ <h3>Could not find the sutta text.</h3>
...
                   let nextSection = section.nextElementSibling;
+                  if (!nextSection) {
+                      this._paddingToMargin(section);
```
TrebuhD: ditto

client/elements/text/sc-simple-text.html
```diff
@@ -381,20 +381,16 @@ <h3>Could not find the sutta text.</h3>
...
+                  }
                   let sectionLink = nextSection.querySelector('.textual-info-paragraph');
                   if (sectionLink) {
                       if (last) {
-                          const marginBottom = window.getComputedStyle(section).paddingBottom;
-                          requestAnimationFrame(() => {
-                              section.style.marginBottom = marginBottom;
-                              section.style.paddingBottom = '0';
-                          })
+                          this._paddingToMargin(section.nodeName === "A" ? section.previousElementSibling : section);
```
TrebuhD: ditto

PR #288, opened 2017-12-21
[SUCE-387] loading glossary info to ArangoDB

SamaneriVimala:


server/server/app.py
```diff
@@ -34,6 +34,7 @@ def app_factory() -> Tuple[Api, Flask]:
     api.add_resource(Images, '/images/<string:division>/<int:vol>')
     api.add_resource(Epigraphs, '/epigraphs')
     api.add_resource(WhyWeRead, '/whyweread')
+    api.add_resource(Glossary, '/glossary/<path:glossword>')
```
gradam: Why is `glossword` a path and not a string?

server/server/common/queries.py
```diff
@@ -460,3 +460,11 @@
     LIMIT @number
     RETURN text.text
 '''
+
+GLOSSARY = '''
+FOR gloss IN glossary
+    RETURN {
+        glossword: gloss.glossword,
+        description: gloss.description
+    }
```
gradam: this is one way to do it. You can also use `KEEP(gloss, ['glossword', 'description'])` both are fine. Just as a tidbit

server/server/data_loader/glossary.py
```diff
@@ -0,0 +1,11 @@
+import json
+from pathlib import Path
+
+
+def load_glossary(db, additional_info_dir: Path):
+    print('Loading glossary')
+
+    glossary_collection = db['glossary']
+    with open(f'{additional_info_dir}/gloss.json') as f:
```
gradam: you could also do `with (additional_info_dir / 'gloss.json').open() as f`. Again just as a tidbit

PR #295, opened 2017-12-22
[SUCE-394] change copylink icon

SamaneriVimala:


client/elements/menus/more-par-menu.html
```diff
@@ -50,7 +50,7 @@
                last-response="{{parallels}}"></iron-ajax>
 
     <paper-item class="table-element button-text" on-tap="_copyLink" title="[[_computeLink(item)]]">
-      <iron-icon class="grey-icon" icon="link"></iron-icon>
+      <iron-icon class="grey-icon" icon="social:share"></iron-icon>
```
TrebuhD: using icons with "social" in the name has an unfortunate side-effect - the icons get blocked by adblock and are not displayed at all. Previously we copied those icons from source and into `sc-svg-icons` to solve this.

PR #297, opened 2017-12-22
[SUCE-381] toast notification on text pages

SamaneriVimala:


client/elements/page-selector.html
```diff
@@ -87,6 +87,22 @@
...
+      .toast {
+        @apply --sc-skolar-font-size-xl;
+        @apply --sc-sans-font;
+        --paper-toast-color: var(--sc-tertiary-text-color);
+        text-align: center;
+        margin-left: calc(50vw - 150px);
+      }
+
+      .success-toast {
+        --paper-toast-background-color: var(--sc-primary-accent-color);
```
TrebuhD: `--sc-toast-success-color`. The primary color can change, but the toast should stay green.

client/elements/page-selector.html
```diff
@@ -229,6 +251,19 @@
...
+              this.addEventListener('par-menu-copied', (e) => {
+                  this.toastMessage = e.detail.message;
+                  const success = e.detail.success;
+                  if (success) {
+                      this.$.success_toast.open();
+                  } else {
+                      this.$.error_toast.open();
+                  }
+              });
+          }
```
TrebuhD: Toasts are now displayed by sending a `show-sc-toast` event to the `sc-toasts.html` file. If you use that, you won't need to define any custom styles here. Check the `sc-donate-now-page` file for an example.

client/elements/page-selector.html
```diff
@@ -87,6 +87,22 @@
         }
       }
 
+      .toast {
```
TrebuhD: We don't need these styles here. I'll explain below.

client/elements/page-selector.html
SamaneriVimala: I actually did not change anything to the code here: I just moved the code that was there from the sc-view-suttaplex to page-selector. That's all. The reason is that the toast for suttaplex cards was unable to show for pages other than in the sc-view-suttaplex like for instance the text pages. By moving it to higher up in the chain it could show up for any page.
Sure I can change the entire code, but considering that it is not my code in the first place I feel a little reluctant to do that.

TrebuhD: Oh, I see. That means we're still using the old style somewhere (that's an overshight, sorry).

PR #301, opened 2017-12-24
force opening dropdowns downwards

TrebuhD:


SamaneriVimala:
Your solution makes indeed more sense!

PR #306, opened 2017-12-28
Hamburger menu fixes

TrebuhD:


client/elements/page-selector.html
```diff
@@ -189,6 +194,10 @@
                   shouldShowSearchPage: Boolean,
                   shouldShowSuttaTextPage: Boolean,
                   shouldShowDictionaryPage: Boolean,
+                  isDrawerOpen: {
+                      type: Boolean,
+                      observer: '_drawerOpenStateChanged'
```
gradam: Can't we use notify?

client/elements/page-selector.html
```diff
@@ -234,14 +243,21 @@
               this.addEventListener('par-menu-copied', (e) => {
                   let success = '';
                   if (e.detail.success) {
```
gradam: I think the variable name should be something more like result. Success suggests binary value rather than string value

TrebuhD: I think this is your code

PR #308, opened 2017-12-28
[SUCE-355]: Language automatical set up per browser settings

TrebuhD:


client/elements/menus/language-base-menu.html
```diff
@@ -106,17 +108,13 @@
                       type: Array,
                       value: () => { return []; }
                   },
-                  selectedLanguageNr: {
+                  selectedLanguageNum: {
                       type: Number,
-                      observer: '_selectedLanguageNrChanged'
+                      observer: '_selectedLanguageNumChanged',
```
gradam: Can't we just use notify: true?

TrebuhD: It's the same thing IMHO.

PR #313, opened 2017-12-29
[SUCE-429] added bibliography static page

SamaneriVimala:


client/elements/static/bibliography-page.html
```diff
@@ -0,0 +1,49 @@
...
+        @apply --sc-mixed-small-caps;
+      }
+    </style>
+
+    <iron-ajax
+        auto
+        id="ajax"
+        url="/api/bibliography"
+        handle-as="json"
+        last-response="{{bibliographyReturns}}"></iron-ajax>
```
TrebuhD: the typical name convention we adopted would be `bibliographyResponse`

PR #315, opened 2017-12-30
[SUCE-386] toasts not appearing

SamaneriVimala:
This is just one solution to the problem of toasts not appearing unless the button is actually clicked, but it might not be the best use of the Redux Store ... let me know if you have other suggestions.


TrebuhD:
A few notes here:
1. You shouldn't need to access the redux store from localstorage at any time. If you read the [polymer-redux](https://tur-nr.github.io/polymer-redux/) documentation, there's an easy way to do get values from the Redux store - using the `statePath` property. In fact, you already were using it correctly [in a previous PR](https://github.com/suttacentral/next-sc/pull/263/files)!

Getting the state from localstorage is incorrect and will lead to problems - please make sure to always use `statePath` to get the current state of the Redux store.

2. IMHO redux doesn't need to be involved in this. If all we want to do is display a toast when a dropdown value is changed, we can check out the [documentation](https://www.webcomponents.org/element/PolymerElements/paper-dropdown-menu) for `paper-dropdown-menu` and see that there's a `selected-item-changed` event emitted by that element. We can simply listen for that event and show a toast at that time. I haven't tested this, but IMHO it should work.

SamaneriVimala:
Eh, well no, I did not use it before. That was a copy from something in a previous version but the statepath was not used there. But thanks for the feedback and I will change it and upload it again in the next few hours.

client/elements/menus/settings-menu.html
```diff
@@ -385,6 +393,12 @@ <h3 class="nerdy-row">Select this to remember these settings next time you visit
               this.shadowRoot.querySelector('#chinese_lookup_menu').addEventListener('value-changed', (e) => {
                   this._chineseLookupChanged(e.detail.value);
               });
+              this.shadowRoot.querySelector('#pali_lookup_menu').addEventListener('selected-item-changed', () => {
+                  this.paliSelectedItemChanged = true;
```
TrebuhD: This is much better, but I'm thinking - couldn't we just call `this._showToast(dictName)` here and get rid of the additional properties? I think `dictName` can be retrieved from the passed event, like this:

```
this.shadowRoot.querySelector('#chinese_lookup_menu').addEventListener('selected-item-changed', (**event**) => {
    console.dir(event); // The dictionary name should be somewhere in this object.
    this.chineseSelectedItemChanged = true;
});
```

SamaneriVimala: I suppose we could. The Dictname has to be converted for the showToast but that's only a minor string conversion. Will get on it and upload soon.

SamaneriVimala: ?? You said :+1: but you dit not approve the pull request ???
So I'm not sure if I was clear or not: the above code (calling the function directly at the EventListener) did not work. Therefore I did not push it to the branch.

TrebuhD: Sorry. Approving now.

PR #319, opened 2018-01-01
[SUCE-377] donations data clean up

SamaneriVimala:


client/elements/sc-donate-now-page.html
```diff
@@ -485,6 +484,16 @@
...
           }
 
+          resetForm() {
+              this.$.donation_form.reset();
+              this.shadowRoot.querySelector('.amount-input').value = undefined;
+              this.shadowRoot.querySelector('.one-time-donation').checked = true;
+              this.shadowRoot.querySelector('.monthly-donation').checked = false;
+              this.shadowRoot.querySelector('.name-input').value = undefined;
+              this.shadowRoot.querySelector('.email-input').value = undefined;
+              this.shadowRoot.querySelector('.message-input').value = undefined;
```
TrebuhD: Are all these values not set by calling `this.$.donation_form.reset();`?

client/elements/sc-page-static.html
```diff
@@ -283,6 +283,7 @@ <h3>{{localize('pageNotFound')}}</h3>
       }
 
       _changeView() {
+          this.$.donate.resetForm();
```
TrebuhD: Will this get called when going from a static page to a non-static page (the suttaplex for example)?

If it doesn't get called, we can also put this function call it a `disconnectedCallback()`.

client/elements/sc-donate-now-page.html
SamaneriVimala: No, that only resets the fields for the amount and the credit card details but not the rest.

client/elements/sc-page-static.html
SamaneriVimala: I tested it and it works fine going from the static to the non-static page. I think that was never a problem in the first place anyway.

PR #324, opened 2018-01-02
Suce 296 paragraph ranges

gradam:


client/elements/text/sc-segmented-text.html
```diff
@@ -664,15 +664,26 @@ <h3>Could not find the sutta text.</h3>
```
TrebuhD: redundant blank lines

client/elements/text/sc-segmented-text.html
```diff
@@ -638,7 +638,7 @@ <h3>Could not find the sutta text.</h3>
               return newSection;
           }
 
-          _processHighlightAndScroll(sectionId, isSmooth) {
+          _processHighlightAndScroll(sectionId) {
```
TrebuhD: this function is way too long. Please split it into more manageable parts.

PR #328, opened 2018-01-03
[SUCE-309] remove italics from texts with different scripts

SamaneriVimala:
According to Sujato's instructions


client/styles/sc-text-styles.html
```diff
@@ -6,6 +6,24 @@
         text-decoration: none;
       }
 
+      *:lang(ar),
```
TrebuhD: 1. Is the star selector necessary here?
2. I measured CSS selector performance [here](http://stevesouders.com/efws/css-selectors/csscreate.php?n=1000&sel=%3Alang%28fr%29&body=background%3A+%23CFD&ne=1000) and it appears `[lang="en"]` is much faster than `:lang(en)`.

SamaneriVimala: This is just what was recommended at the website, but I guess you're right. I will change it.

PR #343, opened 2018-01-04
[SUCE-387] Dictionary pages

SamaneriVimala:


client/elements/sc-page-dictionary.html
```diff
@@ -217,25 +218,30 @@ <h3>Similar Spelling</h3>
...
+                  glossaryList: Object,
+                  dictionaryAdjacent: Array,
+                  dictionarySimilar: Array
               }
           }
 
+          static get observers() {
+              return [
+                  '_getGlossaryItems(glossaryList,adjacentReturns,adjacent)',
+                  '_getGlossaryItems(glossaryList,similarReturns)'
```
TrebuhD: When `glossaryList` changes, `_getGlossaryItems` will get called twice.

client/elements/sc-page-dictionary.html
```diff
@@ -268,42 +272,54 @@ <h3>Similar Spelling</h3>
...
+                  let glossText = '';
+                  if (glossaryList) {
+                      for (let glossWord in inputArray[0]) {
+                          let glossLookup = inputArray[0][glossWord];
+                          glossText = `<a href="/define/${glossLookup}">${glossLookup}</a>`;
+                          if (glossaryList[glossLookup]){
+                              glossText += `<em> (${glossaryList[glossLookup]})</em>`
+                          }
+                          glossaryObject = {"glossword": glossLookup, "glosstext": glossText};
+                          glossary.push(glossaryObject);
```
TrebuhD: 1. Please use camelCase when naming object keys. The double quotes are not needed.
2. This can be inline:
```
glossary.push({ glossWord: glossLookup, glossText: glossText })

File: client/elements/sc-page-dictionary.html [hunk 3]
```diff
@@ -268,42 +272,54 @@ <h3>Similar Spelling</h3>
...
-          _getGlossaryDescription(glossword) {
-              console.log(this.glossaryReturns(glossword));
-              return this.glossaryReturns(glossword);
+          _getGlossaryItems(glossaryList, inputArray, adjacent=false) {
+              if (inputArray) {
+                  let glossary = [];
+                  let glossaryObject = {};
+                  let glossText = '';
+                  if (glossaryList) {
+                      for (let glossWord in inputArray[0]) {
```
TrebuhD: is `inputArray` (either `adjacentReturns` or `similarReturns`) always a one-element array? That is, there will never be a `inputArray[1]`? If so, the arangoDB query should probably return the single object instead of a single-object array.

server/server/app.py
```diff
@@ -34,7 +35,9 @@ def app_factory() -> Tuple[Api, Flask]:
     api.add_resource(Images, '/images/<string:division>/<int:vol>')
     api.add_resource(Epigraphs, '/epigraphs')
     api.add_resource(WhyWeRead, '/whyweread')
-    api.add_resource(Glossary, '/glossary/<string:glossword>')
+    api.add_resource(Glossary, '/glossary')
+    api.add_resource(DictionaryAdjacent, '/dictionary_adjacent/<string:word>')
+    api.add_resource(DictionarySimilar, '/dictionary_similar/<string:word>')
```
TrebuhD: Hmm, `adjacent` and `similar` are subtypes of `dictionary`, so properly addressed they should look like this:
```
api.add_resource(DictionaryAdjacent, '/dictionary/adjacent/<string:word>')
api.add_resource(DictionarySimilar, '/dictionary/similar/<string:word>')
```

client/elements/sc-page-dictionary.html
```diff
@@ -268,42 +272,54 @@ <h3>Similar Spelling</h3>
...
+                          let glossLookup = inputArray[0][glossWord];
+                          glossText = `<a href="/define/${glossLookup}">${glossLookup}</a>`;
```
TrebuhD: this line should be under an `else` condition below.

client/elements/sc-page-dictionary.html
```diff
@@ -268,42 +272,54 @@ <h3>Similar Spelling</h3>
...
+                          if (glossaryList[glossLookup]){
+                              glossText += `<em> (${glossaryList[glossLookup]})</em>`
+                          }
+                          glossaryObject = {"glossword": glossLookup, "glosstext": glossText};
+                          glossary.push(glossaryObject);
+                      }
+                      return (adjacent) ? this.dictionaryAdjacent = glossary : this.dictionarySimilar = glossary;
```
TrebuhD: This line is too complicated. Avoid having any logic inside tertiary operators. I'd move the logic up to a simple `if/else` condition.

client/elements/sc-page-dictionary.html
SamaneriVimala: I don't quite understand Arango DB in this. It gives me back an array with 10 items inside an array. So the first array has only 1 item. But when I try to tell Arango to give me back Array[0] instead, it gives me a completely blank array. How to do this?

client/elements/sc-page-dictionary.html
SamaneriVimala: I tried that but then it is called before the glossaryList is actually loaded properly and not again after it is loaded.

server/server/app.py
SamaneriVimala: They are subtypes of `dictionary_full`, not of `dictionary` but yet, point taken and will change that.

client/elements/sc-page-dictionary.html
TrebuhD: Oh, I see. Yeah, that would be a problem. I think you can fix it by calling `_getGlossaryItems()` once after it's ben loaded, inside`_glossaryLoaded()`. The observers would then react on subsequent changes of `adjacentReturns`, `adjacent` and `similarReturns`.

If it still doesn't work, we can leave it like that, as long as calling `_getGlossaryItems` twice doesn't incur a significant performance hit.

client/elements/sc-page-dictionary.html
TrebuhD: Sorry, disregard this comment. I didn't notice the plus sign in `` glossText += `<em> (${glossaryList[glossLookup]})</em>```.

server/server/app.py
TrebuhD: [REST resource naming](http://www.restapitutorial.com/lessons/restfulresourcenaming.html) is a large and rather difficult subject, but it really pays off to have a proper RESTful API.

client/elements/sc-page-dictionary.html
SamaneriVimala: I tried that too but then it gets called before the inputArrays (`adjacentReturns` & `adjacentSimilar`) are loaded. 
It should not be a significant performance hit because `adjacentReturns` & `adjacentSimilar` are very small. `glossaryList` is pretty big but only gets loaded once and never changes after that.

client/elements/sc-page-dictionary.html
TrebuhD: I just tried - yeah, it's quite weird. Don't worry about it.

PR #353, opened 2018-01-09
Internationalization changes for static file modification by Sujato

TrebuhD:


client/elements/static/about-page.html
```diff
@@ -1,17 +1,19 @@
...
     </style>
 
     <div id="page-wrap">
-      <main>
+      <main>®
         <section>
 
           <article>
-            <h1>About SuttaCentral</h1>
+            <h1>About Sutta <a href="#">asdasd</a> Central</h1>
```
gradam: asdasd???? do we have a new name?

client/localization/extractStaticStrings.py
```diff
@@ -0,0 +1,27 @@
+from lxml import etree
+from lxml.etree import tostring
+from itertools import chain
```
gradam: it should be before 3rd party imports

client/localization/pootle/generatePoFiles.py
```diff
@@ -0,0 +1,26 @@
...
+        'POT-Creation-Date': str(datetime.datetime.now()),
+        'Content-Type': 'text/plain; charset=utf-8'
+    }
+    for key in file_data:
+        entry = polib.POEntry(
+            msgid=key,
+            msgstr=file_data[key]
+        )
+        po.append(entry)
+    po.save('generatedPoFiles/{}.po'.format(element_name))
```
gradam: please use f-strings

client/localization/pootle/generatePoFiles.py
```diff
@@ -0,0 +1,26 @@
...
+
+
+if __name__ == '__main__':
+    for fileName in os.listdir('prePootleFiles'):
+        generate_po_file(fileName, open('prePootleFiles/{}'.format(fileName), 'r'))
```
gradam: please use f-strings

PR #361, opened 2018-01-10
SUCE-401 automate solution

gradam:


client/localization/localizer.py
```diff
@@ -0,0 +1,40 @@
...
+
+def print_help():
+    print('''
+Available commands:')
+*  to_pootle
+*  from_pootle
+
+example usage:
+python localizer.py to_pootle'''
+          )
```
TrebuhD: please format this

PR #376, opened 2018-01-16
[SUCE-443]- dockerize Pootle

TrebuhD:


nginx/conf.d/sc.nginx.conf
```diff
@@ -34,13 +34,28 @@ add_header X-Frame-Options SAMEORIGIN;
 add_header X-XSS-Protection "1; mode=block";
 
 location /api/docs/ {
-    rewrite ^/api/docs/(.*)$ /$1 break;
+    rewrite  ^/api/docs/(.*)$ /$1 break;
```
gradam: why 2 spaces here?

pootle/pootle.conf
```diff
@@ -0,0 +1,201 @@
...
+
+# Local time zone for this installation. Choices can be found here:
+# http://en.wikipedia.org/wiki/List_of_tz_zones_by_name
+# although not all choices may be available on all operating systems.
+# If running in a Windows environment this must be set to the same as your
+# system time zone.
+TIME_ZONE = 'Europe/Berlin'
+
+# Make this unique, and don't share it with anybody.
+SECRET_KEY = os.environ.get('SECRET_KEY', '4JY+9C9kEWASJk5MVkYH8ynd5ka9aRq56jssTwa2key8CZrSipNi1vD/1lPfzcqx/UY=')
```
gradam: "Make this unique, and don't share it with anybody" :laughing:

TrebuhD: The key here is a fallback if the SECRET_KEY environment variable isn't set on the machine, which it should be.

PR #393, opened 2018-01-19
[SUCE-197] parallels and suttaplex display changes

SamaneriVimala:


client/elements/menus/sc-language-menu.html
```diff
@@ -184,6 +189,18 @@
               this.loadResources(this.resolveUrl(`/localization/elements/sc-language-menu/${this.language}.json`));
           }
 
+          _computeLanguageMenu() {
+            const languageMenu = [];
+            for (let item in this.languageChoice) {
+              if (this.languageChoice[item].lang in this.rootLangMappings) {
```
TrebuhD: it's better if we use the `languageMenu` that was passed by `computed` as an argument into this function and operate on that instead of referring to `this.languageMenu`. If we pass `rootLangMappings` in there too, it becomes a [pure function](https://en.wikipedia.org/wiki/Pure_function), which is very good.

client/elements/menus/sc-language-menu.html
```diff
@@ -184,6 +189,18 @@
...
+                languageMenu.splice(0, 0, this.languageChoice[item]);
```
TrebuhD: hmm, shouldn't `this.languageChoice[item]` be just `item`?

client/elements/menus/sc-language-menu.html
```diff
@@ -184,6 +189,18 @@
...
+              } else {
+                languageMenu.push(this.languageChoice[item]);
```
TrebuhD: ditto

client/elements/suttaplex/sc-parallel-item.html
```diff
@@ -117,12 +117,12 @@
...
 
       <div class="parallel-item-main-info-container">
-        <div class="parallel-item-title vertical-margin-xs">[[_computeTitle(parallelItem.*)]]</div>
+        <div class="parallel-item-title vertical-margin-xs" title="[[_getHeadingTitle(parallelItem,localize)]]">[[_computeTitle(parallelItem.*)]]</div>
 
         <div class="parallel-item-details d-flex align-items-center vertical-margin-xs">
           <template is="dom-if" if="[[_getAcronymOrId(parallelItem)]]">
             <div class$="parallel-item-id [[_computeRootIdVisibility()]]">
-              <div title="{{localize('suttaCentralID')}}">[[_getAcronymOrId(parallelItem)]]</div>
+              <div title="[[_getAcronymTitle(parallelItem,localize)]]">[[_transformId(parallelItem.to,expansionData)]]</div>
```
TrebuhD: space

client/elements/suttaplex/sc-parallel-item.html
```diff
@@ -117,12 +117,12 @@
     <div class="parallel-item d-flex justify-content-space-between">
 
       <div class="parallel-item-main-info-container">
-        <div class="parallel-item-title vertical-margin-xs">[[_computeTitle(parallelItem.*)]]</div>
+        <div class="parallel-item-title vertical-margin-xs" title="[[_getHeadingTitle(parallelItem,localize)]]">[[_computeTitle(parallelItem.*)]]</div>
```
TrebuhD: please insert a space after `,` for readability.

client/elements/suttaplex/sc-parallel-item.html
```diff
@@ -214,7 +214,8 @@
                       type: String,
                       statePath: 'siteLanguage',
                       observer: '_siteLanguageChanged'
-                  }
+                  },
+                  expansionData: Array
```
TrebuhD: I don't think we discussed this earlier, but a while ago we decided with Kuba that the

```
expansionData: {
    type: Array
}
```
syntax was superior to this one :). It's your call, but at the end it's about readability and keeping a uniform code style. If you take a look at the Google Polymer elements, I think they insert 2 blank lines between each property.

client/elements/suttaplex/sc-parallel-item.html
```diff
@@ -228,7 +229,15 @@
...
-                  return this.parallelItem.acronym ? this.parallelItem.acronym : this.parallelItem.uid;
+                  return this._transformId(this.parallelItem.to, this.expansionData);
+              }
+          }
+
+          _getHeadingTitle(item,localize) {
+              if (!item || item.translated_title || item.original_title) {
+                  return '';
+               } else {
+                  return this._getAcronymTitle(item,localize);
```
TrebuhD: whitespace after `,`.

client/elements/suttaplex/sc-parallel-item.html
```diff
@@ -258,8 +267,51 @@
...
+                return '';
+            }
+            if(ids.includes('-')) {
+              const [id1, id2] = ids.split('-#');
+              return `#${id1}-${id2}`;
+            }
+            return `#${ids}`;
+          }
+
+          _transformId(rootId, expansionData) {
```
TrebuhD: This function is a bit complicated, I think it should be documented with a comment or two

client/elements/suttaplex/sc-parallel-list.html
```diff
@@ -181,13 +190,15 @@
...
-                  }
+                  },
+                  expansionData: Array
               }
           }
 
           ready() {
               super.ready();
               this.addEventListener('loading-results-changed', this._resizeParentList);
+              if (!this.expansionData) {this.$.uid_expansion_ajax.generateRequest()};
```
TrebuhD: the content of the if block should be on a separate line.

client/elements/menus/sc-language-menu.html
SamaneriVimala: item is just a number. So why would you want to add a number to the first place in the array rather than the whole object?

TrebuhD: Oh, you're right. I got too used to `forEach` loops. I think we should use one here.

TrebuhD: Also, if it's an index, it should be named `itemIndex` and not `item`

SamaneriVimala: OK, will use `forEach`. Looks indeed a bit better.

PR #396, opened 2018-01-19
SUCE-209-static-content-mapping

gradam:


client/elements/sc-donate-now-page.html
```diff
@@ -437,10 +437,11 @@
           }
 
           _showToast(toastType, text) {
+              const localized = this.localize(text);
```
TrebuhD: should be `localizedText`

PR #419, opened 2018-01-24
enable Pootle Translation Memory

TrebuhD:


pootle/pootle.conf
```diff
@@ -88,21 +88,21 @@ CACHES = {
...
-#        # both included. Defaults to ``1.0`` if not provided.
-#        'WEIGHT': 1.0,
-#        # For valid values for MIN_SCORE see
-#        # http://www.elasticsearch.org/guide/en/elasticsearch/reference/current/common-options.html#fuzziness
-#        'MIN_SCORE': 'AUTO',
-#    },
+POOTLE_TM_SERVER = {
+    'local': {
+        'ENGINE': 'pootle.core.search.backends.ElasticSearchBackend',
+        'HOST': 'localhost',
```
gradam: I think it is on the different host than localhost. Maybe sc-elasticsearch?

PR #421, opened 2018-01-24
[SUCE-474] small tweeks to display

SamaneriVimala:


client/elements/suttaplex/sc-parallel-item.html
```diff
@@ -283,6 +283,10 @@
               return item.translated_title && item.original_title ? '' : 'hide-element';
           }
 
+          _removeSuttaText(item) {
```
TrebuhD: It seems this should be named `getTitleWithoutSuttaText` or something like that.

client/elements/suttaplex/sc-parallel-list.html
```diff
@@ -339,7 +339,7 @@
               }
               if (ids.includes('-')) {
                   const [id1, id2] = ids.split('-#');
-                  return `#${id1}--${id2}`;
```
TrebuhD: I think the double dash is needed for range selection in texts. @gradam can you confirm?

SamaneriVimala: I can change it back and remove them elsewhere.

TrebuhD: I've approved the ticket, but please verify that range selection is working. Otherwise, we can wait for @gradam's confirmation that this is ok.

PR #430, opened 2018-01-26
Restore resembling parallels

blake-sc:


server/server/common/queries.py
```diff
@@ -261,7 +261,7 @@
         to: {
             to: e.to,
             volpages: v.volpage ? v.volpage : legacy_volpages,
-            acronym: v.acronym,
+            acronym: v.acronym ,
```
gradam: no need for space here

PR #439, opened 2018-01-28
[SUCE-494] fix internal routing static pages

SamaneriVimala:


client/elements/text/sc-simple-text.html
```diff
@@ -230,6 +230,11 @@
           ready() {
               super.ready();
               this.loadResources(this.resolveUrl(`/localization/elements/sc-text/en.json`), this.fallbackLanguage, true);
+              this.addEventListener('click', () => {
```
TrebuhD: Does this event listener run when clicking the text?

SamaneriVimala: It runs when clicking anywhere inside the text. Normal internal links referring to an ID do not work within these WC so it needs an event listener. I cannot put event listeners on things inside the inner-h-t-m-l so I put it on the main WC instead and then it checks the id that was clicked from the url hash and uses that.

SamaneriVimala: But please wait, I want to try something else that has just come to mind and will notify you when I'm done.

PR #454, opened 2018-01-30
[SUCE-497] lotus to homepage link added

SamaneriVimala:


client/elements/sc-page-selector.html
```diff
@@ -99,6 +103,10 @@
         background-color: var(--sc-primary-color);
       }
 
+      .logo-hidden {
```
TrebuhD: A good naming convention for these classes with only one CSS property is to name them more generally by that property: eg. `display-none`, without referring to any specific element (`logo` in this case). This has 2 upsides:
- Looking at the HTML, you know right away what the `display-none` class does,
- If, in the future, you decide to use `display: none` for another node in this file which isn't a logo, you won't have to rename this CSS class.

client/elements/sc-page-selector.html
```diff
@@ -112,6 +120,10 @@
 
         <app-header id="header" condenses reveals effects="waterfall" slot="header" mode$="[[toolbarMode]]">
           <app-toolbar class="toolbar-header">
+            <a href="/" class$="[[_showHideHomeButton(isDrawerOpen, shouldShowStaticPage)]]">
```
TrebuhD: `_showHide` should probably be `_shouldHide` or `_computeHide`

client/elements/sc-page-selector.html
```diff
@@ -71,6 +71,10 @@
         margin-right: var(--sc-size-md);
       }
 
+      #to-home-button {
```
TrebuhD: ID names use underscores instead of dashes

PR #458, opened 2018-02-01
[SUCE-494] Extending polymer element for static page scrolling

SamaneriVimala:
Have a look at this test for extending the static pages using a class. It works fine as far as I can see.


client/elements/addons/sc-static-page.html
```diff
@@ -0,0 +1,56 @@
...
+              }
+          }
+
+          connectedCallback() {
+            super.connectedCallback();
+          }
+
+          ready() {
+              super.ready();
+              this.shadowRoot.querySelector('main').addEventListener('click', () => {
```
TrebuhD: This will fail with an error if there is no `<main>` element in the subclass. It should either be wrapped in a `try/catch` block or an `if this.shadowRoot.querySelector('main')` condition.

client/elements/addons/sc-static-page.html
```diff
@@ -0,0 +1,56 @@
...
+                      behavior: 'instant',
+                      block: 'start',
+                      inline: 'nearest'
+                  });
+              }
+              return sectionId;
+          }
+      }
+
+      customElements.define('sc-static-page', SCStaticPage);
```
TrebuhD: the convention is to have `'sc-static-page'` defined like this at the top of the script:
```
          static get is() {
              return 'sc-static-page';
          }
```
and to use `SCStaticPage.is` here instead.

client/elements/static-templates/abhidhamma.html
```diff
@@ -1,6 +1,6 @@
 <link rel="import" href="../../bower_components/polymer/polymer.html">
 <link rel="import" href="../../bower_components/app-localize-behavior/app-localize-behavior.html">
-
+<link rel="import" href="../addons/sc-static-page.html" rel="import"/>
```
TrebuhD: two `rel="import"`s.

client/elements/static-templates/discourses.html
```diff
@@ -1,6 +1,7 @@
 <link rel="import" href="../../bower_components/polymer/polymer.html">
 <link rel="import" href="../../styles/static-styles.html">
 <link rel="import" href="../../bower_components/app-localize-behavior/app-localize-behavior.html">
+<link rel="import" href="../addons/sc-static-page.html" rel="import"/>
```
TrebuhD: ditto

client/elements/static-templates/methodology-page.html
```diff
@@ -1,6 +1,7 @@
 <link rel="import" href="../../bower_components/polymer/polymer.html">
 <link rel="import" href="../../styles/static-styles.html">
 <link rel="import" href="../../bower_components/app-localize-behavior/app-localize-behavior.html">
+<link rel="import" href="../addons/sc-static-page.html" rel="import"/>
```
TrebuhD: ditto

client/elements/static-templates/names.html
```diff
@@ -1,6 +1,7 @@
 <link rel="import" href="../../bower_components/polymer/polymer.html">
 <link rel="import" href="../../styles/static-styles.html">
 <link rel="import" href="../../bower_components/app-localize-behavior/app-localize-behavior.html">
+<link rel="import" href="../addons/sc-static-page.html" rel="import"/>
```
TrebuhD: ditto

client/elements/static-templates/subjects.html
```diff
@@ -1,6 +1,7 @@
 <link rel="import" href="../../bower_components/polymer/polymer.html">
 <link rel="import" href="../../styles/static-styles.html">
 <link rel="import" href="../../bower_components/app-localize-behavior/app-localize-behavior.html">
+<link rel="import" href="../addons/sc-static-page.html" rel="import"/>
```
TrebuhD: ditto

client/elements/static-templates/similes.html
```diff
@@ -1,6 +1,7 @@
 <link rel="import" href="../../bower_components/polymer/polymer.html">
 <link rel="import" href="../../styles/static-styles.html">
 <link rel="import" href="../../bower_components/app-localize-behavior/app-localize-behavior.html">
+<link rel="import" href="../addons/sc-static-page.html" rel="import"/>
```
TrebuhD: ditto

client/elements/static-templates/terminology.html
```diff
@@ -1,6 +1,7 @@
 <link rel="import" href="../../bower_components/polymer/polymer.html">
 <link rel="import" href="../../styles/static-styles.html">
 <link rel="import" href="../../bower_components/app-localize-behavior/app-localize-behavior.html">
+<link rel="import" href="../addons/sc-static-page.html" rel="import"/>
```
TrebuhD: ditto

client/elements/static-templates/vinaya.html
```diff
@@ -1,6 +1,7 @@
 <link rel="import" href="../../bower_components/polymer/polymer.html">
 <link rel="import" href="../../styles/static-styles.html">
 <link rel="import" href="../../bower_components/app-localize-behavior/app-localize-behavior.html">
+<link rel="import" href="../addons/sc-static-page.html" rel="import"/>
```
TrebuhD: ditto

client/elements/static/home-page.html
```diff
@@ -439,7 +439,9 @@
      <div class="big-card-title" inner-text="{{localize('24cbc4d3253a2779ae4dab326e42659b')}}">
      </div>
      <div class="three-elements-grid">
-      <paper-card class="three-elements-big paper-card-with-button" elevation="2" image="/img/home-page/14.jpg">
+      <paper-card alt="Sixteen gilt and lacquer leaves with two gilt and lacquer book covers. Wood coated with lacquer,
```
TrebuhD: hmm, this should have been extracted by the localizer extraction script. @gradam

client/elements/addons/sc-static-page.html
SamaneriVimala: Here: https://www.polymer-project.org/2.0/docs/devguide/registering-elements is says specifically to do it like this.

>// Associate the new class with an element name
>customElements.define('my-element', MyElement);

Wile at the element that is being extended they use the `get is()`.

TrebuhD: OK, sounds good! I wonder why they keep two distinct conventions.

PR #462, opened 2018-02-01
Return search results for root languages

blake-sc:


server/server/search/query.py
```diff
@@ -147,9 +148,13 @@ def search(query: str, highlight=True, offset=0, limit=10, language='en',
         indexes.append('en-dict')
     if lang:
         indexes.append(lang)
+        
+
```
TrebuhD: do we need so many blank lines here?

PR #464, opened 2018-02-01
[SUCE-501] Adding polish static pages

SamaneriVimala:
I added some code to extract_static_strings.py to create pl.json for static pages and ran the script. Is that all that's required here?


TrebuhD:
With our fallback functionality enabled, we don't need to store english strings in `pl.json`, empty json files will suffice. Also, the .json files should be generated based for all translated languages based on the list we use.

client/localization/extract_static_strings.py
```diff
@@ -75,6 +75,9 @@ def run():
             formatted_data = {'en': data}
             json.dump(formatted_data, f, indent=4, ensure_ascii=False)
 
+        with (element_dir / 'pl.json').open('w') as f:
```
TrebuhD: @gradam can we generate empty json files for all translated languages here?

PR #468, opened 2018-02-02
add workbox CLI dependency

TrebuhD:


client/builder.sh
```diff
@@ -2,6 +2,7 @@
 echo "BUILDING FRONTEND FOR PRODUCTION"
 npm install
 npm install -g polymer-cli --unsafe-perm
+npm install -g workbox-cli
```
gradam: do we need to install it twice?

TrebuhD: the other one didn't work. I'll delete it.

PR #479, opened 2018-02-04
[SUCE-500] links to texts on search pages use uid

SamaneriVimala:


server/server/search/view.py
```diff
@@ -72,6 +72,10 @@ def get(self):
         query = request.args.get('query', None)
         language = request.args.get('language', current_app.config.get('DEFAULT_LANGUAGE'))
 
+        author_file = Path('nextdata/additional-info/author_edition.json')
```
gradam: you can't use files from nextdata in the server code. All data should be taken from ArangoDB

PR #481, opened 2018-02-05
Suce 501 fallback language

gradam:


TrebuhD:
Does this work? The generated files should be empty, that is like that:
```
{
    "pl": {}
}
```

client/localization/extract_static_strings.py
```diff
@@ -8,6 +8,9 @@
 from bs4 import BeautifulSoup
 from bs4.element import NavigableString
 
+AVAILABLE_LANGUAGES = ['en', 'pl']
```
TrebuhD: should we not be getting the list of languages from the JSON file we're loading into arangodb?

client/localization/extract_static_strings.py
```diff
@@ -8,6 +8,9 @@
...
+AVAILABLE_LANGUAGES.remove('en')
```
TrebuhD: these two lines don't make sense.

client/localization/extract_static_strings.py
gradam: We can't do that easily from the client

client/localization/extract_static_strings.py
gradam: not true. This way you don't have to care about which language is 'source' one so you just write list of all available languages

PR #483, opened 2018-02-05
[SUCE-509]-Original language texts are displaying funny

TrebuhD:


client/elements/text/sc-text-page-selector.html
```diff
@@ -254,11 +254,14 @@ <h3>Could not find the sutta text.</h3>
           }
 
           _shouldDisplayStepper(isLoading, translatedSutta) {
+              if (!translatedSutta) {
```
gradam: Why?

TrebuhD: There was an error if `translatedSutta` was null. From this function it seems the stepper only takes translated suttas into account, so I made it return false if there's no translated sutta.

gradam: that's strange. I mean, the data needed for this to work without translatedSutta should be here so it is probably problem with passing only translatedSutta and not falling back to the original one when there is no translated one.

PR #486, opened 2018-02-05
[SUCE-494] increasing load time

SamaneriVimala:


client/elements/addons/sc-static-page.html
```diff
@@ -36,7 +36,7 @@
               };
               setTimeout(() => {
                       this.currentId = this._scrollToSection(window.location.hash, this.currentId);
-              }, 1000);
+              }, 1500);
```
TrebuhD: Try using `requestAnimationFrame` instead of `setTimeout`, it might just work.

SamaneriVimala: No, tried it but it doesn't.

PR #495, opened 2018-02-06
Add po texts to search

blake-sc:


server/server/common/queries.py
```diff
@@ -19,6 +19,24 @@
         }
 '''
 
+PO_TEXTS_BY_LANG = '''
+FOR text IN po_strings
+    FILTER text.lang == @lang
+    LET root_lang = (
+        RETURN DOCUMENT(CONCAT('root', text.uid)).root_lang
```
gradam: shouldn't it be `CONCAT('root/', text.uid))` ?

server/server/search/texts.py
```diff
@@ -170,29 +217,27 @@ def load_index_config(config_name):
...
-            if uid in to_delete:
-                continue
-            if mtime <= current_mtimes.get(uid):
-                to_add.pop(uid)
-        logger.info(
-            "For index {} ({}), {} files already indexed, {} files to be added, {} files to be deleted".format(
-                self.index_name, self.index_alias, len(stored_mtimes), len(to_add), len(to_delete)))
-        chunks = self.yield_docs_from_dir(self.lang, size=500000, to_add=to_add,
-                                          to_delete=to_delete)
+        #stored_mtimes = {hit["_id"]: hit["_source"]["mtime"] for hit in scan(self.es,
```
gradam: we should not leave redundant code commented. Just delete it

PR #504, opened 2018-02-07
static pages class

SamaneriVimala:
There is no Jira ticket for this because functionality has not changed. I made all functions into one class SCStaticPage and imported that in all static pages. I did not actually manage to get the routing working without the setTimeout which I now set to 2000 but will make a new Jira ticket for this to be sorted out properly. It works most of the time now, but not always depending on user's browser speed.
So I just want to push this one through with the setTimeout.


client/elements/addons/sc-static-page.html
```diff
@@ -15,28 +15,37 @@
...
+                  language: {
+                      type: String,
+                      statePath: 'siteLanguage',
+                      observer: '_siteLanguageChanged'
+                  },
+                  fallbackLanguage: {
+                      type: String,
+                      value: 'en'
+                  },
+                  languageStringPath: {
```
TrebuhD: In hindsight, it would make more sense if this was named `localizedStringsPath` or `localizedStringResourcesPath`. What do you think?

client/elements/addons/sc-static-page.html
```diff
@@ -15,28 +15,37 @@
...
+              super.connectedCallback();
+              this.loadResources(this.resolveUrl(`${this.languageStringPath}${this.fallbackLanguage}.json`), this.fallbackLanguage, true);
               try {
                   this.shadowRoot.querySelector('main').addEventListener('click', () => {
                       setTimeout(() => {
                           this.currentId = this._scrollToSection(window.location.hash, this.currentId);
                       }, 0);
                   });
               } catch (e) {
-                  console.error(e);
```
TrebuhD: If we don't print the error message, this becomes a silently passing error, which makes debugging really difficult in the future.

client/elements/static-templates/donations-page.html
```diff
@@ -3,7 +3,7 @@
 <link rel="import" href="../../bower_components/paper-dialog/paper-dialog.html">
 <link rel="import" href="../../bower_components/paper-dialog-scrollable/paper-dialog-scrollable.html">
 <link rel="import" href="../../bower_components/app-localize-behavior/app-localize-behavior.html">
```
TrebuhD: we don't need those imports in here anymore

client/elements/addons/sc-static-page.html
SamaneriVimala: Maybe, but it fires at every page startup. I think maybe I should make it an if-statement instead.

TrebuhD: What's the error message?

PR #508, opened 2018-02-08
fixed missing prev/next on root texts

gradam:


client/elements/text/sc-text-page-selector.html
```diff
@@ -221,6 +227,13 @@ <h3>[[lastError.statusText]]</h3>
                   if (this.responseData.translation) {
                       this.authorUid = this.responseData.translation.author_uid;
                   }
+                  if(this.translatedSutta) {
+                      this.next = this.translatedSutta.next;
+                      this.previous = this.translatedSutta.previous;
+                  } else {
```
TrebuhD: what happens if `rootSutta` is `undefined`?

client/elements/text/sc-text-page-selector.html
```diff
@@ -221,6 +227,13 @@ <h3>[[lastError.statusText]]</h3>
                   if (this.responseData.translation) {
                       this.authorUid = this.responseData.translation.author_uid;
                   }
+                  if(this.translatedSutta) {
```
TrebuhD: missing space before `(`

client/elements/text/sc-text-page-selector.html
gradam: I think the situation where both, translatedSutta and rootSutta, is impossible. Am I missing something?

TrebuhD: I wouldn't say it's impossible. It could easily happen in case of a server/database error.

PR #513, opened 2018-02-09
[SUCE-514] text images display

SamaneriVimala:


client/elements/text/sc-simple-text.html
```diff
@@ -305,13 +305,16 @@
...
+              let divisionId = /^[a-z]+/.exec(this.sutta.uid)[0];
+              if (divisionId === 'pli') {divisionId = 'vi'};
+              if (divisionId === 'iti') {divisionId = 'it'};
               if (this.paragraphs && this.showParagraphs) {
                   this.paragraphs.forEach((paragraph) => {
                       const refs = this.$.simple_text_content.querySelectorAll(`.${paragraph.uid}`);
                       Array.from(refs).forEach((item) => {
-                          if (item.id.startsWith('pts')) {
-                              this._processVolPageInfo(item, divisionId);
+                          if (divisionId !== 'tha' && divisionId !== 'thi' && !item.id.includes('2ed')
```
TrebuhD: This if condition is awfully complicated, I'd extract it to a separate function with a clear name.

client/elements/text/sc-simple-text.html
```diff
@@ -305,13 +305,16 @@
               if (!this.sutta || !this.sutta.uid) {
                   return;
               }
-              const divisionId = /^[a-z]+/.exec(this.sutta.uid)[0];
+              let divisionId = /^[a-z]+/.exec(this.sutta.uid)[0];
+              if (divisionId === 'pli') {divisionId = 'vi'};
```
TrebuhD: for one-line if conditions, curly brackets are not needed.

client/elements/text/sc-simple-text.html
```diff
@@ -305,13 +305,16 @@
...
+              if (divisionId === 'iti') {divisionId = 'it'};
```
TrebuhD: ditto

client/elements/text/sc-simple-text.html
```diff
@@ -320,22 +323,32 @@
...
+                pageNumber = vol;
+                vol = '1';
+                displayText = `${pageNumber}`;
+              };
+              pageNumber = Number(pageNumber);
               vol = Number(vol);
-
+              if (vol === 0) {
+                displayText = `${pageNumber}`;
+                vol = 1
```
TrebuhD: missing semicolon

client/elements/text/sc-text-image.html
```diff
@@ -62,11 +62,19 @@
...
                       type: Object
+                  },
+                  lastDetail: {
+                      type: Object
                   }
               }
           }
 
           showImage(detail) {
+            if (!detail || (this.lastDetail && detail.vol === this.lastDetail.vol
```
TrebuhD: similarly, I'd extract the second part of this complex if statement to a separate function (named `_didDetailChange` for example) to make it easier to read.

client/elements/text/sc-text-image.html
```diff
@@ -83,6 +91,15 @@
           }
 
           _handleResponse() {
+              if (!this.imageData || !this.imageData[0]) {
+                return;
+              };
+              let pageNumber = 0;
+              for (let item = 0; item < this.imageData.length; item++) {
+                  if (this.imageData[item].pageNumber === this.pageNumber) {
```
TrebuhD: 1. When `this.imageData[item].pageNumber` equals `1`, `item` equals 0. This means `pageNumber` will be always set to `this.imageData[item].pageNumber` minus 1. Is this what we want?
2. `item` refers to an numerical item index, so it's wrongly named.
3. This loop can be simplified to:
```
if (this.imageData.some(page => page.pageNumber === this.pageNumber)) {
    pageNumber = this.pageNumber - 1;
}
```

`some` returns true if there exist any items that satisfy a condition given in the function body.

client/elements/text/sc-text-image.html
```diff
@@ -83,6 +91,15 @@
...
+                      pageNumber = item;
+                  }
```
TrebuhD: If we don't put a `break` statement here, this loop will keep going even after it's found the element we're looking for.

client/elements/text/sc-simple-text.html
SamaneriVimala: Yup, this stuff is so complicated with many exceptions and you cannot find a common ground between them so you have to get all this in. Will make it into a separate function.

client/elements/text/sc-text-image.html
SamaneriVimala: 1. Yes, because the array starts at 0 and not at 1 while we start counting at 1.

client/elements/text/sc-text-image.html
SamaneriVimala: I see. But I replaced the whole thing with your loop now. But thanks for the tip for next time i.e. as soon as value is found, break.

client/elements/text/sc-simple-text.html
```diff
@@ -323,6 +322,20 @@
               }
           }
 
+          _shouldDisplayBookIcon(divisionId, itemId) {
+              let divisionIdMatch = false;
+              let secondEdition = false;
+              let correctId = false;
+              (divisionId === 'tha' || divisionId === 'thi') ? divisionIdMatch = true : divisionIdMatch = false;
```
TrebuhD: `let divisionIdMatch = (divisionId === 'tha' || divisionId === 'thi')`

client/elements/text/sc-simple-text.html
```diff
@@ -323,6 +322,20 @@
...
+              itemId.includes('2ed') ? secondEdition = true : secondEdition = false;
```
TrebuhD: `let secondEdition = itemId.includes('2ed')`

client/elements/text/sc-simple-text.html
```diff
@@ -323,6 +322,20 @@
...
+              (itemId.startsWith('pts-vp-pli') || itemId.match(/pts[1-9]/)) ? correctId = true : correctId = false;
```
TrebuhD: `let correctId = (itemId.startsWith('pts-vp-pli') || itemId.match(/pts[1-9]/))`

client/elements/text/sc-simple-text.html
```diff
@@ -323,6 +322,20 @@
...
+              if (!divisionIdMatch && !secondEdition && correctId) {
```
TrebuhD: `return (!divisionIdMatch && !secondEdition && correctId)`

SamaneriVimala: As we say in Aus: Ta!

PR #517, opened 2018-02-09
[SUCE-474] fixed missing root texts

SamaneriVimala:
I also cleaned up the code with respect to camelCasing and let/const.


SamaneriVimala:
I accidentally pushed the fonts to the wrong branch. Can you have a look and push anyway? Thanks.

TrebuhD:
a question: shouldn't we be pulling noto fonts from Google fonts? It's possible users have them cached already from browsing other websites.

PR #518, opened 2018-02-11
[SUCE-414]: UI for collection choice

TrebuhD:


TrebuhD:
Don't merge yet please. I found one more thing after testing on next.

client/workbox-cli-config.js
```diff
@@ -8,7 +8,8 @@ module.exports = {
         'img/**/*',
         'img/home-page/**/*',
         'img/static-pages/**/*',
-        'files/fonts/**/*.otf'
+        'files/fonts/**/*.otf',
```
SamaneriVimala: I think we can get rid of the .otf files. I deleted them and they are no longer used. Sujato wanted woff2 only (but I have one woff running too).

PR #519, opened 2018-02-12
[SUCE-543] load single image

SamaneriVimala:


client/elements/text/sc-text-image.html
```diff
@@ -81,6 +86,13 @@
             this.$.ajax.generateRequest();
           }
 
+          _getShowUrl(imageData) {
+              return this._getImageUrl(imageData[0].name);
+          }
+          _getImageAlt(imageData) {
```
TrebuhD: there should be a blank line before this line

client/elements/text/sc-text-image.html
```diff
@@ -65,6 +62,14 @@
                   },
                   lastDetail: {
                       type: Object
+                  },
+                  showUrl: {
+                      type: String,
+                      computed: '_getShowUrl(imageData)'
```
TrebuhD: (not important): `_getShowUrl(imageData)` could be inlined in the template

client/elements/text/sc-text-image.html
```diff
@@ -104,23 +116,34 @@
...
                 return;
               };
-              let pageNumber = 0;
-              if (this.imageData.some(page => page.pageNumber === this.pageNumber)) {
-                  pageNumber = this.pageNumber - 1;
-              }
               requestAnimationFrame(() => {
                   const options = {
-                    url: (e) => e.dataUrl,
+                    url: `'src'`,
```
TrebuhD: This line is interesting. How does this work?

client/elements/text/sc-text-image.html
SamaneriVimala: doesn't work in this case because the viewer imports src and it tends to see the function as a sting rather than execute the function. So your url becomes `https://suttacentral.net/[[_getShowUrl(imageData)]]` and it gives an error.

client/elements/text/sc-text-image.html
SamaneriVimala: See readme file with the viewerjs element. It's an interesting component. It takes the src from the image and displays that in the viewer. I have to hide the actual image that is in template itself as display.none.

PR #520, opened 2018-02-12
[SUCE-511] i18n implementation

SamaneriVimala:


client/styles/sc-font-styles.html
```diff
@@ -607,6 +699,13 @@
         line-height: 24px;
       };
 
+      --paper-font-body2-chinese: {
+        @apply --sc-traditional-chinese-font;
+        @apply --sc-dense-font-size-s;
+        font-weight: 500;
```
TrebuhD: I seem to recall @sujato saying we should only use `font-weight` 400 and 700.

SamaneriVimala: I'll ask him. I did not change the font-weight but only copied the paper-font-body2 to a chinese version with chinese font. Good point. Thanks.

PR #546, opened 2018-02-16
[SUCE-558] labels not showing when infomode is off

SamaneriVimala:
This does not fix the issue with keeping an indented space in front of the paragraphs when you turn infomode off.
The change on sc-segmented-text is only because I got an error in the consule if there is no segment. You might want to change that.

### DIFF
--- a/client/styles/sc-text-paragraph-num-styles.html
+++ b/client/styles/sc-text-paragraph-num-styles.html
@@ -2,437 +2,441 @@
   <template>
     <style>
       /*a.bl:before {content:"bl "}*/
-      .adh-v:before {
+      .textual-info-paragraph.adh-v:before {
         content: "Adhik-v "
       }
 
-      .apz:before {
+      .textual-info-paragraph.apz:before {
         content: "APZ "
       }
 
-      .ba:before {
+      .textual-info-paragraph.ba:before {
         content: "BA "
       }
 
-      .bd:before {
+      .textual-info-paragraph.bd:before {
         content: "BD "
       }
 
-      .bhaisv:before {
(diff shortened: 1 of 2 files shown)


client/styles/sc-text-paragraph-num-styles.html
```diff
@@ -2,437 +2,453 @@
...
+      .textual-info-paragraph.vn:before {
         content: "Vn "
       }
 
-      .wal48:before, .wal50:before, .wal52:before, .wal57c:before, .wal58:before, .wal59a:before, .wal60:before, .wal61:before, .wal68a:before, .wal70a:before, .wal70b:before, .wal76:before, .wal78:before, .wal80c:before {
+      .textual-info-paragraph.wal48:before, .textual-info-paragraph.wal50:before, .textual-info-paragraph.wal52:before,
+      .textual-info-paragraph.wal57c:before, .textual-info-paragraph.wal58:before, .textual-info-paragraph.wal59a:before,
+      .textual-info-paragraph.wal60:before, .textual-info-paragraph.wal61:before, .textual-info-paragraph.wal68a:before,
+      .textual-info-paragraph.wal70a:before, .textual-info-paragraph.wal70b:before, .textual-info-paragraph.wal76:before,
+      .textual-info-paragraph.wal78:before, .textual-info-paragraph.wal80c:before {
```
TrebuhD: hmm, maybe `.textual-info-paragraph[class^="wal"]:before` would work here? or if not, maybe `[class^="textual-info-paragraph wal"]:before`?

SamaneriVimala: Doubt it if that would work, they are very different classes.

PR #547, opened 2018-02-16
Various fixes

TrebuhD:
Fixed:
- the translations label appearing over the dropdown label
- shaky cursor over parallels toggle
- console error when computing ids in parallels and suttaplex (the suttaplex one was passing silently)

### DIFF
--- a/client/elements/suttaplex/sc-suttaplex.html
+++ b/client/elements/suttaplex/sc-suttaplex.html
@@ -317,14 +317,16 @@
               <a class="author-link" href="[[_getSuttaTextUrl(item.uid, translation.lang, translation.author_uid)]]">
                 <paper-button class="author-button" flat
                               title$="{{localize('goToTranslation', 'author', translation.author)}}">
-                    [[translation.author_short]]
+                  [[translation.author_short]]
                 </paper-button>
               </a>
             </template>
           </div>
-          <label class="translations-label">
-            {{localize('translations')}}
-          </label>
+          <template is="dom-if" if="[[_anyUserLanguageTranslations(item.translations, language)]]">
+            <label class="translations-label">
+              {{localize('translations')}}
+            </label>
+          </template>
 
           <div class="flex-div">
             <div class="languagedrop">
@@ -506,6 +508,10 @@
               return items.filter(item => item.lang === language);
           }
 
+          _anyUserLanguageTranslations(items, language) {
+              return items.some(item => item.lang === language);
+          }
+
           _toggleParallelsOpened() {
               this.parallelsOpened = !this.parallelsOpened;
               this.dispatchEvent(new CustomEvent('iron-resize', { composed: true, bubbles: true }));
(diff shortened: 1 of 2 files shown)


client/elements/suttaplex/sc-suttaplex.html
```diff
@@ -317,14 +317,16 @@ <h3 class="suttaplex-heading" title="[[_getMainHeadingTitle(item, localize)]]">
...
             </template>
           </div>
-          <label class="translations-label">
-            {{localize('translations')}}
-          </label>
+          <template is="dom-if" if="[[_anyUserLanguageTranslations(item.translations, language)]]">
+            <label class="translations-label">
+              {{localize('translations')}}
+            </label>
+          </template>
```
SamaneriVimala: Thanks for fixing this one! I noticed it before and had meant to look into it. Great that's it's done!

client/elements/suttaplex/sc-suttaplex.html
```diff
@@ -351,8 +353,7 @@ <h3 class="suttaplex-heading" title="[[_getMainHeadingTitle(item, localize)]]">
                   </paper-icon-button>
                   <sc-badge class="absolute-pos behind-button" number="[[item.parallel_count]]"></sc-badge>
                 </div>
-                <paper-tooltip for="parallels_toggle" class="sc-tooltip" offset="-1" fit-to-visible-bounds
-                               position="right">
+                <paper-tooltip for="parallels_toggle" class="sc-tooltip" offset="-1">
```
SamaneriVimala: There is a reason for this "fit-to-visible-bounds". Otherwise the tooltip sometimes disappears off the right side to the screen.

client/elements/suttaplex/sc-suttaplex.html
```diff
@@ -577,7 +584,7 @@ <h2 class="not-found-error">{{localize('itemNotFound')}}</h2>
                   });
                   return scAcronym.replace(/–\s*$/, '');
               } catch (e) {
-                  return;
+                  console.error(e);
```
SamaneriVimala: Why do you do a try-catch here and not in the same function in sc-parallel-list.html

client/elements/suttaplex/sc-suttaplex.html
TrebuhD: it seems that the combination of `fit-to-visible-bounds` and `position="right"` made it go haywire when the window wasn't wide enough to fit the tooltip, so I deleted both to make the tooltip appear bottom from the cursor.

client/elements/suttaplex/sc-suttaplex.html
TrebuhD: the try/catch was already here, I just printed the error. By the way, if the functions are exactly identical, we should definitely consider moving them to a separate element that would have utility functions to follow [https://en.wikipedia.org/wiki/Don%27t_repeat_yourself](DRY).

This `_transformId` function is an ideal candidate to move it to a separate utility class, because it's a pure function. Pure functions only depend on their arguments and don't have side effects. 
Example of a pure function: 
```
add2(x) {
   return 2 + x;
}
```
`add2` is a pure function, because when called with `x=1`, it will always return `3`, Similarly, `_transformId(rootId, expansionData)` will always return the same value for the same arguments and it doesn't use any external variables in the function body.

Pure functions are easy to maintain and don't break stuff when called.

SamaneriVimala: The functions are not entirely identical. There are a few small differences. But I guess nothing that cannot be overcome by adding a few attributes.

PR #549, opened 2018-02-18
Dragonslayer

SamaneriVimala:
I moved some functions that are used in both simple & segmented text files to a separate element.
Please allow me this little title for the function. It's an old joke for us.
I also added the quotehanger function to this so it is no longer double.


client/elements/addons/sc-text-functions.html
```diff
@@ -0,0 +1,96 @@
...
+                  }
+              }
+          }
+
+          connectedCallback() {
+              super.connectedCallback();
+          }
+
+          deathToTheBeast() {
+              const inputText = this.inputElement.innerHTML;
```
TrebuhD: it's better to pass `this.inputElement` as an argument instead of using `this` in the function body, this way the function doesn't touch variables outside of its scope.

client/elements/text/sc-segmented-text.html
```diff
@@ -101,7 +103,7 @@ <h3>{{ localize('couldNotFind') }}</h3>
        * @extends {Polymer.Element}
        * @appliesMixin Polymer.AppLocalizeBehavior
        */
-      class SCSegmentedText extends Polymer.mixinBehaviors([Polymer.AppLocalizeBehavior], ReduxMixin(Polymer.Element)) {
+      class SCSegmentedText extends SCTextFunctions {
```
TrebuhD: 1. We can make `SCSegmentedText` and `SCSimpleText` subclasses of another class, but I can see two issues here:
* `the SCTextFunctions` class mirrors the functionality of `SCStaticPage`, so maybe `SCTextFunctions` should extend `SCStaticPage` and `SCStaticPage` should then be renamed to something like `SCLocalizedElement`?
* Class naming is a bit off - a class name should be singular and the subclass should be a "subtype" of the parent - example: "Dog" or "Cat" extends "Animal". In this case I think `SCSegmentedText` could extend `SCText`.

SamaneriVimala: I'm a bit unsure of what you're saying here. I just created a class to take care of the doubling up of various functions that are used exclusively in texts.

client/elements/addons/sc-text-functions.html
SamaneriVimala: I tried that but if does not seem to like that. Maybe I just don't have my day today but the a11y element just wants to have a function without arguments and I do not see how else to pass that element to the function.

PR #564, opened 2018-02-22
some small adjustments

SamaneriVimala:
Several small adjustments and fixes


client/styles/sc-text-heading-styles.html
```diff
@@ -222,6 +222,17 @@
         padding: 0;
       }
 
+      #segmented_text_content .hgroup:after {
+        content: "NOTE: This is a draft. The proofreading is not complete.";
```
TrebuhD: Will this be internationalized? Why put it in :after?

client/styles/sc-text-heading-styles.html
```diff
@@ -222,6 +222,17 @@
...
+        background-color: var(--sc-secondary-background-color);
+        padding: var(--sc-size-md);
+        display: table;
```
TrebuhD: we should avoid using `display: table`. Flexbox should be enough

client/styles/sc-text-heading-styles.html
SamaneriVimala: Basically just what Sujato gave me to put in and no, it won't be internationalized: It's only on English pages anyway and only temporarily.

PR #571, opened 2018-02-23
fix pwa navigation on localhost IP address

TrebuhD:


client/service-worker.js
```diff
@@ -33,6 +33,6 @@ sw.router.registerRoute(
     sw.strategies.cacheFirst()
 );
 
-if (self.location.hostname !== 'localhost') {
+if (!self.location.hostname.match(/127.0.0.1|localhost/)) {
```
SamaneriVimala: My local environment has gone to 172.18.0.1 for some reason.

TrebuhD: That's weird. That seems to be an internal docker address. How did you discover it was the right IP for the site?

SamaneriVimala: It says in the terminal what address it's using.

TrebuhD: Ah! Yeah, that's obvious. I pushed the change, it should be OK now.

PR #577, opened 2018-02-25
revert "Add new build modes" for now

TrebuhD:


TrebuhD:
these new build modes are necessary for a PRPL pattern implemenation, but while they build fine locally, a weird `killed 41 polymer` appears on the next server and the page doesn't build. I'm deleting the extra ones for now.

PR #578, opened 2018-02-25
[SUCE-573] Suttaplex titles overlay each other

TrebuhD:


TrebuhD:
@SamaneriVimala, this is basically what we talked about in [the other PR discussion](https://github.com/suttacentral/next-sc/pull/575#discussion_r170451290) - this was also the source of this bug :) (my fault, too)!

client/elements/suttaplex/sc-suttaplex-list.html
```diff
@@ -93,7 +95,7 @@
       <iron-list id="suttaplex_list" class="suttaplex-list" items="{{suttaplexReturns}}" as="item"
                  scroll-target="document">
         <template>
-          <div style$="z-index: [[_getItemZIndex(suttaplexReturns, index)]];">
+          <div style$="z-index: [[_getItemZIndex(index)]];">
```
TrebuhD: So, `[[_getItemZIndex(index)]]` acts like a computed property - it'll get called once for every `suttaplexReturns` and `index` change.

But `suttaplexReturns` is also passed into `iron-list` in line 95. That means both these bindings get executed at the same time. I suspect this was causing a race condition - if `[[_getItemZIndex]]` was first, it stayed in the DOM against our will :).

What a true pain to debug this was. I guess this is why it's important to understand how these bindings work.

@SamaneriVimala @gradam

PR #582, opened 2018-02-26
[SUCE-577] send email confirmation

gradam:


TrebuhD:
This seems to double up the code from your previous commit.

PR #602, opened 2018-02-27
[SUCE-587] fix parallel ranges not working

SamaneriVimala:


client/elements/text/sc-segmented-text.html
```diff
@@ -700,6 +700,8 @@ <h3>{{ localize('couldNotFind') }}</h3>
 
           _processHighlightAndScroll(sectionId) {
               let [idFrom, idTo] = sectionId.split('--');
+              if (idFrom.match(/^[0-9]+$/)) { idFrom = `sc${idFrom}` };
```
gradam: please. don't use inline if

client/elements/text/sc-segmented-text.html
```diff
@@ -700,6 +700,8 @@ <h3>{{ localize('couldNotFind') }}</h3>
...
+              if (idTo && idTo.match(/^[0-9]+$/)) { idTo = `sc${idTo}` };
```
gradam: ditto

PR #613, opened 2018-02-28
[SUCE-593] open blurbs when no other items in list

SamaneriVimala:


client/elements/suttaplex/sc-suttaplex-list.html
```diff
@@ -238,6 +238,10 @@
               return this.suttaplexReturns ? (this.suttaplexReturns.length - index) : 0;
           }
 
+          _calculateOpened(suttaplexReturns) {
+              return (suttaplexReturns.length > 1) ? false : true;
```
TrebuhD: `return suttaplexReturns.length <= 1;`

SamaneriVimala: I made it <= 3. I figured that would include all the Abhidhamma texts from non-pali origin and would make more sense.

client/elements/suttaplex/sc-suttaplex-list.html
```diff
@@ -238,6 +238,10 @@
               return this.suttaplexReturns ? (this.suttaplexReturns.length - index) : 0;
           }
 
+          _calculateOpened(suttaplexReturns) {
+              return (suttaplexReturns.length <= 3) ? true : false;
```
TrebuhD: the ternary operator is redundant. `(suttaplexReturns.length <= 3)` already returns only two possible values - `true` or `false`.

PR #672, opened 2018-03-13
Suce 622 non english search

blake-sc:


server/server/data_loader/language_id.py
```diff
@@ -0,0 +1,75 @@
+import subprocess
+import langid
+import tempfile
+import random
+import pathlib
+import lxml.html
+
+def create_model(html_text_path, model_file, sample_size=20*1000*1000):
```
gradam: this function is super long. Maybe you can extract some code to additional functions

server/server/data_loader/language_id.py
```diff
@@ -0,0 +1,75 @@
...
+        for lang_dir in html_text_path.glob('*'):
+            if not lang_dir.is_dir():
+                continue
+            lang_uid = lang_dir.name
+            if lang_uid == 'zz':
+                continue
+            domain_lang_dir = domain_dir / lang_uid
+            domain_lang_dir.mkdir()
+            files = sorted(lang_dir.glob('**/*.html'))
+            rand = random.Random(42)
```
gradam: why 42?

server/server/data_loader/language_id.py
```diff
@@ -0,0 +1,75 @@
...
+            ['LDfeatureselect.py', 'corpus.model'],
+            ['scanner.py', 'corpus.model'],
+            ['NBtrain.py', 'corpus.model']
+        ]
+
+        # also altough langid is python3, the training tools are python2, good times...
+        for command in commands:
+            command[0] = str(train_dir / command[0])
+            print(f'running {command}')
+            subprocess.run(['python2'] + command, cwd=working_dir)
```
gradam: are you sure we have python2 installed in our docker containers?

server/server/data_loader/language_id.py
```diff
@@ -0,0 +1,75 @@
...
+        
+
```
gradam: there should be 1 blank line, not 2

gradam:
Tests on Travis are failing

server/server/data_loader/language_id.py
blake-sc: Well some kind of sampling is needed to make it faster to test, but it needs to be repeatable each time, sorting alone isn't good because then you'd only get `an` texts. So sorting + shuffle with predefined seed is a decent solution. And 42 is as good as any other seed.

server/server/data_loader/language_id.py
blake-sc: It does, but it doesn't contain numpy for python2. Truth be told this is super hacky, but building the model takes a super long time (about 2 hours if using all the texts rather than a sample) so it's not really a big deal to manually tweak the container before running it. It's also unimportant that an up to date model is used, we could never update the model again and it wouldn't really matter because the one I built already handles the languages which 99% of our users would use. It might make sense to run this code a couple of times in the next decade, so basically I don't care.

blake-sc:
By the way I don't intend to make Codacy happy and that warning about assert REALLY needs to be silenced.

PR #673, opened 2018-03-13
[SUCE-627] meta data in index

SamaneriVimala:
This fires an event that changes the meta data in index.html. It works well in Console but when I do "view-page-source' it changes it back to default settings so not sure how twitter/fb will react to this but we can try. If this fails, I have no idea.
Actually, if this works we don't need the Sherby-metadata element so I will take it out but lets test first.

### DIFF
--- a/client/elements/text/sc-text-page-selector.html
+++ b/client/elements/text/sc-text-page-selector.html
@@ -291,7 +291,9 @@
               document.dispatchEvent(new CustomEvent('sherby-metadata', {
                   detail: {
                       title: `${title}—${author}`,
-                      description: description
+                      description: description,
+                      bubbles: true,
+                      composed: true
                   }
               }));
           }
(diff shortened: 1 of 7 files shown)


client/elements/text/sc-text-page-selector.html
```diff
@@ -281,21 +281,19 @@ <h3>[[lastError.statusText]]</h3>
...
-                description: description,
-                'twitter:title': `${title}—${author}`,
-                'twitter:description': twitterDescription
-              };
+
+              document.dispatchEvent(new CustomEvent('sherby-metadata', {
+                  detail: {
+                      title: `${title}—${author}`,
+                      description: description
+                  }
```
TrebuhD: the customEvent will probably need `bubbles: true, composed: true` next to `detail` to penetrate the DOM to index.html.

PR #674, opened 2018-03-14
Suce 620 redirect legacy links

gradam:


client/elements/sc-page-selector.html
```diff
@@ -333,6 +336,18 @@
...
           }
 
+          _redirectFromLegacyLink() {
+            let parts = this.route.path.split('/');
+            parts.splice(1,1);
+            this.set('route.path', parts.join('/'));
+          }
+
+          _shouldRedirect() {
+              const langReg = this.allIsoCodes.join('|');
```
TrebuhD: `allIsoCodes` is not complete - the old site had 2-letter ISO codes where we have 3-letter ones. This can be easily added later though, just making sure we don't forget.

PR #678, opened 2018-03-14
[SUCE-620] redirect links

SamaneriVimala:


client/elements/sc-page-selector.html
```diff
@@ -343,14 +343,20 @@
...
             this.set('route.path', parts.join('/'));
           }
 
           _shouldRedirect() {
               const langReg = this.allIsoCodes.join('|');
-              const legacyLinkReg = new RegExp(`\/(${langReg})\/[a-zA-Z\d]+`);
-              return legacyLinkReg.test(this.route.path);
+              const legacyLinkReg = new RegExp(`\/(${langReg})\/.*[0-9]`);
+              const legacyLinkReg2 = new RegExp(`\/.*[0-9]\/(${langReg})`);
+              const newLinkReg = new RegExp(`\/.*[0-9]\/(${langReg})\/[a-z\d]+`);
```
TrebuhD: Hmm, this is all good but I feel this third regex is unnecessary - the two above could only respond to `/dn1/en` or `/en/dn1` but not `/dn1/en/sujato` for example.

SamaneriVimala: I tried it because I thought the same, but it did not work. I forgot exactly what it did wrong (would have to check again) and therefore I built in the exception to explicitly check for the newLink too.

PR #715, opened 2018-03-22
[SUCE-552] add uid expansion & author to search results

SamaneriVimala:


client/elements/sc-page-search.html
```diff
@@ -592,9 +588,41 @@ <h2 class="search-result-title">[[_calculateTitle(item)]]</h2>
...
+                  return '';
               }
-              return item.heading.title ? item.uid : '';
+              const expansionData = this.expansionReturns;
+              let scAcronym = '';
+              const uidParts = rootId.split('-');
+              let tail = '';
+
+              uidParts.forEach(item => {
+                  if (!expansionData[0][item]) {
```
TrebuhD: What will happen when `expansionData[0]` is `undefined` or an empty object (`{}`)?

client/elements/sc-page-search.html
```diff
@@ -592,9 +588,41 @@ <h2 class="search-result-title">[[_calculateTitle(item)]]</h2>
...
+                      const tailMatch = item.match(/\d+.*/g);
+                      if (tailMatch) tail = tailMatch[0] + '–';
+                      const itemMatch = item.match(/[a-z]*/g);
+                      if (itemMatch) item = itemMatch[0];
+                  }
+                  if (item && expansionData[0][item]) {
+                      scAcronym += `${expansionData[0][item][0]} ${tail}`
```
TrebuhD: are we sure that `expansionData[0][item]` is of type `array`?

client/elements/sc-page-search.html
SamaneriVimala: I've added a try/catch statement just in case but if the expansiondata is not loaded, it should not get this far.

client/elements/sc-page-search.html
```diff
@@ -592,9 +588,44 @@ <h2 class="search-result-title">[[_calculateTitle(item)]]</h2>
...
+                      }
+                      if (item && expansionData[0][item]) {
+                          scAcronym += `${expansionData[0][item][0]} ${tail}`
+                      } else {
+                          scAcronym += tail;
+                      }
+                  });
+                  return scAcronym.replace(/–\s*$/, '');
+              } catch (e) {
+                  console.error(e);
```
TrebuhD: right now, if there's an error somewhere in the `try` block, this function will return `undefined` and that's going to be displayed in the UI.

PR #717, opened 2018-03-22
[SUCE-480] Fix carousel on firefox

TrebuhD:


client/elements/addons/text-carousel.html
```diff
@@ -89,14 +90,26 @@
 
     </style>
 
+    <iron-ajax
+        auto
```
SamaneriVimala: is `auto` the best here?

client/localization/extract_static_strings.py
```diff
@@ -32,6 +32,8 @@ def replace_attrs(element, data):
 def hash_element_content(element, data):
     string = ''.join((str(child) for child in element.children))
     string = ' '.join(string.split())
+    if string == '':
+        return
```
SamaneriVimala: semicolon

client/elements/addons/text-carousel.html
TrebuhD: Good point. I'll check tomorrow if it hinders performance.

client/localization/extract_static_strings.py
TrebuhD: Python file! Fooled you ;)

PR #1063, opened 2018-06-25
[SUCE-916] fixing text caroussel

SamaneriVimala:
There were some inconsistencies in the code so those might account for the problems but not sure.
I cannot recreate this problem locally so have to look on Staging.


Aminah-SC:
In handover notes, Ven Vimila mentioned that although the caroussel now gives the right quote with the right link, sometimes the caroussel is empty when first loading the page.

PR #1202, opened 2018-12-05
Fixed loading pages while in background tab. Also fix for latest scro…

Bosmanfrx:
…lling fix. Sidebar should mostly behave itself


sujato:
Background tabs is good, sidebar still needs some work.

PR #1272, opened 2019-01-15
Offline notifications

janlat:


client/elements/addons/sc-error-icon.js
```diff
@@ -0,0 +1,52 @@
...
+      "no-network": {
+        icon: '/img/nonetwork.svg',
+        title: 'networkError',
+        message: 'offline'
+      }
+    };
+  }
+
+  static get properties() {
+    return {
```
robertmusil: I'd move static before constructor

PR #1285, opened 2019-01-16
Polymer 3

robertmusil:


client/package.json
```diff
@@ -54,13 +55,18 @@
...
     "raven-js": "^3.23.3",
     "redux": "^3.7.2",
     "viewerjs": "^0.10.0",
     "web-animations-js": "^2.3.1",
     "workbox-sw": "2.1.2"
   },
   "devDependencies": {
+    "babel-cli": "^6.26.0",
+    "babel-plugin-macros": "^2.4.2",
+    "htmlparser2": "^3.9.2",
```
agilgur5, line 68: ✅ `htmlparser2` seems to have not been further used -- I have removed it in #3567

client/package.json
```diff
@@ -3,11 +3,9 @@
   "version": "1.0.0",
   "private": true,
   "scripts": {
-    "serv": "polymer serve -p 8000 --compile",
-    "dev": "npm run serv | npm run watch",
+    "dev": "webpack-dev-server --env dev",
     "test": "polymer test",
-    "build": "polymer build",
-    "watch": "browser-sync start --proxy 0.0.0.0:8000 --files \"**/*.*, index.html, *.js\" --no-open",
```
agilgur5, line 10: ✅ `browser-sync` was no longer used after being removed here, but was left in the devDeps -- I have removed it in #3567

client/package.json
```diff
@@ -1,19 +1,80 @@
...
     "redux": "^3.7.2",
     "viewerjs": "^0.10.0",
+    "web-animations-js": "^2.3.1",
     "workbox-sw": "2.1.2"
   },
-  "devDependencies": {}
+  "devDependencies": {
+    "babel-cli": "^6.26.0",
+    "babel-plugin-macros": "^2.4.2",
+    "browser-sync": "^2.26.3",
```
jhanarato, line 68: RE: #3567

So PR #25 introduced the use of `browser-sync` but didn't install the dependency? Would it run?

agilgur5: The dep was installed globally in #25 in its Dockerfile with `npm install -g`: https://github.com/suttacentral/suttacentral/pull/25/changes#diff-117b27bbc2e3061c2064994beb2f8b7bc2f1ec23301a55455eee479e3b9cf526R8

So two different installations

jhanarato: Got it. `client/Dockerfile-livereload` was responsible.

PR #1311, opened 2019-01-24
Aria labels

robertmusil:


robertmusil:
They have a fallback and translations can always be added later

PR #1316, opened 2019-01-25
Add language pages

janlat:


Bosmanfrx:
It could use some of mobile tweaks. Some paddings and pie chart is not centered on mobile. Rest looks good to me :)

PR #1397, opened 2019-03-26
SUCE-1171 parallels display corrected

SamaneriVimala:


server/server/data_loader/arangoload.py
```diff
@@ -366,6 +366,20 @@ def generate_relationship_edges(change_tracker, relationship_dir, additional_inf
                             'resembling': any(x.startswith('~') for x in [first_uid, from_uid]),
                             'remark': remark
                         })
+                        try:
+                            to_nr = int(regex.findall(r'.*?([0-9]+)$', to_uid)[0])
+                        except BaseException:
```
blake-sc: It's dangerous to use BaseException or any non-specific exception, as for example that would also catch a regular expression compile error or a value error (such as to_uid being `None`) and silently ignore the error which can potentially create much debugging "fun", instead use the specific exception or exceptions, such as:
`except IndexError:`

It's also generally better form to not use an exception at all unless the case is actually exceptional, if it's something you expect to happen in normal cases use if-then-else structure. This isn't a hard and fast rule but it makes it very specific what condition is being checked:
```
m = regex.search('[0-9]+$', to_uid)
if m:
  to_nr = int(m[0])
else:
  to_nr = None # unless the number is actually zero, probably better to use None
```

It can also be done as a two liner using a ternary statement:
```
m = regex.search(r'[0-9]+$', to_uid)
to_nr = int(m[0]) if m else None
```

SamaneriVimala: The BaseException was used earlier in the code as well for the same thing - I just copied it.
So it should be changed in line 314 also.
The number indeed has to be 0 if all else fails.

PR #1516, opened 2019-10-08
Update html_text data loading in line with restructuring of pali_concord.json

chansikpark:
This is a minimal fix that adapts the new data format to the algorithm for importing PTS volpages.
Related change is suttacentral/sc-data@d669d888eb6f87decab83f1a82ab1de348ad7c19


sujato:
Chan, can you create an issue to demonstrate this issue more clearly, in particular:

- The change in the data that forces this
- a URL on the staging site we can refer to see the difference
- explanation of change made.

Also, if this is just a quick patch, which seems to be the case, then fine, but going forward the correct file is `all_pali_concordance.json`, which will completely replace `pali_concord.json`. So any substantial work should be on the new file.

chansikpark:
Yes, sir. I've created #1518.

The `load-data` task was failing in `textdata.py`'s `PaliPageNumbinator` class, which is used while importing legacy html texts in Pali into Arango. The code was using the old data format, in the form:

*Old Format*
`"1v:2:pts1": ["3", "1"]`

*New Format*
`"ms1V_2": "pts3.1"`

Strictly speaking, with the edition information going from the key into the value, the new format is not valid JSON because there are multiple different values specified in the dictionary for the same key. Conveniently, Python's JSON API had a way to handle this via an extra parameter to the `json.load` call.

I don't suspect much of a problem in updating the PR to use `all_pali_concordance.json` so that any dependence on `pali_concord.json` can be removed entirely. 

But overall, I get the feeling that there's a broader view that I'm missing here and am hoping to discuss a plan.

chansikpark:
Folded into topic branch:

https://github.com/chansikpark/suttacentral/tree/1518-html-volpages

Issue #1353, opened 2019-02-05
Some vol-page info in suttaplex nerdy row show in jeans-&-tshirt form


sujato:
## Description:

The nerdy row contains ID information for the sutta. This is sometimes displayed using the "jeans-&-tshirt" form of the ID (in `uid_expansion`, the `uid`), whereas it should be displayed in the "suit-&-tie" version (in `uid_expansion`, the `acro`).

I believe this was created when we fixed https://github.com/suttacentral/suttacentral/issues/1214

## Steps to reproduce:

examples:

https://suttacentral.net/pli-tv-bu-vb-pj

https://suttacentral.net/an3-patipadavagga 

https://suttacentral.net/an5-sammutipeyyala

## Expected behaviour:

**All IDs display in `acro` aka "suit-&-tie" form. A user never sees the `uid` form, except in the URL.**

## Screenshots:
![screenshot from 2019-02-06 09-39-42](https://user-images.githubusercontent.com/6112010/52309188-73e9ba80-29f3-11e9-9999-3e478768ceea.png)

chansikpark:
The `volpages` attribute returned by the query for suttaplex list is populated by the `volpage` attribute from the corresponding `root` collection object or from the first non-null `volpage` attribute of any corresponding objects from `html_text` or `po_strings` (see [queries.py](https://github.com/suttacentral/suttacentral/blob/20307d3d3c04765d53f9d00ee52f6a46bf3eee21/server/server/common/queries.py#L260)).

In the case of 'an5-sammutipeyyala', `an5.272` has a non-null `volpage` in the `root` collection whereas `an5.273-285` only has a non-null `volpage` in `po_strings` (in the entry for Pali). It's the same situation for the other two cases listed.

It would seem that there are at most four ways to address this:

1. Populate the `volpage` attributes in `root`.
2. Populate the `volpage` attributes in `po_strings` in "suit-&-tie" form.
3. Create a database mapping from "jeans-&-tshirt" form to all the possible "suit-&-tie" forms.
4. Implement some fail-safe mapping from "jeans-&-tshirt" to "suit-&-tie" in [views.py](https://github.com/suttacentral/suttacentral/blob/20307d3d3c04765d53f9d00ee52f6a46bf3eee21/server/server/api/views.py#L433).

[...]

PR #1523, opened 2019-10-13
Fix legacy html volpage-loading

### LINKED ISSUE #1518: Loading legacy html texts into Arango is broken for Pali (at least)
#### Description:

It appears that in the process of resolving #1094, a file containing mappings from Mahasangiti `ms` references to various other Pali editions was restructured (suttacentral/sc-data@d669d888eb6f87decab83f1a82ab1de348ad7c19 ff.). The file in question, `pali_concord.json`, is now made obsolete by `all_pali_concordance.json`.

However, `pali_concord.json` is still used by the `make load-data` task, specifically when determining the PTS volume-and-page references for legacy html texts in Pali. This results in the task failing when building the app from scratch, and ultimately a bunch of the texts aren't loaded. The code that uses `pali_concord.json` needs to be retired out of the main line (or made to fail more gracefully), or updated in line with the new structure.

#### Steps to reproduce:

0. Run data loading from scratch

1. Go to https://staging.suttacentral.net/bv

2. None of the texts are loaded:

#### Expected behaviour:

There should be "1 edition of the root text", multiple translations available, and there should be vol-page reference to a PTS edition.

#### Screenshots:

On staging, Buddhavamsa texts aren't loaded and have no vol-page references:

Screenshot from 2019-10-09 18-13-32

This is what their values should be:

Screenshot from 2019-10-09 18-14-30

chansikpark:
This resolves #1518 and replaces PR #1516. Details in commit comments.

Resultant volpages were manually tested within Arango against what they were before the first change to `pali_concord.json` by running the old algorithm against the old data and subsequently re-running the new algorithm against the new data.

Aside from changes resulting from (I assume) debaking (ie. in Jatakas, Sutta Nipata), the Mahasangiti references for the Cariyapitaka (eg "ms21C1" vs "ms21Cp_2") were different in a way that I wasn't sure were intentional or not. These references no longer show up in suttaplex.

Also, some of the pts-vp-pli references for Bv seem to have typos (eg. "pts-vp-pli1ed.17" vs "pts-vp-pli1ed1.17"). These show up in suttaplex without the volume number.

***

I should note that in the process of deprecating the previous PR #1516 and wanting to fold the development on this issue neatly into a topic branch, I've complicated the commit history unnecessarily.

This branch's ref was mistakenly pushed while its head was still pointing to the starting point of my master branch. Instead, it should have been rebased to point to the HEAD of the master branch before pushing. This has led to an unnecessary complication of what was already an unnecessarily complicated commit history.

I attach an image here that may or may not clarify the history. The commit nodes are labelled 1 through 9 to correspond to the 9 commits below.

[image removed by Aminah]

Note that this commit's parents are the eigth and third commits.

It should all just work, but auditing the history may be a bit taxing. If need be, I could create another account or something to maintain a cleaner fork.


sujato:
Thanks so much for this, it is *extremely* valuable. I would really appreciate it if you would look into these references as much as possible to ensure we resolve any outstanding issues.

With the Bv issue, I have fixed it.

With Cp, I believe I have fixed it, but it is not yet perfect. It seems the underlying issue here is that the source files registered the first segment in a series as either 1.1 or 1.1.1, which then appear as separate segments, whereas obviously they are the same. I have used a simple regex to resolve this, standardizing the ID forms as I do so. However, there are certainly still some duplicate references. It is expected that, since we are merging data from different sources, we will have off-by-one errors. Any deduping of these would be helpful. As a rule, if a segment ID is repeated on a subsequent segment, delete the second one.

chansikpark:
Glad to hear it, sir.

> I would really appreciate it if you would look into these references as much as possible to ensure we resolve any outstanding issues.

Unless advised otherwise, I'll start looking into the testing framework that's in place and see if there's a nominal way to maintain tests for data integrity with reference to all relevant issues. 

...and after the stop-gap on #1353, I'll also look at what the code might be like if it were integrated with bilara-data and the dependencies on pootle files were removed.

> With the Bv issue, I have fixed it.

Verified on my machine.

> With Cp, I believe I have fixed it, but it is not yet perfect.

I'll submit a PR in fork.

chansikpark:
Obsoleted by #1537

Issue #1504, opened 2019-09-20
Add Pali concordance data to bilara-data/reference


sujato:
Following completion of 

https://github.com/suttacentral/suttacentral/issues/1094

The various pali concordance references are now in one convenient file.

https://github.com/suttacentral/sc-data/blob/master/misc/all_pali_concordance.json

This should be imported to bilara-data/reference. The references are keyed off the bilara segment IDs and must be matched to them.

- On import, discard `sc` and `msdiv` numbers, they do not add any value.
- Import should be testy: reject any references that do not exactly match existing segment IDs. We can figure out why and fix it.
- These references will be used in the new text pages that Hongda is building.

chansikpark:
Continuing to work through cleaning up the concordances prior to import, and here proposing a tweak to the reference format, namely in regards to two editions–`bj` (Buddhajayanti) and `cck` (Chulachomklao).

In the migration from the original `pali_concord.json`, there were references containing slashes that were either flattened or dropped altogether. For `bj`, some of the volumes of this edition have two separate "parts" and the data migration appears to have just flattened references to multi-part volumes, which has resulted in hundreds of duplicate references. For example: `bj2/1.2` and `bj2/2.2`, which refer to two very different pages, seem to have become simply `bj2.2`, which is strictly speaking ambiguous because the second volume of the Buddha Jayanthi Tipitaka, is divided across two books, or parts, each having a page number 2. For `cck`, it appears that slash-containing references were dropped altogether. This is evidenced by there being no references to volumes 22 and 23 (which point to AN 5-9) in `all_pali_concordance.json` aot. `pali_concord.json`.

More formally, the original `pali_concord.json` file mapped the triple (Mahasangiti book id, Mahasangiti segment number, Edition id) to the couple (volume #, page #). Volume # is either an empty string, a simple number, or two numbers separated by a slash (eg. "22/2"). The last case occurs only in `bj` and `cck`, and was more or less dropped in the migration to `all_pali_concordance.json`.

***

The question is most generally: assuming that this wasn't intentionally done, how should it be resolved? More specifically, I can think of three different ways to disambiguate: which would be advisable?

Namely,

[...]

Issue #1474, opened 2019-08-29
Alternative vol/page numbers are not shown properly in suttaplex nerdy row


sujato:
#### Description:

In some cases, the PTS 9and possibly other) have published the same text in two editions, with different vol/age numbers. We have the data for both editions, and it should be displayed.

By default we show the *latest* edition. However reference works such as the PTS dictionary will often have the early edition.

Thus both editions should be shown as a tooltip when hovering. However currently this is broken: the tooltip displays blank data. Here is an example:

```
<span class="vol-page nerdy-row-element" title="PTS volume and page:
1st edition = {pts1}
2nd edition = {pts2}">
              SN i 64
            </span>
```

#### Steps to reproduce:
Go to:

https://suttacentral.net/sn1-adittavagga

Hover over vol/page on suttaplex.

#### Expected behaviour:

Where there are multiple editions, show the data on hover in suttaplex.

ihongda:
![深度截图_选择区域_20190903020406](https://user-images.githubusercontent.com/4037966/64135384-6b872000-ce1a-11e9-85d0-1090b805c407.png)

This bug is fixed.

sujato:
Unfortunately, I think this has been broken again by my recent updates to the uid-expansion. Sorry about that! 

In the past, we used `pts1` and `pts2` to mean "PTS first edition" and "PTS second edition". However this created errors, we also want to say `pts1.66` is "volume 1, page 66". So the first number after `pts` sometimes meant "edition" and sometimes meant "volume".

The current form of UID avoids this by using the form `pts1ed` and `pts2ed`. 

In addition, we now always specify the edition in full, that is it must always be `pts-vp-pli1ed` or `pts-vp-pli-2ed`.

The meaning of them is:

- `pts-vp-pli1ed`: "Volume and page of the first Pali edition of the Pali Text Society."
- `pts-vp-pli-2ed`: "Volume and page of the second Pali edition of the Pali Text Society."

These are found in `uid_expansion`:

https://github.com/suttacentral/sc-data/tree/master/misc

ihongda:
Okey, Bhante, I will adjust according to the latest situation 😄

chansikpark:
I see that this behaviour relies on volpages in the database being stored in "//"-separated format (eg. "SN i 32//SN i 66"). I believe currently, the only source of volpages that take advantage of this are in `sutta.json` and only for the first volume of SN. The other two sources of volpages is from pootle files and legacy html files.

[...]

Issue #1487, opened 2019-09-12
A Platonic ideal of data as yet undreamt on


sujato:
## History

SC began with a bunch of data: parallel tables, to be precise. Since the notion of a "sutta parallel" was fundamental to those tables, we made the sutta ID fundamental to how SC works. The scaffold of content is built up from this.

By now, most of SC's data is in well-formed and clean JSON. This is true, for example, of our `parallels.json`, a single file that contains all the thousands of parallels. The innards look like this:

```
[
    {
        "parallels": ["sag#1.1","sn1.20#28"]
    },
    {
        "parallels": ["sag#2.1","sn1.20#20","iti63#3"]
    },
```

Just sets of IDs related by JSON. The neat thing is  that when you have the ID, you can then hang anything off it; in fact the whole set of data known as a "suttaplex" is assembled this way.

 Here's another, more complex, example. From `languages.json` we have a list of the languages on SC and associated data.

```  {
    "uid": "pra",
    "name": "Prākrit",
    "is_root": true,
    "iso_code": "pra",
    "contains": [
      "pdhp",
      "pf"
    ]
  },
  {
    "uid": "uig",
    "name": "Uighur",
    "is_root": true,
    "iso_code": "uig",
    "contains": [
      "uf"
    ]
  },
```
So here we see a nice definition of the essential data about a language: Its "uid" (based on the ISO code), the name (which appears when spelled out in full), the distinction between root and translated languages, and a list of texts that are contained in that language. You can immediately see what everything is about. 

Not everything in our data files is so beautiful, though. In this post I shall amble through our data files and see what knotty bits or niggardly edges I might find. Perhaps this might eventuate in a new and better data approach! Perhaps not! Sometimes there is a reason things are the way they are; after all, it works. 

## Philosophy of data

Data should be:

- Consistent. We shouldn't use different names for the same thing.
- Simple. Keep each file to a single purpose.
- Obvious. You should be able to open a file and see what everything is.
- DRY. Don't repeat yourself.
- Single. One source of data for one kind of thing.
- Predictable. Simple exceptions trigger complex resolutions.
- Abstracted to the right level. Often it is not obvious what this is until afterwards.
- Fundamental. Our data files should just record the basic data, not application logic.

## Tree Structure

Our texts form a tree structure, so let us use the conventions of tree data.

https://en.wikipedia.org/wiki/Tree_(data_structure)

[...]

PR #1537, opened 2019-11-10
Volpage Fixes

chansikpark:
### Volpages in the front-end

Volume and page numbers into a physical edition ("volpage") is the preferred way to make reference to Pali texts in some academic contexts. Within Sutta Central, these references are supplied in four places (as determined by a project-wide case-insensitive search for "volpage"):

#### (1) Nerdy row of the Suttaplex card
Screenshot from 2019-11-11 06-23-52
#### (2) Nerdy row of a parallel item in the Suttaplex card (the hover needs some attention)
Screenshot from 2019-11-11 06-28-01
#### (3) Citation data copied to the clipboard using the Share function in the Suttaplex card
Screenshot from 2019-11-11 06-30-42
> Parallels for MN 2 Sabbāsava Sutta (MN i 6): AN 6.58 (AN iii 387); EA 40.6//T 125.40.6 (T ii 740a25); MA 10//T 26.10 (T i 431c13); T 31 (T i 813a05); Up 2.069 (mngon: ju 91b5 / tu 104b2); Mnd 16#217 (Mnd 444)
> Parallels for MN 2#2 Sabbāsava Sutta (MN i 6): Iti 102#2 (Iti 103); SN 12.23#1 (SN ii 29); SN 22.101#1 (SN iii 152); SN 56.25#1 (SN v 434); Kv 2.2#27 (Kv 173); Kv 2.3#16 (Kv 180)
> Parallels for MN 2#22 Sabbāsava Sutta (MN i 6): AN 4.114#9 (AN ii 116); AN 4.157#3 (AN ii 143); AN 4.165#3 (AN ii 153); AN 5.140#12 (AN iii 161); MN 125#22 (MN iii 128); pli-tv-bi-vb-pc71#2 (Vin iv 327); pli-tv-bu-vb-pc65#3 (Vin iv 129); pli-tv-kd1#439 (Vin i 1)
> Retrieved from http://localhost/mn2 on Mon Nov 11 2019 06:30:07 GMT+0900 (Korean Standard Time). 

#### (4) Right margin of text pages with textual information switched on
Screenshot from 2019-11-11 06-37-46

### Volpages in middleware

From the standpoint of the API, the endpoints that these four places use to populate volpage are:

(A) /api/suttaplex/<uid> (1)
(B) /api/parallels/<uid> (2),(3)
(C) /api/suttas/<uid> (4)

### Volpages in the database

In the case of (C), the volpages are embedded in markup strings which are found in Arango collections named `html_text` (original html markup) and `po_strings` (segmented pootle markup).

In the cases of (A) and (B), the volpages are an attribute of each Arango document from the same two collections as well as the `root` collection. These attribute-volpages represent the volume and page numbers in which the beginning of the associated text appears. When references into multiple manuscripts are available for a given text, the Suttaplex (cases (1), (2), (3)) expects the different volpage values to be concatenated with a "//". In reality, the current implementation essentially reserves "//" exclusively for Pali texts, specifically to Pali Text Society manuscripts with at most two different editions. This is to say that multiple non-Pali references and non-PTS references are unsupported. It may be preferrable going forward to revise the way volpages are stored and retrieved in order to handle multiples more robustly.

### Volpages in sc-data

The attribute-volpage values for the three collections are sourced from three different places.

The volpage attribute of `html_text` is sourced by parsing markup files in sc-data/html_text. Specifically, Mahasangiti references are plucked (eg "p_21Bu_2"), and these are converted into PTS references (suit-&-tie: eg "Bv i 1") using Pali concordance data. Currently, this applies to Khuddaka Nikaya (excluding the elder verses), Abhidhamma and Vinaya texts.

The volpage attribute of `po_strings` is sourced by parsing markup files in sc-data/po_strings. Here, the PTS references embedded in the markup are plucked and used directly (jeans-&-tshirt: eg "pts-vp-pli1.1"). Currently, this applies to the four Nikayas, the elder verses, and what appears to be much of the Vinaya.

The volpage attribute of `root` is sourced from `sutta.json`. These are in suit-&-tie format, and appears to cover all five Nikayas (imperfectly). The volpage in `root` is preferred, resorting arbitrarily either to `html_text` or `po_strings` if one isn't found there.

## Included fixes

***Updated***

Loading volpages into `html_text` is performed in `textdata.py`. Three notable changes are included:

1. An error during volpage loading would halt the entire process (solves #1518). Now, an error message is printed and loading continues.
2. Recent structural changes to the Pali concordance data required changes to adapt to the new structure (related to #1504).
3. When multiple PTS references were present, only the latest one was being used. ~~Now all of them are concatenated with "//" as expected by the front-end~~ Now all of them are returned in a map using the same keys as in `all_pali_concordance.json` (`pts-vp-pli` is assumed to mean `pts-vp-pli1ed`) (contributes to #1474).

Loading volpages into `po_strings` is performed in `po.py`. Two notable changes are included.

1. Volpages were being loaded in jeans-&-tshirt format as-is from markup. These are converted into suit-&-tie in similar manner to the conversion in `textdata.py` (solves #1353).
2. ~~"//" concatenation is adopted as opposed to the ", " concatenation that was there~~ Multiple references are returned in a map as above (contributes to #1474).

Note that the changes to `po.py` will be obsoleted by bilara. This PR replaces #1523.

Changes to suttaplex, suttaplex parallel item, suttaplex share menu were made to essentially gather all the volpage code into one place and to future-proof it a little for displaying non-PTS info.

## Miscellaneous notes

Volpages are also involved in search and text images.

In sc-data/structure/division, oddly, three dharmapadas are the only ones to include volpage (ie kn.json, sutta/dharmapadas/gdhp.json & pdhp.json).


sujato:
Thanks, these fixes look awesome and should substantially improve our handling of such cases.

Regarding references for multiple editions, the notation // is a holdover from our very earliest days and should be replaced with an appropriate JSON form.

Regarding the vol/page info found in the 3 Dhammapadas, I agree this is odd. Can we move this data to a better place? This relates indirectly to #1487.

chansikpark:
> Regarding the vol/page info found in the 3 Dhammapadas, I agree this is odd. Can we move this data to a better place? This relates indirectly to #1487.

For all of these, strictly speaking, the information is redundant since it can all be derived from `sutta.json`. The main difference is that the info in the division files are, well, per division (eg. dhp1-20 is given a vol/page reference as opposed to each of the verses having their own in `sutta.json`) And in fact, for `kn.json`, removing vol/page info improves the situation because then the suttaplex query falls back to retrieving the vol/page plucked from the markup files which has info for _both_ PTS editions.

For `pdhp`, `gdhp` and `g3dhp`, though, removing the information there would need to be accompanied either by modifications to `sutta.json` so that the respective entries are grouped up (eg. `pdhp1`,`pdhp2`... `pdhp13` => `pdhp1-13`), or the information would have to be added to the texts' markups along with a bit of code to extract it. Notably, all the `pdhp`, `gdhp` and `g3dhp` references are exactly just `Cone 1989`, `Brough 1962/2001` and `Falk 2015` respectively, without any page numbers.

The former would seem to be the correct thing to do, assuming that as per the markup, each vagga is to be treated as a single text, rather than each poem.

sujato:
Thanks yes I agree, best to add all reference info to the relevant JSON file, rather than in the text.

For now, we do keep reference information in the HTML legacy files, but it would be easy to envisage a future where this was no longer the case; we could extract it all as we do for the segmented texts. I wouldn’t want to make this a priority at the moment, our main focus is on getting the segmented texts working properly. However for the future it may be a useful improvement.

ihongda:
Thanks @chansikpark , 

I've reviewed the code for the front end, which looks good!
But I am not familiar with python, @blake-sc , can you help me to see the python code?

PR #1538, opened 2019-11-12
Add Chinese localization to navigation menu

ihongda:
Add Chinese localization to navigation menu.


firepick1:
It was very interesting to see how localization is done in LitElement. It is very similar to Vuetify.

ihongda:
Thanks, Karl!
-------------------
It's interesting how "this" is required for LitElement but not Vuetify/Vue.
--
PolymerElement also not need 'this' :smile:

PR #1540, opened 2019-11-13
Rewrite site components in LitElement (8): sc-page-dictionary.js

ihongda:
1. Rewrite sc-page-dictionary.js
2. Use Fetch instead iron-ajax


ihongda:
> You code is very readable with long names. If the lines get too long, sometimes I use abbreviations such as "dict" instead of "dictionary". Shorter lines are easier to read, but if the words get too short, that is also hard to read! No need to change anything, but feel free to experiment with what is easiest for you to read and work with. If it's easy for you to work with, it will probably be easy for others to work with.

Karl, Thank you for your valuable advice!

PR #1545, opened 2019-11-19
Rewrite SC component with litElement:  sc-lzh2en.js

ihongda:
1. Rewrite SC component with litelement: sc-lzh2en.js.
2. Using fetch instead of iron-ajax.


ihongda:
Thank you very much for your help, Karl!

PR #1546, opened 2019-11-20
Rewrite SC component with litElement: sc-pli.js

ihongda:
1. Rewrite SC component with litElement: sc-pli.js.
2. Using fetch instead of iron-ajax.


ihongda:
> I've never seen await await before.
> That's deeeep. :D

I believe that you have many good things worth learning for me :)

Issue #1448, opened 2019-06-09
Heal the sliced toolbar


sujato:
The static pages use a so-called "prominent" version of the toolbar, with extra height, and including the SC title and logo. When you scroll down, the "prominent" toolbar disappears and is replaced by a "standard" toolbar, as found elsewhere on SC. 

If you open up the Home page and scroll down, you'll see that the toolbar slices into two separate layers as it contracts. There's an unwanted shadow that meaninglessly divides the area. This sometimes creates unfortunate effects.

![Screenshot%20from%202019-06-05%2010-13-56|690x74](https://discourse.suttacentral.net/uploads/default/original/3X/1/4/149b10ce2b18fcbc4aa319bf3bf38d05113f9874.png) 

This is in fact just a leftover from the initial mockup, it was never meant to be like this. We just always had bigger priorities and never got around to fixing it.

The correct transition looks like this:

https://storage.googleapis.com/spec-host-backup/mio-design%2Fassets%2F1d3jb571H-QqU8rMvcvksU1e8-Di0lgu0%2Ftopappbars-behavior-scrolling-prominent-to-regular.mp4

Per Material design Guidelines:

https://material.io/design/components/app-bars-top.html#behavior

The only difference is that we do *not* shrink the title as per the demonstration. Rather, we use the site name and logo on the prominent toolbar, and replace with the page title on the standard toolbar. (See next item).

- **The transition between the "prominent" and "standard" toolbar should be smooth, with no jank or artifacts at any point.**

ihongda:
![Peek 2019-07-15 21-51](https://user-images.githubusercontent.com/4037966/61221332-e3a26500-a74a-11e9-9a41-bcbc18cc402f.gif)

sujato:
This issue is still not solved. Remember, the point of this is that the toolbar must behave in a way that is *elegant and organic*. It must appear in all cases as a **seamless and integrated whole**, not as a bunch of parts clanging together and interfering with each other. The toolbar is the primary thing that people see on SC, and it must work 100% in all cases.

Here is a screencast that shows a number of problems:

![Screencast from 03-08-19 09_50_11(1)](https://user-images.githubusercontent.com/6112010/62404404-59298480-b5d7-11e9-91d8-efe47181b4b9.gif)

1. When you scroll down, the top part of the toolbar covers part of the site title. **The site title must never be covered or clipped by another part of the toolbar.**
[...]

PR #1568, opened 2019-12-09
Attach the Adaptive Tab Bar to the standard toolbar

ihongda:
1. Attach the Adaptive Tab Bar to the standard toolbar
2. Rewrite sc-static-page-selector.js with litelement
3. Reduce horizontal padding for toolbar elements on small screens **when window resize**
4. Improve for : Heal the sliced toolbar #1448

### DIFF
--- a/client/elements/sc-page-selector.js
+++ b/client/elements/sc-page-selector.js
@@ -10,6 +10,7 @@ import { Localized } from './addons/localization-mixin.js';
 import './menus/sc-toolbar.js';
 import './text/sc-segmented-text.js';
 import './text/sc-simple-text.js';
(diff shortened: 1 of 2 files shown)


ihongda:
> I'm always a bit nervous seeing raw numbers in the code like "212". I like to replace them with const declarations that have comments. I know the constants are pre-existing, so you and I are both scratching our heads about what is "212" or "480" or "265". The thing about having numbers naked in the code is that the relationships between them are hard to understand. With comments and const declarations, future maintainers will feel more confident making changes.

Thanks, Karl, 
Indeed, hard code is not a good habit, and I will improve it later in the update.

Issue #1499, opened 2019-09-17
Revise objects in sc-simple-text.js


sujato:
This file contains the expansions for textual info in the plain HTML files. Since the new version will serve all Pali texts from bilara, we should **remove unnecessary names**, leaving only the ones below.

In addition, **check that the classes are still accurate**.

```
editionsExpansionData: {
        type: Object,
        value: {
          'ms': 'Mūlasarvāstivādavinayavastu, part 1-4',
          'dutt': 'Mūlasarvāstivādavinayavastu, part 1-4, Dutt (1939-1959)',
          'divy': 'Divyāvadāna (1886)',
          'ms84': 'Mūlasarvāstivādavinayavastu, part 1-4, (1984)',
        }
      },
```

PR #1573, opened 2019-12-23
Rewrite sc-simple-text.js component with litElement.

ihongda:
1. Rewrite sc-simple-text.js component with litElement.
2. Revise objects in sc-simple-text.js #1499
3. Fix: Pali lookup and Chinese lookup in firefox are invalid in the simple text page.


ihongda:
> I really don't understand deathToTheBeast. It puts in the character it rips out!?!?

This component is rewritten from sc-text-page.js It seems that this method is simply to change the 'ṃ' character. As for why to change it, I don't know yet :smirk:

ihongda:
Thanks Karl :smile:

PR #1576, opened 2020-01-15
Solve search field  responsive behaviour abnormal in firefox.

ihongda:
Solve search field  responsive behaviour abnormal in firefox.


ihongda:
Thanks Karl!

PR #1604, opened 2020-04-11
Refactoring and optimize sc-page-search.js

ihongda:
1. Refactoring and optimize sc-page-search.js
2. Use sc-bouncing-loader.js instead of paper-spinner-lite.js
3. Remove reference to iron-icon.js and iron-flex-layout.js
4. Modify sc-error-icon.js to increase error types


ihongda:
> Also, I recommend rethinking error handling when using async blocks. If the fetch fails, it fails silently I believe. Might be better to go with promise wrappers for async to handle failure cases gracefully. It's more coding but fewer random crashes.
> 
> let searchResult = await (await fetch(requestUrl)).json();

Thanks, Karl!
Indeed, although the current async blocks code is less, but not robust, I will gradually check and rewrite the relevant code!

firepick1:
Travis CLI has been broken for a month. https://travis-ci.org/github/suttacentral/suttacentral/builds/664246323

The latest pull request has nothing to do with the build error.
I guess we're not relying on Travis any more.

PR #1611, opened 2020-04-25
Use bouncing-loader instead of paper-spinner-lite.

ihongda:
Use bouncing-loader instead of paper-spinner-lite.


firepick1:
For a while I wondered why you deleted this:
`if change_tracker.is_any_file_new_or_changed(files):`
Then I saw you put it back. 😆

ihongda:
> For a while I wondered why you deleted this:
> `if change_tracker.is_any_file_new_or_changed(files):`
> Then I saw you put it back.

:smile: It's just a temporary modification for Bhante to deploy to his local computer.

Issue #1149, opened 2018-10-24
Implement thumbnail minimap for text pages


sujato:
It's quite common these days for text editors, following the example of Sublime, to feature a minimap, a small visual representation of the text in a file. This essentially enhances the role of a sidebar, allowing you to visualize where you are in a file and quickly click to go somewhere else. 

Research suggests that such functions aid reading by emulating some of the physical benefits of a paper book, allowing the reader to subconsciously keep place of where they are in a text.

https://www.baldurbjarnason.com/2018/09/03/neither-paper-nor-digital-does-active-reading-well/

We could consider implementing a minimap for texts, especially longer ones.

The coolest way to make a minimap, especially if you want to support Firefox only!

https://css-tricks.com/using-the-little-known-css-element-function-to-create-a-minimap-navigator/

- The user can turn the minimap on or off in the text settings menu. 
- On by default for wider screens, off for narrow screens

ihongda:
![深度截图_选择区域_20191212195556](https://user-images.githubusercontent.com/4037966/70710116-74d93000-1d19-11ea-93de-97a7a3f119b2.png)

It looks ok, but unfortunately only Firefox is currently supported.

sujato:
Okay, that looks good.

sujato:
I've done some more testing of the minimap addon, and it has too many bugs currently. We should remove it from the current roadmap and consider re-implementing at a future time. Note however that this functionality can be installed via a browser plugin.

1. Only spans a small vertical height, so not very useful. On long texts the image is too compressed to show anything. This is different from the reference implementation, which spanned full height.
2. The range-thumb appears outside the area of the imaged text. This bug is also not in the original.
3. Range-thumb does not scroll when scrolling the text, it just stays put. This bug is present in the original.

PR #1619, opened 2020-04-29
remove minimap

sujato:
I have removed the minimap, it has too many bugs and we need to focus on more important features. See here:

https://github.com/suttacentral/suttacentral/issues/1149#issuecomment-621509375


ihongda:
Yes, minimap is not very good yet, and it is not suitable for all browsers.

PR #1634, opened 2020-05-07
Use sc-top-sheet instead of sc-setting-menu.

ihongda:
Use sc-top-sheet instead of sc-setting-menu.


firepick1 (commented):
It's great to see localization for SC!

client/elements/addons/sc-top-sheet.js
```diff
@@ -0,0 +1,478 @@
...
+          type: 'TOGGLE_TEXTUAL_INFORMATION_ENABLED',
+          enabled: enabled
+        })
+      },
+      downloadParagraphs(data) {
+        store.dispatch({
+          type: 'DOWNLOAD_PARAGRAPH_DESCRIPTIONS',
+          descriptions: data
+        })
+      },
```
agilgur5, line 433: `toggleTextualInfo` and `downloadParagraphs` seem to have been copy+pasted from the `SettingsMenu` component, but are not used in this component and so are dead code here. I've removed these in #3714

PR #1636, opened 2020-05-09
Adjust sc-top-sheet according to test feedback.

ihongda:
Adjust sc-top-sheet according to test feedback.


firepick1:
I'm seeing a lot of pixel measurements in the stylesheet. I'd recommend use of <kbd>rem</kbd> and <kbd>em</kbd> to keep page layout consistent with different page sizes. Here is a nice article about when to use <kbd>rem</kbd> vs. <kbd>em</kbd>. The code you are working with is legacy, so I'm accepting this pull request. But remember this for solving future layout issues as they arise

PR #1685, opened 2020-07-16
New navigation system.

ihongda:
New navigation system.


ihongda:
Thanks to Karl, it's very important to have a good name for variables and methods. I still need to learn this aspect!

> You probably already know that "!important" is highly suspicious. Perhaps a simple comment might tell us why it is needed here?

This can be attributed to my laziness, because I didn't add! Important, CSS didn't work there, and I used it directly instead of seeking the cause of the problem. I'll check the root cause of the problem later :)

Issue #1664, opened 2020-05-28
Nav cards: steps (basic)


sujato:
By default, each nav card leads to the **next level** of the hierarchy. The link is the whole card.

The steps themselves are **identical with the current implementation in the sidebar**. The data for this is, I believe, in sc-data/structure.

For example, consider DN. here the hierarchy is: 

- DN
  - vagga
     - sutta

So the default navigation proceeds from the whole nikaya to the list of vaggas, to the suttas for each vagga.

Let us take imagine someone is navigating to DN 3.

First they are on https://staging.suttacentral.net/pitaka/sutta

>- DN
> - MN
>- SN
> - AN
>- KN

Then click on DN and see:

>- Silakkhandhavagga
>-  Mahā Vagga
>-  Pāthika Vagga

Then click on Silakkhandhavagga and see the 13 suttas in that vagga. Go to the sutta that you want!

Okay, now consider MN. here we have:

- MN
    - pannasa
        - vagga
            - sutta

It's exactly the same, except there is an extra level in the hierarchy, the pannasa (group of 50 discourses).

The default navigation proceeds from the whole nikaya to the list of pannasas, to the vaggas for each pannasa, to the suttas for each vagga.

As a final example, consider SN, which has perhaps the most deeply nested hierarchy of any collection.

- SN
   - Vagga
       - samyutta
           - (pannasa)
               - vagga
                     - sutta

We start with the five "big" vaggas, then the samyuttas for each vagga. Some of the longest samyuttas have pannasas, others do not. Then there is the "little" vagga as usual, and finally the sutta.

sujato:
This is not implemented.

sujato:
Reviewing latest changes, based on the same cases I tested above.

## DN

This is now incorrect. As stated in previous comment, it was fine then. But now there is another step where it goes from vagga to list of suttas, then to an individual suttaplex. This is incorrect.

![Screenshot from 2020-08-27 09-21-27](https://user-images.githubusercontent.com/6112010/91366387-ed036980-e846-11ea-80b3-323480d2dacf.png)

As stated above, there is no list of plain suttas. There is only the list of suttaplexes. This is correct.

![Screenshot from 2020-08-27 09-23-47](https://user-images.githubusercontent.com/6112010/91366499-3bb10380-e847-11ea-9f95-4173132c6e97.png)

## MN

Looks good!

## SN 

Looks good!

sujato:
[...]

PR #1703, opened 2020-09-01
Fix wrong navigation level.

ihongda:
Modify the navigation level according to the feedback of issue #1664.

### DIFF
--- a/client/elements/navigation/sc-navigation-common.js
+++ b/client/elements/navigation/sc-navigation-common.js
@@ -49,19 +49,23 @@ export const navIndex = new Map([
     pathParamIndex: -1,
     navArrayLength: 6
   }],
-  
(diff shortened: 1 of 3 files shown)


ihongda:
> Looks good!
> 
> The following is odd. I thought some() returns true or false, so why test?
> 
> ```
>   this.vaggaChildren.some(child => ['div', 'division', 'subdivision'].includes(child.type)) ? true : false; 
> ```
> 
> [developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some)

Thx, Karl!
When the type field of an item in the data list is equal to div or division or subdivision, it indicates that there is still sub item data, so I use some to check.

PR #1715, opened 2020-10-01
Modify new navigation system and others according to feedback.

ihongda:
0. Keep the navigation level of the navigation system consistent with the original sidebar menu level.
1. Use semantic HTML for header app-bar, including navigation.
2. Replace '❭' with mwc-icon: chevron_right.
3. The "spacing" toggle should only be displayed on pages where it affects.
4. Remove the "a" and "ripple" tags from the last item of breadcrumbs.
5. Remove the loading spinner that affects the user experience.
6. small style tweaks for breadcrumbs.
7. use last-child selector instead of class.
8. Fix essay link display error.
9. Fix the error title and place of the essay.
10. Using morph-ripple instead of paper-ripple.
11. Migration of static page menu and optimization of home page animation.


firepick1 (approved):
Everything looks solid. But I see islands of complexity that are tough to understand:
```
 ${this.navArray.length !== i + 1 ? html`
                <li @click=${() => this._navClick(nav)}>
                  <a href='${nav.url}'>${this.localize(nav.title)}<paper-ripple></paper-ripple></a>
                  <mwc-icon>chevron_right</mwc-icon>
                </li>` : html`<li class='lastLiPadding'>${this.localize(nav.title)}</li>`}
              ` : ''}
```
In Vue I would just put all that code into a method with lots of variable to explain what's going on. I'm not sure what LitElement allows.  Maybe in the future we can split up these monsters?

ihongda:
> Everything looks solid. But I see islands of complexity that are tough to understand:
> 
> ```
>  ${this.navArray.length !== i + 1 ? html`
>                 <li @click=${() => this._navClick(nav)}>
>                   <a href='${nav.url}'>${this.localize(nav.title)}<paper-ripple></paper-ripple></a>
>                   <mwc-icon>chevron_right</mwc-icon>
>                 </li>` : html`<li class='lastLiPadding'>${this.localize(nav.title)}</li>`}
>               ` : ''}
> ```
> 
> In Vue I would just put all that code into a method with lots of variable to explain what's going on. I'm not sure what LitElement allows. Maybe in the future we can split up these monsters?

This part is really obscure, and I'll see if there's a better way to express it. :smile:

PR #1725, opened 2020-10-10
Sujato work1

sujato:
This makes two changes. 

One is simple: add a space in the title for texts so there is a space on either side of the dash.

The second is more subtle. There was `will-change: transform` applied to `.transitionTransform`. This is a hint to the browser that the element will be transformed, and gives the browser time to optimize. 

However, this property is harmful if overused, as it prematurely optimizes and can end up wasting browser resources. As such, it is recommended that it is only used when necessary, and usually it is added and removed by JS. See [this article](https://dev.opera.com/articles/css-will-change-property/) for details.

For this reason, in my original demo, I added and removed this class via JS, see the note in `sc-page-selector-styles.js`. However now the class is hard-coded. 

One side-effect of this, which I cannot explain, is that it makes the animatable Home `titlebarSitetitle` appear above the more-menu, rendering the "Donations" item unclickable.

Screenshot from 2020-10-11 10-01-25

I don't know why `will-change: transform` has this effect. But in any case, removing it solves this problem.

It may well be the case that `will-change: transform` is unnecessary. The animations seem to work fine without it. However we should check on mobiles.

This PR solves the immediate problem, I remove `will-change: transform` and `titlebarSitetitle` no longer appears above the more-menu.

**But!**

My original note on `.transitionTransform` says:

> /* Add and remove via js to things that move. It is removed to prevent unwanted animations on zoom or page resize. */

Now that this class is hard-coded, we have the problem of unwanted animations on page zoom.

animate

I will make a new issue for this.

https://github.com/suttacentral/suttacentral/issues/1726

### DIFF
--- a/client/elements/styles/sc-site-layout-styles.js
+++ b/client/elements/styles/sc-site-layout-styles.js
@@ -314,7 +314,6 @@ export const SCSiteLayoutStyles = css`
   }
 
   .transitionTransform {
-    will-change: transform;
     transition: transform 200ms ease-in-out;
   }
 `;
--- a/client/elements/text/sc-text-page-selector.js
+++ b/client/elements/text/sc-text-page-selector.js
@@ -440,7 +440,7 @@ class SCTextPageSelector extends LitLocalized(LitElement) {
     if (!title) {
       title = this._transformId(this.suttaId, this.expansionReturns);
     }
-    this.actions.changeToolbarTitle(`${title}—${author}`);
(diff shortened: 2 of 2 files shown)


ihongda:
>I don't know why will-change: transform has this effect. But in any case, removing it solves this >problem.

I have been troubled by this problem for a few days. On the PC side, it is possible to remove will-change, but using a browser to simulate the mobile side still does not solve the problem.

PR #1728, opened 2020-10-15
Sujato header

sujato:
This makes some very small changes to the Home page, mainly the header.

- Use `<h1>` tag on Home page title to make it more semantic.
- Remove `<a>` tags from Home page title and subtitle. These are not links.
- Separate the layout and text styling CSS for the main heading.
- Adjust styles, remove unneeded CSS.
- Use proper `var()` for font-family.

In addition, I use flex properly to size contents of nav cards on Home page. Here you can see the problem that this solves.

Screenshot from 2020-10-15 19-29-27


firepick1 (commented):
You've been BUSY. 👍

PR #1760, opened 2020-11-12
Make user settings persist; remove "remember settings".

ihongda:
Make user settings persist; remove "remember settings" option from the top sheet.
Fix sc-navigation.js bug and remove unnecessary code.


firepick1 (approved):
Yay 2-space indents. 👍

PR #1772, opened 2020-11-18
 Add blurbs to nav cards.

ihongda:
Since getting the blurb related api returns a lot of data, the first display of the blurb will be slower, but I cached the blurb data to the localstorage, so the second display will be faster.

### DIFF
--- a/client/elements/navigation/sc-navigation-styles.js
+++ b/client/elements/navigation/sc-navigation-styles.js
@@ -2,11 +2,9 @@ import { html } from "lit-element";
 
 export const navigationNormalModeStyles = html`
   <style>
-
-
-      main {
-        max-width: 720px;
-        margin: 0 auto 64px;
+    main {
+      max-width: 720px;
+      margin: 0 auto 64px;
     }
 
     .main-nav {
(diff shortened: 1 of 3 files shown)


sujato:
Okay, we can look to optimizing this later on.

ihongda:
Not all blurbs are present. Here you can see the blurb for DN, but the others have no blurb.
---
I already know the problem and I fix it as soon as possible :)

ihongda:
>Not all blurbs are present. Here you can see the blurb for DN, but the others have no blurb.

Bhante, This problem has been fixed!

sujato:
Looking good, I can see most of the blurbs locally now.

There are still a few missing, though.

The blurbs in the Samyutta are often missing at the samyutta level, which may be a result of a bug in the navigation routing.

Screenshot from 2020-11-19 19-28-57

Sometimes the blurbs appear, sometimes not, and sometimes they both appear and do not appear!

Screenshot from 2020-11-19 19-33-19

ihongda:
Bhante, I've fixed a bug that doesn't show blurb in some cases.

However, the routing error has not been reproduced for the time being. I will check it again.

sujato:
okay, looks good.

firepick1 (approved):
Should I worry about "unsafeHTML"?
😄

ihongda:
> Should I worry about "unsafeHTML"?
> 😄

I checked out the data source of blurb and the only HTML tag I can see so far is <i>, so it should be safe. 😄

PR #1824, opened 2020-12-09
Feature/add field for display of notes in top sheet

### LINKED ISSUE #1809: Add field for display of notes in top sheet
The top sheet must have a display for notes. This was present in the original mockup, but has gotten lost along the way.

We support two kinds of notes in the segmented texts, which are stored in the respective directories in bilara-data. These two kinds of notes are identical, except:

- Variant readings are found in root texts.
- Comments are found in translations. 

Variant readings are already present in the text. 

Comments are not yet present as they are not finalized for publication (hence the comments folder is only in the "unpublished" branch). Nevertheless, we must ensure support for comments as they will be ready to use shortly. We should be able to simply add the comments folder to the "published" branch and everything works.

The original spec is here (much has changed since!)

https://github.com/suttacentral/suttacentral/issues/1496

You can see the basic spec for the notes. Updating it a little, what we need is this. This is just the skeleton, obviously it must be localized and use `mwc` etc. In this list the **second** option is on by default. 

Also don't forget to adjust the CSS grid for the extra field.
```
<details><summary>Show notes and variants</summary>
        <p>Where available, view comments on the translation 
           and variant readings for root texts.</p></details>

- [ ] None
- [X] Tooltip on asterisk
- [ ] Show as sidenotes
``` 

More on the implementation of the notes is here.

https://github.com/suttacentral/suttacentral/issues/1483

So far as I know, they should work just fine, it's just a matter of switching them on and off and changing the CSS to toggle tooltip and sidenote view.

### LINKED ISSUE #1818: Reorganize top-sheet files, they are messy
The top sheet on text pages is now implemented and mostly works. However it is not organized well.

The code for the info and parallels top-sheets lives inside /suttaplex/card/. But there is no intrinsic relation between the top sheet and the suttaplex. It just happens that one of the top sheets *includes* a suttaplex. This is confusing, and we should separate suttaplex from top-sheet.

Then there is `sc-top-sheet.js`in /addons/. This handles the top sheet for the "Views". 

- **Let us put all the files for top-sheet in /addons/**

Also, we should rename these files consistently:

```
sc-top-sheet-common.js
sc-top-sheet-parallels.js
sc-top-sheet-views.js
sc-top-sheet-info.js
```

It's usually a good idea to name the file the same thing as you see in the UI.

Prograd:
resolves #1809
resolves #1818

### DIFF
--- a/client/elements/addons/sc-top-sheet-views.js
+++ b/client/elements/addons/sc-top-sheet-views.js
@@ -10,9 +10,9 @@ import '../addons/sc-bouncing-loader';
 import '../addons/sc-toasts';
 
 import { store } from '../../redux-store';
-import { LitLocalized } from '../addons/localization-mixin'
+import { LitLocalized } from './localization-mixin'
 
-class SCTopSheet extends LitLocalized(LitElement) {
+class SCTopSheetViews extends LitLocalized(LitElement) {
   static get properties() {
     return {
       selectedTextView: { type: String },
@@ -613,4 +613,4 @@ class SCTopSheet extends LitLocalized(LitElement) {
   }
 }
 
-customElements.define('sc-top-sheet', SCTopSheet);
+customElements.define('sc-top-sheet-views', SCTopSheetViews);
--- a/client/elements/addons/sc-top-sheet-views.js
+++ b/client/elements/addons/sc-top-sheet-views.js
@@ -225,7 +225,7 @@ class SCTopSheetViews extends LitLocalized(LitElement) {
         background-color: var(--sc-secondary-background-color);
         box-shadow: var(--sc-shadow-elevation-4dp);
       
-        grid-template-columns: 240px 360px 240px 240px 240px 780px 360px;
+        grid-template-columns: 240px 240px 360px 240px 240px 240px 780px 360px;
       }
       
       .tools {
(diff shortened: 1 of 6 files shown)


sujato:
I've tested these changes locally and they are working great.

client/elements/addons/sc-top-sheet-views.js
```diff
@@ -369,6 +369,28 @@ class SCTopSheet extends LitLocalized(LitElement) {
...
+        <details><summary>${this.localize('noteSummary')}</summary>
+        <p>${unsafeHTML(this.localize('noteDescription'))}</p></details>
+        <div class="form-controls">
+          ${this.noteDisplayTypeArray.map(({displayType, displayTypeLabel}) => html`
+            <mwc-formfield label="${this.localize(displayType)}">
+              <mwc-radio
+                name="noteDisplayType"
+                value="${displayType}"
+                data-type="${displayTypeLabel}"
+                ?checked="${this.selectedNoteDisplayType === displayType ? true : false}"
```
robertmusil, line 384: You can simplify it to `{this.selectedNoteDisplayType === displayType}`

Prograd: too late, it's already merged. I can change it during developing the other ticket but to be honest it should be catch be linting tool like eslint, I've seen the same in the other getters.

Issue #1806, opened 2020-12-02
Load `name` and `tree` data to ArangoDB


sujato:
## Background

- Data flow: https://github.com/suttacentral/suttacentral/wiki/Data-flow-for-Bilara-and-SC
- `tree` and `name` files: https://github.com/suttacentral/suttacentral/issues/1714

## Task

- The data in the `tree` and `name` files must be loaded to the database via `arangoload.py`. 

These files are located here:

https://github.com/suttacentral/sc-data/tree/master/structure

## Required for

Once the data is in ArangoDB it will be queried for the navigation API.

https://github.com/suttacentral/suttacentral/issues/1781

## Note

Currently these files include names in root language only. The translations are currently being made in Bilara. We will add the translated names at the next step.

alexander-vo:
As we discussed

Now all data are loaded from `name` files from in the `sc-data/structure/name/` folder and from the `uper-name.json` file. All data are uploaded to a collection called `super_nav_details` in our Arango database.
All relations between the data from the previous paragraph are described in `tree` files from the `sc-data/structure/tree/` folder and from the file `super-tree.json`. These data are also uploaded to the collection `super_nav_details_edges` and use the fields `_from` and `_to` to describe relations between the files in the `super_nav_details` collection.

Document structure from the `super_nav_details` collection:
```
{
    'uid': '',
    '_key': '',
    'name': '',
    'root_lang': '',
    'volpage': '',
    'biblio_uid': '',
    'acronym': ''
}
```
Where:
`uid` - unique id of the document, keys from `name` files (dn, dk, etc.)
`_key` - same as uid, unique id in the database
`name` - ame of the document, values from `name` files
`root_lang` - document source language, could be found in the `language.json` by document `uid`
`volpage`, `biblio_uid`, `acronym` - could be found in the `super_extra_info.json` by document `uid`. If data couldn't be found in this file, then we are trying to search this in the `text_extra_info.json` in the same way.

sujato:
Alex, for `root_lang` use instead the new file `super_root_lang.json` as discussed here.

https://github.com/suttacentral/suttacentral/issues/1821

The reason is that we do not want to list the root lang for every text, but only at the places where it is useful. That's what is found in the list above. This gives us the correct `root_lang_iso` and `root_lang_name` for the places where it is needed.

alexander-vo:
@sujato done in the latest commit, document structure description changed too

sujato:
[...]

PR #1828, opened 2020-12-10
1806 load name and tree data to arangodb

alexander-vo:
#1806 issue resolved
Details [here](https://github.com/suttacentral/suttacentral/issues/1806#issuecomment-742526566)


alexander-vo:
@blake-sc @robertmusil Hey guys, should we use some models or classes in our application? It seems to me that `dicts` are very obscure data structures here...

sujato:
I'm all for having clear data structures. Do you have any proposal for a better way to do it?

blake-sc:
> @blake-sc @robertmusil Hey guys, should we use some models or classes in our application? It seems to me that `dicts` are very obscure data structures here...

I think it makes sense in terms of how ArangoDB has a central role in the application and generally I would like python to do the bare minimum required to get the data into ArangoDB, and then view-side the python should do as little as possible to get the data from ArangoDB to the client (ArangoDB AQL queries are very powerful for generating complex JSON and I've found almost always preferable to writing Python code).

Once upon a time SuttaCentral was entirely a python application, and we used classes a lot more at that time, but at the moment most of what the python is doing is working with JSON data. It might be the case that it would be best to use some kind of dataclass to make it clear that this isn't arbitrary data which could be stored in a python dict but rather should be as strict subset that serializes to JSON, though I haven't investigated this possibility in detail.

server/server/data_loader/arangoload.py
```diff
@@ -83,14 +83,28 @@ def collect_data(repo_dir: Path, repo_addr: str):
...
+    Method to process super_extra_info.json and text_extra_info.json files
+
+    Args:
+        extra_info_file - path to the file
+    """
+    info = json_load(extra_info_file)
+    data = {}
+    for item in info:
+        uid = item.get('uid')
+        data.update({uid: item})
```
blake-sc, line 107: I am unsure why you are using `data.update({uid: item})` instead of simple assignment `data[uid] = item` or a dict comprehension.

blake-sc (commented):
Generally looks good.

server/server/data_loader/arangoload.py
alexander-vo: Indeed, thanks!
I missed that.
Fixed in the next commit!

blake-sc (approved):

server/server/data_loader/arangoload.py
```diff
@@ -183,6 +194,100 @@ def process_division_files(
...
+                'uid': uid,
+                '_key': uid,
+                'name': name,
+                'root_lang': root_languages.get(uid, None),
+                'volpage': extra_info.get(uid, {}).get('volpage', None),
+                'biblio_uid': extra_info.get(uid, {}).get('biblio_uid', None),
+                'acronym': extra_info.get(uid, {}).get('acronym', None)
+            }
+
+            docs.append(entry)
```
Prograd, line 233: Seems quite complex, also I can see it take the responsibility of filter function and you've done here some kind of mapping. I'm not the python guy but maybe you can use [map](https://docs.python.org/3/library/functions.html#map) and [filter](https://docs.python.org/3/library/functions.html#filter) instead of raw loop.

server/server/data_loader/arangoload.py
```diff
@@ -183,6 +194,100 @@ def process_division_files(
...
+            if type(item) == dict:
+                edges.extend(
+                    [{'_from': name, '_to': subgroup_name} for subgroup_name in item.keys()]
+                )
+                edges.extend(parse_tree_recursive(item))
+            else:
+                edges.append({
+                    '_from': name,
+                    '_to': item,
+                })
```
Prograd, line 257: The same here you can use map function.

server/server/data_loader/arangoload.py
```diff
@@ -250,16 +355,38 @@ def add_root_docs_and_edges(change_tracker, db, structure_dir):
...
+        nav_details_docs = process_names_files(
+            names_files,
+            root_languages,
+            super_extra_info,
+            text_extra_info
+        )
+
+        super_tree_edges = process_super_tree_file(super_tree_file)
+        nav_details_edges = process_tree_files(tree_files)
+        nav_details_edges.extend(super_tree_edges)
```
robertmusil, line 388: I'd probably do `process_tree_files(super_tree_file, tree_files)`, just to keep the logic of generating nav_details in one place, but this one is fine too.

server/server/data_loader/arangoload.py
```diff
@@ -183,6 +194,100 @@ def process_division_files(
...
+
+    Returns:
+        list of processed data
+    """
+    docs = []
+    names_files.sort(key=lambda path: len(path.parts))
+    for name_file in names_files:
+        entries: Dict[str, str] = json_load(name_file)
+        for uid, name in entries.items():
+            if not name or type(name) != str:
```
robertmusil, line 220: how can `type(name)` not be a string? is that a data problem?

server/server/data_loader/arangoload.py
robertmusil: In python you'd totally use for loops for that

Prograd: Is there any reason to not use map and filter function?

server/server/data_loader/arangoload.py
alexander-vo: unfortunately yes
for example this file: 
https://github.com/suttacentral/sc-data/blob/2606d251c73ec72e3212ed3ea92dbff4c41f8f45/structure/name/root/abhidhamma/pr-name.json#L2

but then we could found such uid here:
https://github.com/suttacentral/sc-data/blob/2606d251c73ec72e3212ed3ea92dbff4c41f8f45/structure/name/root/super-name.json#L201

that's why we need such "check"
maybe you have a better idea? :thinking:

server/server/data_loader/arangoload.py
alexander-vo: actually, I was thinking about it, next time I will create a separate function, thanks!

server/server/data_loader/arangoload.py
alexander-vo: Well actually we can, yes, but personally I don't think it's the "Pythonic way" to use the `map` function here. I'd rather use `map` if I had to change every element of the incoming list in some way and want a list/generator of changed objects in the output, although even then I'd consider using list comprehensions or generator expressions (though again - it all depends on the case).
but personally I think that in this case regular loops are more readable and don't cause any discomfort at all

Issue #1826, opened 2020-12-09
Add acronym and sutta ranges to navigation-nerdy-row


sujato:
We should have acronyms with sutta range numbers where appropriate in the navigation nerdy row. These show the user where the suttas they need are.

For example, they may see "Chapter on Similes" but they don't know if the sutta they are looking for is in there. So we add "MN 31–40" to help.

This is similar to the implementation in the old SuttaCentral, where these numbers are displayed in the sidebar. Now, however, we have more room so we can do it better.

## Use child_range.json!

The other details below are probably unnecessary now. Just use `child_range.json`, it tells you exactly what IDs to supply the info. The info is complete and does not require any additions or transformations: just add the literal strings and we're good to go.

## Implementation

An initial implementation has been made by Hongda here:

https://github.com/suttacentral/suttacentral/pull/1823/commits

These comments aim to correct this.

Start with DN. Actually it's perfect, great, move on.

Now, with MN we have a bug. At the pannasa display (http://localhost/pitaka/sutta/middle/mn) we have the acronym `MN` for every case. 

- Here we should show the range: `MN 1–50` and so on. 

Next we have the vagga level (http://localhost/pitaka/sutta/middle/mn/mn-mulapannasa)

Here  we see the name of the pannasa in the acronym range. Eg `Mulapannasa 1–10`. This is incorrect, we **always** use the acronym of the **collection**. 

- So this should be `MN 1–10` etc.

Next to SN. Let's see the Vaggasamyutta level: http://localhost/pitaka/sutta/linked/sn

Here we have acronym `SN`. 

- This should be, eg. `SN 1–12`, etc.

Then to the individual samyuttas (http://localhost/pitaka/sutta/linked/sn/sn-sagathavaggasamyutta). 

- These should say `SN 1`, `SN 2`, etc.

Where there are pannasas, they too are incorrect: http://localhost/pitaka/sutta/linked/sn/sn-khandhavaggasamyutta/sn22

- They should be e. `SN 22.1–50` etc.

Then to the vaggas: http://localhost/pitaka/sutta/linked/sn/sn-sagathavaggasamyutta/sn1

- These should say, `SN 1.1–10`, `SN 1.11–20`, etc.

Now to the Anguttara. The nipata level is correct, great. http://localhost/pitaka/sutta/numbered/an

But the vagga level is not. http://localhost/pitaka/sutta/numbered/an/an1

- It should be `AN 1.1–10`, etc.

[...]

PR #1832, opened 2020-12-11
#1826 Load child_range into the database

alexander-vo:
child ranges from from the sc-data/structure/child_range.json file will be loaded to the child_range collection in the ArangoDB

### DIFF
--- a/server/server/data_loader/arangoload.py
+++ b/server/server/data_loader/arangoload.py
@@ -8,6 +8,8 @@ from collections import defaultdict
 from itertools import product
 from pathlib import Path
 from typing import Any, Set, List, Dict
+
+from arango.database import Database
 from flask import current_app
 from git import InvalidGitRepositoryError, Repo
 from tqdm import tqdm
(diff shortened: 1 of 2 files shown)


alexander-vo:
Collection state:

image

sujato:
Looks good to me.

server/server/data_loader/arangoload.py
```diff
@@ -720,6 +740,9 @@ def print_stage(msg):
     print_stage("Building and loading navigation from structure_dir")
     add_root_docs_and_edges(change_tracker, db, structure_dir)
 
+    print_stage("Loading child ranges from structure_dir")
+    load_child_range(db, structure_dir)
+
```
Prograd, line 745: Do we need this empty line? It seems to me like unnecessary change.

server/server/migrations/migrations/add_child_range_collection_027.py
```diff
@@ -0,0 +1,13 @@
...
+from migrations.base import Migration
+
+
+class SecondMigration(Migration):
+    migration_id = 'add_child_range_collection_027'
+    tasks = ['create_collections']
+
+    def create_collections(self):
+        db = get_db()
+
```
Prograd, line 11: The same here, I guess you can remove it.

server/server/data_loader/arangoload.py
robertmusil: Why? It adds nice separation of logical blocks.

Prograd: If you say so, then it's. I usually avoid such such separator.

server/server/migrations/migrations/add_child_range_collection_027.py
alexander-vo: Well, yes, you can do it without skipping a line, although it is only used to visually separate logic
but I also don't think it's a big deal to reopen the pull-request for that 
you can also look at other migration files and see that "annoying" line skip there =D

PR #1843, opened 2020-12-16
1781 | Menu API redesigned for the new data

alexander-vo:
+ MENU and SUBMENU queries changed
+ removed po_strings collection usage in the v_text view
+ removed unnecessary code from Menu view
+ updated get_data method
+ make_yellow_brick_road returns list with updated values
+ has_translated_descendent moved to the body of the Menu class
+ _cache parameter removed from has_translated_descendent, @cache.memoize decorator used instead

Things todo:
+ add more support for the `language` parameter after adding translations for documents


alexander-vo:
**Important note!!!**
The language parameter does NOT translate menu item names with the `language` parameter now. This pull request is made to not delay @ihongda in his work. Additional support for this parameter will be added after the translations will be loaded into the database in ticket #1806

server/server/migrations/migrations/add_text_search_view_025.py
```diff
@@ -19,12 +19,6 @@ def create_view(self):
                             'lang': {'analyzers': ['identity']},
                         }
                     },
-                    'po_strings': {
```
robertmusil, line 22: You shouldn't edit migrations ever, this has already executed on staging, you should create a new one that removes the field.

server/server/api/views.py
```diff
@@ -241,194 +224,80 @@ def get(self, submenu_id=None):
...
 
-    def get_data(
-        self,
-        submenu_id=None,
-        menu_query=MENU,
-        submenu_query=SUBMENU,
-        language=None,
-        bind_vars=None,
-    ):
+    def get_data(self, submenu_id: str = None, language: str = None) -> List[dict]:
```
robertmusil, line 257: how long does it take to generate a menu with empty cache?

robertmusil (changes requested):

server/server/api/views.py
alexander-vo: If we try to get the full menu (like `/api/menu`) the first time (or without cache) it can take almost 2-3 seconds and loads instantly with cache
Same with submenu like `/api/menu/{submenuId}`
The first time takes < 1 sec, and is almost instantaneous with cache

robertmusil (approved):

PR #1846, opened 2020-12-17
NOTICKET: Added prettier settings & git hook

robertmusil:
This will add consistent formatting for all the users out of the box (as long as they run `make prepare-host`)

After merging this one more PR is going to be required, one person will need to run `npx prettier -w .`  in `client/` folder, which will in turn format all the client files. I didn't do that in this PR for clarity of changes. Person running the changes will also need to walk through the app to check whether it runs properly as well.


sujato:
Looks good, please go ahead

ihongda:
:+1:

PR #1847, opened 2020-12-17
1556 | Table of Contents for segmented and legacy texts

Riwuko:


client/elements/sc-site-layout.js
```diff
@@ -346,8 +356,9 @@ class SCSiteLayout extends LitLocalized(LitElement) {
       }
       let displaySettingMenu = store.getState().displaySettingMenu;
       let displaySuttaParallels = store.getState().displaySuttaParallels;
+      let displaySuttaToC = store.getState().displaySuttaToC;
```
Prograd, line 359: Why you use let instead of const? You're not changing the reference or value of it. You should use const as much as it possible, stateless code is easier to understand.

client/elements/text/sc-bilara-segmented-text.js
```diff
@@ -93,6 +92,12 @@ class SCBilaraSegmentedText extends SCLitTextPage {
     this.translationTextSelector = '.translation .text';
     this.rootTextSelector = '.root .text';
     this.commentSpanRectInfo = new Map();
+    
+    this._hashChangeHandler = () => {
+      setTimeout(() => {
+        this._scrollToSection(window.location.hash.substr(1));
+      }, 0);
```
Prograd, line 99: You can omit the second parameter it's the default value. Btw why you use setTimeout is there any reason for that?

client/elements/text/sc-bilara-segmented-text.js
```diff
@@ -140,11 +144,26 @@ class SCBilaraSegmentedText extends SCLitTextPage {
       <sc-bottom-sheet></sc-bottom-sheet>
     `;
   }
-
+  
   firstUpdated() {
+    // location-changed event is added by iron-location component.
+    // iron-location disables default behavior of hash-links and re-emits a different event.
+    // hashchange can still fire, but only due to browser back & forward buttons.
+    // TODO: In case iron-location is removed, this will break, but it's not possible to bypass it now
```
Prograd, line 152: If there is any issue which you're not able to fix right now, create an issue and add here at least an gihub id.

client/elements/text/sc-bilara-segmented-text.js
```diff
@@ -179,6 +204,20 @@ class SCBilaraSegmentedText extends SCLitTextPage {
   _shouldShowLoadingIndicator() {
     return ((!this.error && this.loading) || this.isTextViewHidden);
   }
+  
+  // Scrolls to the chosen section
+  _scrollToSection(sectionId, margin = 120) {
+    if (!sectionId) return;
+    try {
+      let targetElement = this.shadowRoot.querySelector(`#${CSS.escape(location.hash.substr(1))}`);
```
Prograd, line 212: The same issue here, use const.

client/elements/text/sc-bilara-segmented-text.js
```diff
@@ -179,6 +204,20 @@ class SCBilaraSegmentedText extends SCLitTextPage {
   _shouldShowLoadingIndicator() {
     return ((!this.error && this.loading) || this.isTextViewHidden);
   }
+  
+  // Scrolls to the chosen section
+  _scrollToSection(sectionId, margin = 120) {
+    if (!sectionId) return;
```
Prograd, line 210: To be honest I prefer adding brackets instead of returning nothing.

client/elements/text/sc-bilara-segmented-text.js
```diff
@@ -384,18 +423,20 @@ class SCBilaraSegmentedText extends SCLitTextPage {
...
+      this.actions.showToc([]);
       return;
     }
     const dummyElement = document.createElement('template');
     dummyElement.innerHTML = this.markup.trim();
     let arrayTOC = Array.from(dummyElement.content.querySelectorAll('h2')).map(elem => {
       const id = elem.firstElementChild.id;
-      if (sutta.strings[id]) {
-        return { link: id, name: this._stripLeadingOrdering(sutta.strings[id]) };
+      if (sutta[id]) {
```
Prograd, line 433: What is under the sutta in this context? Maybe there is a better name for that.

client/elements/text/sc-bilara-segmented-text.js
```diff
@@ -384,18 +423,20 @@ class SCBilaraSegmentedText extends SCLitTextPage {
...
+        return { link: id, name: this._stripLeadingOrdering(sutta[id]) };
       }
     });
     arrayTOC = arrayTOC.filter(Boolean);
-    return arrayTOC;
+
+    this.actions.showToc(arrayTOC);
   }
```
Prograd, line 440: You should not add structure name in variable name, it should be an abstract name.

client/elements/text/sc-simple-text.js
```diff
@@ -151,6 +154,11 @@ class SCSimpleText extends SCLitTextPage {
     this.localizedStringsPath = '/localization/elements/sc-text';
     this.currentId = '';
     this.inputElement = {};
+    this._hashChangeHandler = () => {
+      setTimeout(() => {
+        this._scrollToSection(window.location.hash.substr(1));
+      }, 0);
```
Prograd, line 160: The same question and suggestion here.

client/elements/text/sc-simple-text.js
```diff
@@ -282,14 +308,30 @@ class SCSimpleText extends SCLitTextPage {
...
     this._showHighlightingChanged();
     this._referenceDisplayTypeChanged();
+    this.navItems = this._prepareNavigation();
+    setTimeout(() => {
+      this._hashChangeHandler()
+    }, 100);
+  }
+
+  _prepareNavigation() {
+    let sutta = this.sutta["text"];
```
Prograd, line 322: let

Prograd (changes requested):

sujato:
Thanks for the review! 

Also, the ToC should be generated from `<h2>` as per the reference implementation on the old site:

https://suttacentral.net/dn1/en/sujato

I should have specified this in the original ticket, sorry!

Currently the ToC is generated from `<h3>`. Only `<h2>` should be considered.

sujato:
Oh, one more thing, the `<ol>` inside the `<nav>` element had a class. there's no need, just leave it as a plain `<ol>` element. One of the benefits of ShadowDOM!

Issue #1835, opened 2020-12-12
Remove the paper dropdown


Prograd:
#### User story
This is a part of bigger issue #1310.

#### Feature description
Remove the [paper dropdown](https://www.webcomponents.org/element/@polymer/paper-dropdown-menu) dependency form the project. Use [material menu](https://github.com/material-components/material-components-web-components/tree/master/packages/menu) instead.

**The scope has been change a bit. The main idea is the remove the dependency from the more and language menu. Currently, it works correctly but there are a few issues which needs to be address.**

#### Known issues check:
- [x] the list of languages should be modern languages only i.e. not Pali, Sanskrit, etc.
- [x] sc-language-base-menu should have the same width as the main more menu, the transition would be smoother.(**It's done partially, the min and max width is set properly but scrollbar do not follow this value**) 
- [ ] specify overflow-y: scroll; on the sc-language-base-menu (not the more menu).
- [x] There's a weird positioning bug if you scroll down the page and open the menu, it is not in place.
- [x] We have both id="more-menu" and id="more_menu". We should use more clear naming
- [ ] The ripple effect on the switches is bounded, even though it is set to unbounded. This is because .mdc-list-item__text is set to overflow: hidden. Again, fix it if it is an easy fix, else leave it.

#### Acceptance criteria (the list of things that need to be done for the ticket to be considered finished):

- Affected components works in the same way as previously, the design as well
- paper dropdown dependency is removed from the project 

#### Done check:
- [x] Project builds without errors
- [ ] Peer Code Review performed/pull request approved 
- [ ] Project deployed on the stage environment identical to production platform
- [ ] Feature is tested against acceptance criteria
- [ ] Feature ok-ed by Product Owner (moving to Closed on the Board)
- [ ] Refactoring completed

Prograd:
I'm not able to to find the usage of **sc-search-filter-menu**. I've seen that it is imported to the **sc-settings-menu** but I haven't seen any usage of it. We can remove it if doesn't use anymore.

ihongda:
Hey, Jakuba, It is also used in sc-page-search.js.

sujato:
The `sc-search-filter-menu` is just a simple dropdown selector which lacks the rich functionality of the toolbar menu. It is rarely used used (as the correct search results should float to the top anyway). TBH I have never used it. I'd like to get rid of it, but that would require a deeper overhaul of search.

[...]

PR #1864, opened 2020-12-29
Feature/menu with mwc component

Prograd:
#1835


Prograd:
@sujato Please do not merge it instantly, let's wait at least for @robertmusil approval.

sujato:
testing locally, mostly it is good, a few issues remain:

- the ripple on the main menu has a wrong color (blue), the one on the language chooser is correct.
- the switches for theme and toolbar don't work.

Apart from that, I can only see the minor issues of the width and the switch ripple overflow, both of which can be left if there is no obvious fix.

sujato:
Everything seems good except the ripple effect, which is still blue. We should replace `morph-ripple` with `mwc-ripple`, this will just work.

(FYI, we are using `morph-ripple` only because previously there was another context where `mwc-ripple` was not working. But maybe they've fixed it, anyway it works fine here.)

I've made a PR with this fix. Also I fixed a small issue with vertical alignment of menu items.

ihongda (approved):
It's nice to be able to phase out the use of the paper components.

client/elements/menus/sc-language-base-menu-css.js
```diff
@@ -1,85 +1,101 @@
...
-  .language-menu-paper-item {
-    font-size: var(--sc-skolar-font-size-md);
-    color: var(--sc-primary-text-color); 
-    /*19px for the icon, 16px for the margin */
-    --paper-item-icon-width: calc(var(--sc-size-language-icon) + var(--sc-size-md));
-  }
+    .language-menu-paper-item {
+      font-size: var(--sc-skolar-font-size-md);
+      color: var(--sc-primary-text-color);
+      /*19px for the icon, 16px for the margin */
```
robertmusil, line 45: this comment is likely to be totally inaccurate, since those are vars and not absolute values

client/elements/menus/sc-language-base-menu.js
```diff
@@ -85,15 +78,17 @@ class LanguageBaseMenu extends LitLocalized(LitElement) {
...
+      this.selectedLanguageNum = this._findChosenLanguageIndex(event.target.id);
       const chosenLanguage = this.languageListResponse[this.selectedLanguageNum];
       // If it's not a main language change menu (but a clone), dispatch an event
       if (this.cloneName) {
-        dispatchEvent(new CustomEvent(`${this.cloneName}-language-changed`, {
-          detail: { isoCode: chosenLanguage.iso_code, name: chosenLanguage.name },
-          composed: true,
-          bubbles: true
-        }));
+        dispatchEvent(
```
robertmusil, line 85: I know it's just formatting here, but I'm pretty sure error throws here, shouldn't it be `this.dispatchEvent`?

client/elements/menus/sc-more-menu.js
```diff
@@ -170,32 +96,45 @@ class SCMoreMenu extends LitLocalized(LitElement) {
...
     this.menuCreated = false;
     this.localizedStringsPath = '/localization/elements/sc-more-menu';
+    this.languageListResponse = [];
     this.routeName = store.getState().currentRoute.name;
     this.alwaysShowUniversalToolbar = store.getState().alwaysShowUniversalToolbar;
+    this.languageIsVisible = store.getState().languageMenuVisibility;
+  }
+
+  connectedCallback() {
+    super.connectedCallback();
```
robertmusil, line 114: can be removed since it doesn't do anything

client/redux-store.js
```diff
@@ -186,6 +186,8 @@ const reducer = (state, action) => {
       return ({ ...state, navDataCache: action.navDataCache });
     case 'CHANGE_ALWAYS_SHOW_UNIVERSAL_TOOLBAR_STATE':
       return ({ ...state, alwaysShowUniversalToolbar: action.alwaysShowUniversalToolbar });
+      case 'CHANGE_LANGUAGE_MENU_VISIBILITY_STATE':
+      return Object.assign({}, state, { languageMenuVisibility: action.languageMenuVisibility });
```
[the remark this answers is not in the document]
Prograd: Sure, I haven't notice that the convention here has been changed.

client/elements/menus/sc-language-base-menu-css.js
Prograd: I'm wondering now if this css is still need because it was for paper item and which as I remember we don't use anymore.

client/elements/menus/sc-language-base-menu.js
Prograd: I'm not able to check whether it's work, cloneName is not set anywhere. Looks like legacy code, I think we can remove this dispatch at all. @ihongda  Do you know anything about this event? Is it still used?

client/redux-store.js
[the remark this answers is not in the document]
Prograd: Changed in 07643a2f

client/elements/menus/sc-language-base-menu-css.js
Prograd: Removed in 07643a2f

client/elements/menus/sc-more-menu.js
Prograd: Removed in 07643a2f

PR #1866, opened 2020-12-31
remove morph ripple, fix vertical alignment of items

sujato:
- fix ripple
- fix vertical alignment


sujato:
also remove sentry.js. No-one's ever looked at it and it adds maybe 70kb of js.

PR #1875, opened 2021-01-08
Sujato topsheet select

sujato:
- use native select
- remove mwc-select
- fix bug with more-menu appearing behind top-sheet. This has removed the z-index from top sheet. It seems okay, but best watch for any unintended side-effects.


sujato:
also fix issues with color and ripple on menu select

ihongda (approved):

PR #1876, opened 2021-01-11
fixes to toc

sujato:
this fixes two out of three of the outstanding ToC issues

### DIFF
--- a/client/elements/text/sc-simple-text.js
+++ b/client/elements/text/sc-simple-text.js
@@ -49,10 +49,6 @@ class SCSimpleText extends SCLitTextPage {
           margin-top: var(--sc-size-xl);
         }
 
-        .highlight {
-          background-color: var(--sc-disabled-text-color-opaque);
-        }
-
         article p,
         .word,
         .translated-text,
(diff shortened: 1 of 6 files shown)


sujato:
Also add proper lang to subtitle on tipitaka home page

client/elements/text/sc-simple-text.js
```diff
@@ -326,7 +322,7 @@ class SCSimpleText extends SCLitTextPage {
     }
     const dummyElement = document.createElement('template');
     dummyElement.innerHTML = sutta;
-    let arrayTOC = Array.from(dummyElement.content.querySelectorAll('h3')).map((elem, index) => {
+    let arrayTOC = Array.from(dummyElement.content.querySelectorAll('h2')).map((elem, index) => {
```
robertmusil, line 325: Are you sure about this change? I was pretty sure toc elements were h3 on the examples we've seen

sujato: No, the suttas are properly structured, they start with h1, then the subheadings are h2 etc. Only a few texts have h3 or deeper, and there the list of headings usually becomes so long there's not much point. (Some texts actually go down to h6!) 

Keep it simple, h2 only.

PR #1878, opened 2021-01-11
Use new pali script converter.

ihongda:


firepick1 (approved):

client/elements/addons/sc-top-sheet-views.js
agilgur5, line 27: `paliScripts` was removed and replaced by `scriptIdentifiers` in the constructor and in the `paliScriptsTemplate` function, but it was left as a property in the component here.

I've removed it in #3738

PR #1884, opened 2021-01-13
Supply the YBR numbers at every level if possible.

ihongda:


firepick1 (approved):
I don't like data in code. But we can fix it later.

  if (!['sutta', 'vinaya', 'abhidhamma'].includes(this.pitakaUid)) {

Issue #1780, opened 2020-11-21
API 3: Get dictionary information


sujato:
## URL

```
/api/dictionaries/lookup?from=[langiso]&to=[langiso]
```

e.g.

```
https://staging.suttacentral.net/api/dictionaries/lookup?from=pli&to=en
```

## Description 

Get the dictionary of the relevant language. 

## Parameters 

| Name | Data Type | Description |
| --- | --- | --- |
| from | String | e.g. `pli`, `lzh` ... |
| to | String | e.g. `en`, `lzh` ... |

 ## Data format returned: JSON 

Data source is here. This should be added to sc-data/dictionaries

https://github.com/suttacentral/suttacentral/files/4934360/ncped.zip

## Other description 

At present, this API already exists, but the old data source needs to be replaced by a new data source. Please refer to the relevant issue for details. 

## Related issue 

https://github.com/suttacentral/suttacentral/issues/1508

PR #1887, opened 2021-01-14
# 1780 | API 3 get dictionary information

alexander-vo:
- #1780 done
- remove old dictionaries collections and usages
- added indexes to dictionaries collections
- separate dictionaries views from other views
- update load-data script with new dictionaries loading
- dictionaries migrations added
- queries updated


server/server/api/views/dictionary.py
```diff
@@ -0,0 +1,50 @@
...
+        from_lang = request.args.get('from', None)
+
+        if from_lang is None:
+            return {'message': 'from not specified'}, 422
+
+        db = get_db()
+        data = list(
+            db.collection('dictionaries_simple').find({'to': to_lang, 'from': from_lang})
+        )
+        return data
```
robertmusil, line 32: Seems like the only place you don't add the return code

robertmusil (approved):

server/server/api/views/dictionary.py
alexander-vo: oh, thanks!

Issue #1906, opened 2021-01-17
Nav cards inconsistently apply `blurb:empty`


sujato:
When there is no blurb, there is still the containing `<div>` . 

- (Is this the best way? Wouldn't it be better to only add the `<div>` when needed?)

This has margin and padding applied which leaves a visible gap in the interface. 

![Screenshot from 2021-01-17 16-31-10](https://user-images.githubusercontent.com/6112010/104831969-9ecc3100-58e1-11eb-9fe3-05ab768558a3.png)

This should be removed by `blurb:empty`.

![Screenshot from 2021-01-17 16-33-51](https://user-images.githubusercontent.com/6112010/104832001-e6eb5380-58e1-11eb-9ea9-ba059d3ad9fa.png)

But it is not always applied.

So :

1. **Either**: Only add `<div class='blurb'>` when there is actually a blurb
2. **Or** make sure `blurb:empty` is always applied.

Issue #1893, opened 2021-01-15
In nav cards, show root language when it is useful


sujato:
Currently we don't explicitly show the root language in the nav cards.

In this commit I make a naive first attempt at doing so:

https://github.com/suttacentral/suttacentral/commit/a30ff0352ea5e455427b9aada0627580d8bd966e

But this shows the language on every card. It should only do so at the top level of the collection to be defined.

This can be easily done using the data from sc-data/structure/language.

The logic is as follows:

- **If** a UID is  included in ` "contains"`
- **Then** add class `show-root-language` to nerdy-row `subTitle`.
- Put your feet up and relax, job well done! (I'll make it work with CSS)

ihongda:
>If a UID is included in "contains"

Bhante, What does that mean?

sujato:
In the `contains` field here:

https://github.com/suttacentral/sc-data/blob/master/structure/language.json

sujato:
This is working well in the latest commit.

PR #1908, opened 2021-01-18
Only add <div> when there is actually a blurb;show root language when it is useful.

ihongda:
1. Only add `<div class='blurb'>` when there is actually a blurb. #1906

2. In nav card, show root language when it is useful. #1893


sujato:
Running this locally and it looks good.

PR #1910, opened 2021-01-18
Feature/donate form

### LINKED ISSUE #1614: Migrate to Stripe Checkout for payments
Stripe has updated their payment method. The new way is to do the payment on Stripe itself, so we just have a button that sends data to stripe. No need to worry about forms, submissions, security, accessibility, and all that. Yay!

Stripe has always set great standards for both UX and security, so this looks really cool.

https://stripe.com/docs/payments/checkout

https://stripe.com/docs/payments/checkout/migration

## How it will work

So far as I understand, the basic donation page will remain unchanged.

https://suttacentral.net/donations

However, when you click the "donate" button, it will take you to Stripe to fulfill all the details. We set it up on Stripe, and that's it. 

We still need to load `stripe.js` though, so the tracking issue remains:

https://github.com/suttacentral/suttacentral/issues/1613

If we cannot remove this, then we must put a statement informing users.

## Security

We have recently been getting a lot of fraudulent donations via bot, and need to do something about it. Updating to the latest version of stripe is obviously a start.

We have been in contact with Stripe about this, and they recommend implementing reCaptcha. However, if more security is needed I would rather avoid using a Google product. [hCaptcha](https://www.hcaptcha.com/) would be better.

However, all captcha services are poor UX and easily hacked. A newer approach is using "proof of work". What this does is give your browser a problem to solve before the form is submitted. This creates a computational cost that is, in theory, unobtrusive for genuine users but prohibitive for bots, which would have to solve these problems at a massive scale.

https://friendlycaptcha.com/

Another approach is to use the new Web Authentication API:

https://webauthn.guide/

https://w3c.github.io/webauthn/

https://webauthn.me/

https://caniuse.com/?search=webauthn

## Stripe/lit integration

I'm not sure if this is relevant, it seems to use the older stripe method?

https://github.com/bennypowers/stripe-elements/tree/master

Prograd:
close #1614
We don't have success page yet. I think it'll be worth to wait for that until we merge it.


server/server/api/views/views.py
```diff
@@ -669,147 +670,60 @@ def get(self):
...
+            currency = body['currency']
+            amount = body['amount']
+            frequency = body['frequency']
+
+            secret_key = os.environ.get('STRIPE_SECRET')
+            stripe.api_key = secret_key
+
+            incoming_uri = urlparse(request.url)
+            cancel_url = '{uri.scheme}://{uri.netloc}/donate-now'.format(uri=incoming_uri)
+            success_url = '{uri.scheme}://{uri.netloc}/thank-you'.format(uri=incoming_uri)
```
robertmusil, line 684: Hmm. Is `uri.netloc` validated somehow? I.e. like Django does "ALLOWED_HOSTS"? Otherwise this is susceptible of header poisoning.

server/server/api/views/views.py
```diff
@@ -669,147 +670,60 @@ def get(self):
...
-                    description=f'''Donation by {name if name else ""}, 
-                                message {message if message else ""}''',
+        body = request.get_json()
+        if body is not None and all(item in list(body.keys()) for item in ['currency', 'amount', 'frequency']):
+            currency = body['currency']
+            amount = body['amount']
+            frequency = body['frequency']
+
+            secret_key = os.environ.get('STRIPE_SECRET')
+            stripe.api_key = secret_key
```
robertmusil, line 680: Why not assign it in a single line?

Prograd: The previous code done it like this. As a result, I did the same. I'll change it.

server/server/api/views/views.py
Prograd: tbh I don't know. Should we really bother it? It's only plain link. If it's a problem, it should be fix in separate PR.

Prograd: Maybe I can take the domain form some kind of configuration. I'll check it.

server/server/api/views/views.py
Prograd: Fixed in 8abd8af

PR #1911, opened 2021-01-19
Use sc bilara data names for navigation

alexander-vo:
- change the data source for navigation API
- use languages.json for the misc names


sujato:
I've pulled the latest version and I am still seeing invalid language definition. `null` is not a language. The tag should only define a language if there is a language to define.

Screenshot from 2021-01-19 22-02-11

This should either say:

```
<span class="subTitle" translate="no" lang="pli">Dīgha</span>
```

Or if you can't get the language data, then:

```
<span class="subTitle">Dīgha</span>
```

But no language is ever defined in HTML or JSON as `"null"` The JSON should be:

```
"lang": null
```

Not 

```
"lang": "null"
```

alexander-vo:
@sujato honestly, I don't understand what we're talking about right now.
As far as I understood this data from the API menu
the data about the language for navigation is taken from the super_root_lang.json and if no language was found for the given uid - it will have a null value.
I think there's something I don't fully understand

sujato:
Okay, so the JSON data for the language is fine. (The volpage is incorrect, though, it should be `"volpage": null` not `"volpage": "null"`.)

Therefore there is a bug in how this data is getting from here to the front end. Let's bring @ihongda in.

Just to be clear, let me explain.

- the `lang` attribute in HTML defines a language
- `null` is not a language. In JSON it means "no language is defined". In HTML it does not mean anything.
- `<foo lang="null">` is invalid and must never occur.
- if the JSON says `"language": null` then there must be no `lang` attribute in the HTML: `<foo>`
- That's it.

alexander-vo:
@sujato actually `"null"` as a string is taken from supeer_extra_info.json

image

sujato:
Okay, I'll fix it there.

sujato:
The names and languages appear to be correct in ArangoDB.

sujato (commented):
Names and languages appear top be correct in DB.

PR #1912, opened 2021-01-20
Fixing suttaplex API

alexander-vo:
- relationship edge collection now watch super_nav_details rather then root
- removed po usage for suttaplex
- `root`, `root_names`, `root_edges` usages replaced by `super_nav_details`, `names` and `super_nav_details_edges`
- new type definitions


sujato (commented):
The suttaplex page is now showing the correct names. Problem is, that's _all_ it's showing!

Screenshot from 2021-01-21 07-13-59

alexander-vo:
That's interesting
@ihongda can you check?

ihongda:
> That's interesting
> @ihongda can you check?

Alex, I've found the reason
When the value of the type field changes from text to leaf, it cannot be displayed.
I will modify the corresponding front-end code :)

alexander-vo:
Great news @ihongda !
Merge these small changes into the branch and we can send them to the master

PR #1913, opened 2021-01-20
Remove `division` related code.

ihongda:


sujato (commented):
Everything looks good.

alexander-vo:
@ihongda Have you checked where the collections that are marked here (`divisions` and `text_divisions`) are used? 
Have you tried to replace them with something, or to make sure that accessing them does not cause errors? 

@robertmusil is it legal to delete migration files like this?
and do we need to create a migration to remove unused collections (`divisions` and `text_divisions`)?

ihongda:
@alexander-vo Alex, I've checked the code and there's no code used the `text_divisions` collection at the time, but I'll check it out again. :)

server/server/migrations/migrations/add_divisions_010.py
```diff
@@ -1,20 +0,0 @@
...
-
-        paragraphs = db.create_collection('divisions')
-        paragraphs.add_hash_index(fields=['uid'], unique=True)
-
-    def create_text_divisions_collection(self):
-        db = get_db()
-
-        paragraphs = db.create_collection('text_divisions')
-        paragraphs.add_hash_index(fields=['uid'], unique=True)
-        paragraphs.add_hash_index(fields=['division'], unique=False)
```
alexander-vo, line 20: In general, I am not sure that we should delete the migration file, as far as I know, it is not quite right :thinking:

ihongda: There doesn't seem to be any code that references the two collections at the moment, but we can keep it and delete it later when we're sure it's no longer needed.

alexander-vo: Thanks for bringing him back, he didn't deserve this fate =D

it's not that we need it now, it's just bad practice, maybe later I can say more
personally, I think we need to know what was created when, in what form, for what and who created it + may need additional migration in the future to remove collections that were created here

alexander-vo (approved):

PR #1920, opened 2021-01-21
Use new data sources for dictionaries.

ihongda:


sujato:
Hongda, you may be aware of this, but just in case! 

I'm using this latest branch, and I can see the new dictionary in the lookup, so that's great. There are some issues with CSS, but I'll take care of them when you are done.

Meanwhile, in terms of content, the markup is sometimes wrong when dealing with complex entries. See here, for example, the entry for *ekaṁ*.

Screenshot from 2021-01-22 18-02-29

The words in the definitions collide together. This is because they should have separate `<li>` markup.

In the source, they are a JSON array. So the idea is that an array is an ordered list, and **each item in the array is an item in the list**.

ihongda:
> Hongda, you may be aware of this, but just in case!
> 
> I'm using this latest branch, and I can see the new dictionary in the lookup, so that's great. There are some issues with CSS, but I'll take care of them when you are done.
> 
> Meanwhile, in terms of content, the markup is sometimes wrong when dealing with complex entries. See here, for example, the entry for _ekaṁ_.
> 
> Screenshot from 2021-01-22 18-02-29
> 
> The words in the definitions collide together. This is because they should have separate `<li>` markup.
> 
> In the source, they are a JSON array. So the idea is that an array is an ordered list, and **each item in the array is an item in the list**.

Okay, Bhante, I'll fix it :)

ihongda:
>The words in the definitions collide together. 

Bhante, this problem has been fixed!

sujato:
That's great, I can see the fix locally.

ihongda:
> That's great, I can see the fix locally.
> 
> The new dictionary must be used in both the lookup and the search results, but for me I only see it in the lookup. The search results no longer have NCPED.

This is a change for lookup and I will continue to modify the search results :)

sujato (commented):
It looks fine as far as I can see on the local site

alexander-vo (approved):
:+1:

firepick1 (approved):

Issue #1790, opened 2020-11-27
Purge all traces of Pootle from SC system


sujato:
Before using Bilara, we relied on the Pootle translation engine. That is now deprecated and all references to Pootle should be removed.

PR #1927, opened 2021-01-25
#1790 | Purge all traces of pootle and root from sc system

ihongda:


server/server/api/views/views.py
```diff
@@ -535,7 +534,7 @@ def get(self, uid, author_uid=''):
 
         result = results.next()
         self.convert_paths_to_content(result)
-        for k in ('root_text', 'translation'):
+        for k in ('translation',):
```
robertmusil, line 537: there's no point of iterating over a single-entry tuple

server/server/api/tests/test_views.py
```diff
@@ -18,9 +18,9 @@ def test_languages(client):
 
 
 def test_menu(client):
-    roots = utils.generate_roots(amount=10)
+    roots = utils.generate_navigation_docs(amount=10)
```
robertmusil, line 21: you can adjust the `roots` name too then

server/server/common/queries.py
```diff
@@ -149,7 +129,7 @@
 LET navigation_doc = DOCUMENT('super_nav_details', @submenu_id)
 
 LET descendants = (
-    FOR descendant IN OUTBOUND navigation_doc super_nav_details_edges
+    FOR descendant IN OUTBOUND navigation_doc super_nav_details_edges
```
robertmusil, line 132: random formatting oopsie?

PR #1928, opened 2021-01-26
Suttaplex and sutta API:Use the bilara instead of the PO.

ihongda:


sujato (commented):
running locally, it looks good

PR #1932, opened 2021-01-27
Set count of translations in YBR

alexander-vo:


sujato (commented):
Running this branch locally the count of suttas looks to be correct.

PR #1941, opened 2021-01-29
Use bilara text query instead of PO text query.

ihongda:


server/server/data_loader/generate_sitemap.py
```diff
@@ -53,6 +53,24 @@ def create_url(text):
...
+        db.aql.execute(
+            '''
+        FOR doc IN sc_bilara_texts
+            FILTER 'root' IN doc.muids OR 'translation' IN doc.muids
+            RETURN {
+                _key: doc._key,
+                uid: doc.uid,
+                lang: doc.lang,
+                author_uid: doc.muids[2],
+                name: null,
```
alexander-vo, line 67: `null` is this ok?

alexander-vo (approved):

server/server/data_loader/generate_sitemap.py
ihongda: I tried to read name from the names collection at the same time, but because sc_bilara_texts had more records, the query time actually exceeded expectations, and I saw that the 'name' field was not used in this method, so it was temporarily set to Null, and so on needed to be populated later.

Issue #1925, opened 2021-01-24
delete old collections from DB


sujato:
We have transitioned away from the PO data, we need to make sure unused collections are removed from DB.

Running the latest version, following the PO removal, and deleting and rebuilding DB, I see the following collections that should be removed.

most of them are already empty of data.

- [ ] available_languages: empty
- [ ] divisions: empty
- [ ] function_hashes: ? Not sure about this,  don't know what it is **leave for now**
- [ ] grouping: empty
- [ ] pitaka: empty
- [ ] po_markup: empty
- [ ] po_strings: empty
- [ ] root: empty
- [ ] root_edges: empty
- [ ] root_names: empty
- [ ] sect: empty
- [ ] segmented_data: replaced by sc_bilara_data
- [ ] text_divisions: empty

alexander-vo:
`function_hashes` is used by the ChangeTracker tool. This tool is designed as a "watchdog" for data changes (functions, modules, files, etc.) in the application. It's used mostly by load-data and if it sees that some data (files or functions) have not been updated, load-data will not update that data in the database, because it implies that it's up to date.
I don't know if it's needed and I'm not sure I know all the places where it's used, but it's there
I think @blake-sc can give you more details about it.

alexander-vo:
@sujato and what should I do with `available_languages`? The collection is empty and not used anywhere

sujato:
Sure, delete it, I have added it to the list above.

Issue #1755, opened 2020-11-10
Code cleanup


sujato:
With all the changes, we must make sure that anything unused is removed. The repo should only contain things that are actually used in the site.

- old code
- old data
- files
- dependencies

Also remove old and unused collections from DB

PR #1943, opened 2021-01-31
1925 delete old collections from db

alexander-vo:
- serverside code clean up #1755 
- old collection removed #1925


sujato (commented):
I've downloaded locally, run `make delete-database`, `make migrate`, `make load-data`. Everything seems to work fine. I've also checked the DB, it looks good , all the collections we need are there, and the ones we don't need are not there. I'm not seeing the `guides` and the `pali_reference` that we've been working on, I assume that's expected.

Issue #1942, opened 2021-01-30
Load guides and pali_reference_edition into the database


alexander-vo:
### Task
We need to load files `sc-data/structure/guides.json` and `sc-data/misc/pali_reference_edition.json` into the database.

### Details
The `guides.json` must be loaded into the `guides` collection.
The `pali_reference_edition.json` must be loaded into the `pali_reference_edition` collection.

### DB structure

- Document structure for `guides`:

_...IN PROGRESS..._

- Document structure for `pali_reference_edition`:

All documents must be loaded exactly the same as they appear in the `pali_reference_edition.json` file.

sujato:
Don't separate out the grouped editions. They are grouped for a reason: they will all appear in the UI as one set, and will be toggled on/off with one switch. Give me a minute, I'll see if I can make this more explicit in the data.

sujato:
I've pushed an updated version, hopefully it makes the intent clearer. We now have, for example:

```
  {
    "edition_set": "sya",
    "includes": [
      "sya1ed",
      "sya2ed",
      "sya-all"
    ],
    "name": "Other Thai editions"
  }
```

Each `edition_set` must be preserved as a single entity. 

Whether you split each entry into a separate document is up to you, it's an implementation detail. But conceptually there is no need: all this data will invariably be queried and served at one time. It is one set of data to create one UI interface.

alexander-vo:
@sujato ok, all documents in `pali_reference_edition` will be loaded as-is

alexander-vo:
@sujato now we have a structure similar to this (`guides.json`):
```
{
  "lang": {
    "en": {
      "sutta": "discourses-guide-sujato",
      "vinaya": "vinaya-guide-brahmali",
      "abhidhamma": "abhidhamma-guide-sujato",
      "dn": "dn-guide-sujato",
      "mn": "mn-guide-sujato",
      "sn": "sn-guide-sujato",
      "an": "an-guide-sujato"
    }
  }
}
```

and from what I understand so far, I can suggest the following structure in the database for each document:
```
{
    "type": "<root_key>",
    "lang": "<parent_key>",
    "uid": "<key>",
    "guide": "<value>"
}
```

and then from the existing document we will get the following documents in the database:
```
[{
    'guide': 'discourses-guide-sujato',
    'lang': 'en',
    'type': 'lang',
    'uid': 'sutta'
  },
  {
    'guide': 'vinaya-guide-brahmali',
    'lang': 'en',
[...]

PR #1944, opened 2021-01-31
#1942 | Added guides and pali_reference_edition

alexander-vo:
- migration with collections added
- guides.json file parsing
- pali_reference_edition.json file parsing
#1942


sujato (commented):
running locally, refreshed all the data, I can now see the guides and pali reference edition in DB.

All good!

PR #1946, opened 2021-02-01
Modify the dictionary lookup page based on the new dictionary data source.

ihongda:


sujato (commented):
Running this branch locally, after reloading DB, it isn't working properly. 

1. If i search for "atta" I get a results from the PTS dictionary at `search?query=atta`. **The first result must always be from NCPED, not PTS**. Only give a PTS result in the first place if the term, is missing from NCPED.
2. If I then click through to `define/atta` I get the NCPED dict and PTS dict, so that's good. But I also get an Indonesian dictionary! Which is bad.

Let's try another search.

1. `search?query=savatthi` gives me the DPPN. This is bad, it should have the entry from NCPED. I also see the google map, so this is good.
2. `define/sāvatthī` ives me NCPED and DPPN, which is good. But also a Spanish dictionary. Weird!

One more.

1. `search?query=karuna` Also gives me PTS. as first result, it should be NCPED.
2. here again we get three dictionaries, one of them in Indonesian!

The simple dictionaries are kept in sc-data/structure in folders named after their languages. the file names, also define the languages, eg. `pli2en_ncped`. obviously only serve the dictionaries in the appropriate language! These are recorded in the `to` field in the `dictionaries_simple_ collection in the DB.

ihongda:
> Running this branch locally, after reloading DB, it isn't working properly.
> 
> 1. If i search for "atta" I get a results from the PTS dictionary at `search?query=atta`. **The first result must always be from NCPED, not PTS**. Only give a PTS result in the first place if the term, is missing from NCPED.
> 2. If I then click through to `define/atta` I get the NCPED dict and PTS dict, so that's good. But I also get an Indonesian dictionary! Which is bad.
> 
> Let's try another search.
> 
> 1. `search?query=savatthi` gives me the DPPN. This is bad, it should have the entry from NCPED. I also see the google map, so this is good.
> 2. `define/sāvatthī` ives me NCPED and DPPN, which is good. But also a Spanish dictionary. Weird!
> 
> One more.
> 
> 1. `search?query=karuna` Also gives me PTS. as first result, it should be NCPED.
> 2. here again we get three dictionaries, one of them in Indonesian!
> 
> The simple dictionaries are kept in sc-data/structure in folders named after their languages. the file names, also define the languages, eg. `pli2en_ncped`. obviously only serve the dictionaries in the appropriate language! These are recorded in the `to` field in the `dictionaries_simple_ collection in the DB.

Okay, Bhante, thank you for your feedback, I'll revise it again.

ihongda:
>search?query=savatthi gives me the DPPN. This is bad, it should have the entry from NCPED. I also see >the google map, so this is good.
>define/sāvatthī ives me NCPED and DPPN, which is good. But also a Spanish dictionary. Weird!

`savatthi` can't found in dictionaries_simple, but `sāvatthī` can found (dictname='ncped').
also
`karuna` can't found in dictionaries_simple, but `Karuṇā` can found (dictname='ncped').

Bhante, Is this normal?

sujato:
I'm still getting results from non-English dictionaries, eg. Indonesian.

Also, the results on the search page are inconsistent. Eg `/search?query=attha` gives me a result from NCPED, but `search?query=dhamma` is from PTS.

ihongda:
> but search?query=dhamma is from PTS.

Because `dictionaries_simple` have not a definition of dhamma in English.:smile:

sujato:
Umm …

```
  {
        "entry": "dhamma",
        "grammar": "masculine (& neuter)",
        "definition": [
          "how the world of experience works, 
```

PR #1948, opened 2021-02-01
Fix the PWA.MENU AQL execution error.

ihongda:


server/server/common/queries.py
```diff
@@ -759,7 +759,7 @@
 
 class PWA:
     MENU = '''
-LET langs = UNION(@languages OR [], @include_root ? (FOR lang IN language FILTER lang.is_root RETURN lang.uid) : [])
+LET langs = UNION(@languages ? @languages : [], @include_root ? (FOR lang IN language FILTER lang.is_root RETURN lang.uid) : [])
```
alexander-vo, line 762: in general, we could leave `@languages OR []` since it is the same as `@languages ? @languages : []`
but if we want to use the second one, we need to split this string into 2 lines, because it takes 128 characters in one line, which exceeds python standards (120), and since we've just finished refactoring tasks - it would be good to follow the rules further
in general, I think it's easier to leave `@languages OR []`, but if you don't want to - you have to split the string)

alexander-vo (commented):
this request is slow, but that's not the main thing right now
All in all, it's good! but pay attention to the comment

Issue #1947, opened 2021-02-01
Optimize DB with indexing


sujato:
Since introducing the new data system, the DB calls seem slow, especially for the texts and suttaplexes.

Let's make them fast with Indexes!

https://www.arangodb.com/docs/stable/indexing-index-basics.html

PR #1952, opened 2021-02-02
#1947 | Add indexes for names and sc_bilara_texts

alexander-vo:
- add names index for uid, lang and uid+lang fields
- add sc_bilara_texts index for uid and uid+muids[*]
#1947


sujato (commented):
Oh yeah, awesome. Do it.

robertmusil (approved):

PR #1954, opened 2021-02-03
#1755 | Dependencies update

alexander-vo:
- removed unused dependencies
- dependencies versions updated
#1755


server/server/common/arangodb.py
```diff
@@ -76,13 +76,13 @@ def __init__(self, app=None):
...
         config = current_app.config['ARANGO_CLIENT']
-        return ArangoClient(host=config['host'], port=config['port'])
+        return ArangoClient(hosts=f'http://{config["host"]}:{config["port"]}')
 
     @property
     def client(self) -> ArangoClient:
         """
         Puts client object in g object.
-        :return: ArangoClient object connected to the databse.
+        :return: ArangoClient object connected to the database.
```
sujato, line 85: The changed line here throws an error with `make delete-database`. There appears to be a typo in it, but even when that's corrected there's still an error.

sujato (commented):
There's an error in arongodb.py.

server/server/common/arangodb.py
alexander-vo: What kind of error? I just ran it and it's okay...

sujato:
After running `make run-preview-env` it works fine.

Issue #1845, opened 2020-12-17
`sc-random-quote` element


sujato:
## Old version

The current SuttaCentral uses an element on the Home page to serve a rotating series of quotes.

The element is `sc-text-carousel`, and it uses `text-carousel.js`. The data sources may be found in sc-data/addition-info. I am not sure of the API endpoints, but they are found in ArangoDB under `epigraphs` and `why_we_read`.

It is used twice on the Home page with different content:

- A set of "epigraphs", i.e. sutta quotes: "The Buddha's Words".
    - source data: `epigraphs.json`
    - these each have a different link associated to the relevant sutta.
- A set of user quotes: "Why we read"
    - source data: `whyweread.json`
   - These all have the same static link to our forum, where the quotes were sourced.

## New version

For the new version, simply serve a new quote every time the page is refreshed. Don't change the content  once the page has loaded.

The staging site current serves a single quote as a placeholder. The HTML is:

```
<section class="plain">
          <blockquote>
               <span>The Buddha said to the monks who were quarrelling: “If animals can be courteous to each other, so can you.”</span>
          </blockquote>
          <a class="link-button quote-button" href="/">Read this sutta</a>
</section>
```

You'll probably want to aim for something like:

```
<section class="plain">
    <sc-random-quote class='epigraphs'>
          <blockquote>
              <span>The Buddha said to the monks who were quarrelling: “If animals can be courteous to each other, so can you.”</span>
          </blockquote>
          <a class="link-button quote-button" href="/">Read this sutta</a>
    </sc-random-quote>
</section>
```

Don't worry about styling, I'll do it. I think you won't need to use ShadowDOM for this component, but if you do, add a `css` literal in the component for styling.

sujato:
This works, except for the link on the sutta quote button. 

It should be something like:

```
<a class="link-button quote-button" href="/sn1.1/en/sujato">Read this sutta</a>
```

The UID is given in the `epigraphs` collection.

Note that this does NOT apply to the "Why We Read": this just has a static link, and is fine already.

PR #1961, opened 2021-02-05
#1845 | fix url

Prograd:


sujato (commented):
Running it locally, it looks good.

PR #1967, opened 2021-02-08
Improve the speed of conversion of Pali text.

ihongda:


sujato (commented):
I have run this locally and it works well, the speed problem is solved.

Issue #1822, opened 2020-12-09
Use Unicode for the API, not HTML entities


sujato:
The JSON in the API uses HTML entities instead of Unicode. Let us use Unicode all the way through. I don't think there is any reason to convert back and forth among encodings.

Unicode will be:

- Easier to read
- Smaller data transfer

PR #1970, opened 2021-02-09
#1822 | Use Unicode for the API, not HTML entities

alexander-vo:
- Add API response interceptor which dumps response data with unicode characters
#1822


sujato (commented):
Excellent, I can see this works, the API in this branch returns results natively in Unicode.

Issue #1966, opened 2021-02-05
Polymer remove: Update package files etc. 


sujato:
Once all the polymer elements are removed from front end, also ensure that dependencies are updated in files such as the following. This is not a complete list, I don't understand how these things work!

- [x]  `package.json`
- [ ] `polymer.json`
- [x]  in `index.html` delete `<script>window.polymerSkipLoadingFontRoboto = true;</script>`
- [x] in `sc-more-menu` see `iron-select`

PR #1973, opened 2021-02-10
Removing polymer remnants #1966

robertmusil:


sujato (commented):
I'm running it locally, everything looks good to me.

PR #1977, opened 2021-02-10
Re-adjust the navigation module according to the new routing rules.

ihongda:


server/server/api/views/views.py
```diff
@@ -1010,3 +1012,24 @@ def load_json(path):
...
+        for level in range(1, 10):
+            try:
+                results = db.aql.execute(SUTTA_PARENT, bind_vars={'uid': uid, 'level': level}).next()
+                if results:
+                    full_path.append(results)
+            except StopIteration:
+                break
+
+        full_path.append('pitaka')
+        return '/' + '/'.join(list(reversed(full_path)))
```
alexander-vo, line 1033: I think the logic of going up through the parents and getting the path we want can be implemented completely in AQL.
I think it would be more correct to try to implement it

server/server/common/queries.py
```diff
@@ -1001,3 +1001,8 @@ class PWA:
     LIMIT 1
     RETURN {@uid: doc.filepath}
 '''
+
+SUTTA_PARENT = '''
+FOR parent_doc IN @level INBOUND DOCUMENT('super_nav_details', @uid) super_nav_details_edges
+    RETURN parent_doc.uid
+'''
```
alexander-vo, line 1008: this query may contain the logic you described through python loops

alexander-vo, line 1008: forgot the blank line at the end of the file =)

alexander-vo (changes requested):
@ihongda in general I described everything in the comments, I can try to do everything myself, but if you want to do it yourself - no problem, you might be interested)

server/server/api/views/views.py
alexander-vo, line 1033: and of course, we have to return JSON documents, not plain strings, so here I would recommend that you do this
```
return { 'full_path': <result_path_variable> }
```

server/server/common/queries.py
ihongda: Yes, it can be executed up to ten times :smile:

alexander-vo (approved):

PR #1978, opened 2021-02-10
Sujato next gen

sujato:
Makes a few changes to the lastest build:

- remove webcomponents-loader
- restore suttaplex to search results
- remove mwc-textfield from search, use native `<input>`
- small style tweaks


ihongda (approved):

ihongda:
深度截图_选择区域_20210211102611

Bhante, is the font smaller?

sujato:
Hmm, doesn't look like that to me. Hang on, I'll give it an explicit font-size.

Issue #1103, opened 2018-08-09
Swath preprocessor for Thai


aminah-sc:
@sujato commented on [Mon Jul 17 2017](https://github.com/suttacentral/legacy-suttacentral/issues/207)

As per this Discourse thread, we should use Swath to preprocess all Thai text. 

https://discourse.suttacentral.net/t/smart-word-analysis-for-thai/5833/3

sujato:
Closing, because when implemented it did not offer a meaningful improvement and in some cases of Pali names made the breaks worse.

PR #1980, opened 2021-02-11
#1103 | Swath preprocessor for thai

alexander-vo:
- Add swath processing for thai text files
#1103


sujato:
After testing, it became apparent that the preprocessor did not substantially improve word breaks, and in some cases made them worse.

PR #1983, opened 2021-02-12
Feature/add references from database

### LINKED ISSUE #1754: Implement reference settings in top sheet
The top sheet contains a set of check buttons for displaying the reference details of texts.

- None: Display no references. This is default.
- Main: Display segment number references. This currently works.
- Other Pali editions: a set of about twenty editions of the Pali canon for which we have reference data. Each of these can be individually toggled. **Currently this is a mockup only**

Currently the Pali editions are hard coded. Remove this and use the data from `pali_reference_edition.json`

https://github.com/suttacentral/sc-data/tree/master/misc

This will have to be loaded to DB first #1942 
:warning: the order is significant, please preserve it.

- Display a checkbox for each `uid` or set of `uid`s. 
- Note that in some cases, multiple related editions are toggled with the same checkbox.
- Show the name as the label of the checkbox.
- When checked, toggle all reference IDs in text for that `uid` or set of `uid`s.
- All settings should persist on the site until the user changes it.

Prograd:
close #1754


server/server/api/views/views.py
```diff
@@ -1010,3 +1013,17 @@ def load_json(path):
...
+
+class PaliReferenceEdition(Resource):
+
+    @cache.cached(key_prefix=make_cache_key, timeout=default_cache_timeout)
+    def get(self):
+        db = get_db()
+        results = db.aql.execute(SUTTA_PALI_REFERENCE)
+        pali_references = []
+        for reference in results:
+            pali_references.append(reference)
```
alexander-vo, line 1026: can be replaced by
```
pali_references = list(db.aql.execute(SUTTA_PALI_REFERENCE))
```

alexander-vo (approved):

sujato (commented):
The editions are present, and the basic on/off functionality works, but there are three main bugs remaining:

- [x]  The editions are doubled, i.e. each edition is present twice.
- [x] These are checkboxes, which means you can select multiple items. I should be able to show as many editions as I want. (Radio buttons are used to toggle a single selection, see https://material.io/components/selection-controls#usage)
- [x] The toast announces each switch, but it is not necessary. We can remove the toast.

Screenshot from 2021-02-15 20-19-37


Now, currently with the exception of "Main" and "Pali text society", the switches don't work, i.e. nothing is displayed. We have to make sure the content is actually present in the text. I'll make a separate issue for this. This will be a job for either you or Hongda next week.

https://github.com/suttacentral/suttacentral/issues/1984

sujato:
doubled editions problem is fixed

server/server/api/views/views.py
Prograd: Fixed in 1497974

server/server/common/queries.py
```diff
@@ -1001,3 +1001,12 @@ class PWA:
     LIMIT 1
     RETURN {@uid: doc.filepath}
 '''
+
+SUTTA_PALI_REFERENCE = '''
+FOR pali in pali_reference_edition
+    Collect edition_set = pali.edition_set, name = pali.name, short_name = pali.short_name
```
alexander-vo, line 1007: and one more thing, my mistake, I missed it last time, sorry =(
but I think we should write keywords in AQL in caps, because we do that everywhere (`in` => `IN`, `Collect` => `COLLECT`, etc.)

Prograd: Fixed in e7658b8

ihongda (approved):

sujato:
I can now check multiple editions, so that is correct.

However  "None" still does not work correctly.

None means none. When "none" is selected, **every other checkbox is unselected**. This is the default state. "None" clears every other checkbox, always.

client/elements/addons/sc-top-sheet-views.js
```diff
@@ -693,6 +686,12 @@ class SCTopSheetViews extends LitLocalized(LitElement) {
           showHighlighting: showHighlighting,
         });
       },
+      setDisplayedReferences(displayedReferences) {
```
agilgur5, line 689: `setDisplayedReferences` replaced `setReferenceDisplayType` (above on line 671), but didn't delete that function. I've removed it in #3714

PR #1997, opened 2021-02-17
#1690 | Implement pali hyphenator

alexander-vo:
- add hyphenator for Pali and Sanskrit texts
#1690


sujato:
I've added the full list of words from the glossary, and corrected the one case of ṃ to ṁ. (The file should check for both, but because of a find and replace error we had `ṃṃ` instead of `ṃṁ`.)

Now it seems to work just great, all the cases I have checked look good. I was worried about the processing time with the large glossary, but it ends up as only 12s or so, so no problem.

I also added some style changes by mistake, oops! They fix the issue raised by Sabbamitta yesterday. I mixed up my branches. But anyway they should be safe to merge.

PR #2000, opened 2021-02-18
Fixed #1956 - Dhammapada texts not showing

### LINKED ISSUE #1956: Dhammapada text collections are not showing
Legacy html texts are not working on latest master. Eg. Dhammapada is buggy and messy:

https://staging.suttacentral.net/dhp1-20/en/sujato

This still persists in the latest local version.

It applies to all the Dhammapada texts.

http://localhost/pitaka/sutta/minor/dharmapadas

And maybe others too.

Is it the same issue as this?

https://github.com/suttacentral/suttacentral/issues/1950

robertmusil:


client/elements/text/sc-bilara-segmented-text.js
```diff
@@ -607,57 +581,33 @@ class SCBilaraSegmentedText extends SCLitTextPage {
...
+      this._addRootSuttaMarkupToSpan(CSS.escape(key));
     });
   }
 
-  _addRootSuttaMarkupToSpan(suttaRootElement, key) {
-    if (suttaRootElement) {
-      const segmentElement = suttaRootElement.querySelector(`#${key}`);
-      if (segmentElement) {
-        segmentElement.appendChild(this._addRootSuttaSpan());
-      } else {
```
robertmusil, line 629: Removing this else is slightly risky, but it allows for all of this amazing optimization. I tested it very thoroughly and can't see it ever being used.

So a quick summary of what it did:
- If there's a fragment with and id in a root text that doesn't exist in bilara-data html markup it then adds a new segment with this fragment without corresponding translation elements.

I believe this case would never happen and this was here for legacy reasons and I tested through many of the texts and I didn't find any cases in which this would happen. Still, if a case is found where a part of root text seems missing, this is what's responsible, but without proper test-case I cannot fix it.

sujato: I'm not exactly sure what the situation is here, but let me just explain the general idea. 

As a general rule, the complete set of segments is in bilara-data/html. The case where a segment is found in either /root or /translation but *not* in /html should never occur. This should be tested by the text integrity scripts that are being implemented by dayunbao and Alex.

However, it frequently happens that texts in /translation will omit segments found in /root. 

It also happens, although less often, that there are segments in /translation that are not in /root. This happens where, for example, a passage is abbreviated in /root but the translator wishes to have it in full; or if the translator wishes to add headings not found in /root. This does not happen in my sutta translations at all, I avoided doing this. But it is in Brahmali's Vinaya translations, which are not yet published.

sujato (commented):
The Dhammapada texts appear to work just fine.

client/elements/text/sc-bilara-segmented-text.js
robertmusil: So if there shouldn't be any segments in /root that are not in /html then this is safe to merge.

PR #2004, opened 2021-02-19
Fixed: In some cases, the breadcrumbs are not updated

ihongda:
In some cases, the breadcrumbs are not updated
e.g.
http://localhost/an7.1
http://localhost/dn7
http://localhost/sn55.1
...


sujato (commented):
Testing locally, it seems to work well.

Issue #1743, opened 2020-10-29
Add publication info to segmented texts


sujato:
Legacy texts have the licensing and other info hard-coded in the file, eg.

```
<footer>
<p>Translated by Venerable <span class='author'>Dhammarakkhita</span>.</p>
<p>Creative Commons Attribution-NonCommercial-ShareAlike 4.0 International.</p>
<p>Prepared for SuttaCentral by Bhikkhu <span class='editor'>Sujato</span>.</p>
</footer>
```

This is then displayed via the "info" button as a dialog (this may change, but anyway there is some means of displaying it).

We need a way of displaying similar info for the segmented texts. 

## Data source

The relevant information is in `bilara-data/_publication.json`.

Here I give an HTML structure.
- json keys as class names.
- `{{data}}` is indicated like this
- In the case of `edition` and `author` there may be more than one, so check this.
- If a field is empty in `_publication.json` then omit it.
- Include rdfa metadata

Just put the following on the page, insert the `{{data}}` from `_publication.json`, and we're good to go!

```
<footer>
<h2>About this text</h2>
<section class='text_metadata' xmlns:dc='http://purl.org/dc/elements/1.1/' about='{{source_url}}'>
<dl class='main_details'>
<dt class='translation_title'>Translation title</dt>
<dd class='translation_title' property='dc:title'>{{translation_title}}</dd>
<dt class='translation_subtitle'>Translation subtitle</dt>
<dd class='translation_subtitle' property='dc:title'>{{translation_subtitle}}</dd>
<dt class='translation_language'>Translation language</dt>
<dd class='translation_language' property='dc:language'>{{translation_language}}</dd>
<dt class='root_title'>Root title</dt>
<dd class='root_title' property='dc:title'>{{root_title}}</dd>
<dt class='root_language'>Root language</dt>
<dd class='root_language'>{{root_language}}</dd>
<dt class='author_name'>Translator</dt>
<dd class='author_name' property='dc:creator'>{{author_name}}</dd>
</dl>
<dl class='descriptive_details'>
<dt class='translation_description'>Translation description</dt>
<dd class='translation_description' property='dc:description'>{{translation_description}}</dd>
<dt class='translation_process'>Translation process</dt>
<dd class='translation_process' property='dc:description'>{{translation_process}}</dd>
</dl>
<dl class='metadata_details'>
<dt class='text_uid'>Text identifier (UID)</dt>
[...]

PR #2005, opened 2021-02-19
#1743 | The draft solution

Prograd:


sujato (commented):
Note that this component is the Bilara equivalent of what in the legacy HTML texts is `sc-top-sheet-info.js`. Every text will have either one or the other, never both.

So we should make this clear and obvious right from the start. 

Two ways of doing this:

- **either** just have one component, i.e. combine the new element with `sc-top-sheet-info.js`.
- **or** adopt a consistent and meaningful naming pattern for the two components. 

Taking the second option, we could use the names:

- `sc-top-sheet-publication-legacy.js`
- `sc-top-sheet-publication-bilara.js`

Prograd:
There is still one outstanding problem, in the respons we don't have the `Number of volumes`. Unfortunately I don't know how I should get it.

ihongda (approved):

PR #2007, opened 2021-02-21
Fixed #1995: toolbar navigation transitions are not great.

### LINKED ISSUE #1995: toolbar navigation transitions are not great
When I proposed the new navigation system, I had in mind something that would be smooth, elegant, and informative. A good navigation should reassure the user that they are on their way to their destination. We are nearly at the point of ironing out the final bugs in the *content* of the navigation, but the feel is still not as good as it should be.

The basic problem is that the toolbar is janking and rebuilding itself constantly. It shouldn't do that at all: it should be a SPA type experience. The toolbar should always remain there, only the content should change (the exception being the Home page.)

I had hoped this would be fixed by the various changes and updates to the navigation, but that hasn't happened, unfortunately.

This problem is not so obvious when developing, as the transitions are fairly good in some cases. Specifically, Chrome seems to do them much better than Firefox. And cache seems to make a big difference. So developers aren't likely to see such a big problem. Try it in FF with the cache cleared and you'll see something like this.

https://www.youtube.com/watch?v=hC3J4zjEEQM

There are some specific problems:

- previous pages appear when clicking the next page
- the toolbar is vanishing and rebuilding on each page transition
- wrong data is flashing into view and then being replaced by right data

But the more general issue is that it just isn't working smoothly.

So, how do we achieve this? 

In theory it is simple: keep the toolbar constant, swap in new data, and swap in new page.

What is the simplest, most elegant way to achieve this in practice?

The Turbo library aims to do just this.

https://github.com/hotwired/turbo

However, it is designed to be used with HTML not JSON. But it does illustrate the idea I am getting at.

ihongda:
Rewrite the relevant code according to the method provided by Robert, which works well, 
Thx @robertmusil :smile:


sujato:
I'm getting a bug where, when I am on a sutta and I return to the Home page, the title of the sutta is still there:

Screenshot from 2021-02-21 15-49-21

sujato (commented):
The page transitions are still janky and clumsy especially in Firefox. Chrome is better.

here's a screencast:

https://www.youtube.com/watch?v=vb6O3lYdHAM

Compare the previous version screencast:

https://youtu.be/hC3J4zjEEQM

sujato:
There's another bug where in the breadcrumbs, "Discourses" changes to "Sutta".

Screenshot from 2021-02-21 16-24-12
Screenshot from 2021-02-21 16-24-16
Screenshot from 2021-02-21 16-24-21

ihongda:
Bhante, I have fixed a compatibility issue related to firefox and believe the problem has been resolved.

ihongda:
> The page transitions are still janky and clumsy especially in Firefox. Chrome is better.
> 
> here's a screencast:
> 
> [youtube.com/watch?v=vb6O3lYdHAM](https://www.youtube.com/watch?v=vb6O3lYdHAM)
> 
> Compare the previous version screencast:
> 
> [youtu.be/hC3J4zjEEQM](https://youtu.be/hC3J4zjEEQM)

Bhante, I have fixed a compatibility issue related to firefox and believe the problem has been resolved.

ihongda:
> There's another bug where in the breadcrumbs, "Discourses" changes to "Sutta".
> 
> Screenshot from 2021-02-21 16-24-12
> Screenshot from 2021-02-21 16-24-16
> Screenshot from 2021-02-21 16-24-21

This problem has also been resolved.

ihongda:
>I'm getting a bug where, when I am on a sutta and I return to the Home page, the title of the sutta is still there:

This problem also resolved :smile:

Issue #1915, opened 2021-01-20
Keep navigation data in sc-data/structure not JS components


sujato:
**LOW PRIORITY!!** 

Currently the navigation components include some data that duplicates the data in sc-data/structure, ore else should better be kept there. We should keep a clean separation between data and logic.

In `sc-navigation-common.js` we have a list of "Guides". These should be added as a separate JSON file in /structure. I have done this with this commit:

https://github.com/suttacentral/sc-data/commit/d02a47b3a42b21d3c96f83c9ff59ccfbb110f3dc

This list of guides does two things:

- It specifies guides per language. In future we will have translated guides.
- It adopts a consistent URL. The previous URL for discourses, etc. were underspecified and too ambiguous. Obviously this change would have to be reflected in the page names and so on.

In the same file, we also find a list of `shortcuts`. These live in `sc-data/structure/shortcuts.json` and should be imported from there.

***

Another example: in `sc-text-page-selector.js ` we have a list of language ISO codes. These should imported from `sc-data/structure/language.json`.

See comment here:

https://github.com/suttacentral/suttacentral/pull/2037#pullrequestreview-605880569

PR #2037, opened 2021-03-06
Fixed:failure to access Japanese sutta that can't be routed.

ihongda:
The iso code list of routing modules does not contain 'jpn', This causes the URL to not be parsed correctly.


sujato (commented):
This solves the issue for now, but would it not be better to import the language ISO codes from `sc-data/structure/language.json`? Otherwise we are just asking for trouble by maintaining two separate lists. See #1915 for similar cases. (I will add this case to that issue for the future.)

sujato (approved):

Issue #2046, opened 2021-03-08
Prev/next localization and breadcrumbs are still buggy


sujato:
Here is a Japanese text. I got to it by selecting japanese from the menu, then navigating through the nav cards.

As you can see, the breadcrumbs are missing items. And the prev/next is in German!

![Screenshot from 2021-03-08 18-16-21](https://user-images.githubusercontent.com/6112010/110287640-f9108500-803a-11eb-82bb-1c3a2bc3aae9.png)

If I change language to English and navigate there, the breadcrumbs are correct, but the prev/next is still in german!

![Screenshot from 2021-03-08 18-22-11](https://user-images.githubusercontent.com/6112010/110287871-4c82d300-803b-11eb-9782-eea76dc12a98.png)

ihongda:
Currently, the prev/next name is always obtained from the `names` collection, such as `an4.58`, with only `pli` and `German` versions, no English version, and the pali name is displayed by default if there is no version of the language.

PR #2054, opened 2021-03-10
Fixed Prev/next localization and breadcrumbs are still buggy #2046

ihongda:
Currently, the prev/next name is always obtained from the names collection, such as an4.58, with only pli and German versions, no English version, and the pali name is displayed by default if there is no version of the language.

### DIFF
--- a/client/elements/navigation/sc-navigation-common.js
+++ b/client/elements/navigation/sc-navigation-common.js
@@ -318,8 +318,13 @@ export async function RefreshNav(uid) {
   let currentURL = '/pitaka';
   const currentNav = store.getState().navigationArray;
 
+  const truePathLength = URLs.filter(x => x !== null && x !== '' && x !== 'pitaka').length;
+  const navSuttaPathLength = currentNav.filter(
+    x => x != null && x.groupId !== uid && x.type !== 'home'
+  ).length;
+
   const fatherLevelExists = currentNav.some(x => x !== null && x.groupId === URLs[URLs.length - 1]);
-  if (fatherLevelExists) {
+  if (fatherLevelExists && navSuttaPathLength === truePathLength) {
     return;
   }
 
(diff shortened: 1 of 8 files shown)


sujato (commented):
It's still buggy. Running this branch locally I get "Data load error" on certain texts, eg. dn1/en/sujato or mn3/de/sabbamitta.

If I set to English on dn2/en/sujato, then change to German, nothing happens. If I set it to german on the Home page, then navigate to a text, then reset to English, again, nothing happens. Even if I do a hard refresh, the language in the breadcrumbs and the prev/next is still wrong.

Same thing on the suttaplex list. If I am on dn-silakkhandhavagga and the breadcrumbs are in german, I change to English, the suttas update correctly in the suttaplex, but in the breadcrumbs, the blurbs, and the sutta titles, nothing happens. Now I do a hard refresh, they update to German. Change back to english, only the suttas update, not the breadcrumbs or titles or blurbs. Hard refresh, then the titles and blurbs update, but not the breadcrumbs.

sujato (commented):
Mostly good, but there are still some remaining issues: the language does not always change. I will detail these in #2054

sujato:
Test conditions: latest Hongda-work2 branch, Chrome incognito, with all site cache cleared.

We still have inconsistent and unpredictable results when changing languages. Bugs are **bold**.

- Home page: English --> German, works well.
- /sutta: English --> German, works well, except **title is missing from context toolbar**. It appears on refresh.
- /sutta/long: English --> German: only context toolbar changes. **Nav cards titles and blurbs do not change**. Again they appear on refresh.
- /sutta/long/dn: English --> German:  everything is correct!
- /dn-silakkhandhavagga: English --> German:  Only default translations change. **Title missing from context toolbar. Suttaplexes still in English.** On refresh, context toolbar and suttaplexes are correct. However even after refresh, **breadcrumb does not update to German**.
- /dn1/de/franke: English --> German: Everything updates correctly, including the breadcrumbs. However the **suttaplex dropdown does not update titles or blurbs.**

ihongda:
>Home page: English --> German, works well, except tipitaka blurbs do not change to German. These are translated (in super-blurbs.json) yet they never appear on the site.

Fixed!

Issue #2051, opened 2021-03-09
Use checkbox instead of switch in more-menu


sujato:
The Mwc components have a builtin method for using checkbox in mwc-list. There's no such thing for switch. Let's use the checkbox, it'll be an easier implementation, and we won't have such a hassle styling it.

https://material-components.github.io/material-components-web-components/demos/list/

sujato:
There's a small bug affecting this.

The mwc-check-list-item has an odd UI where there are two clickable fields: the item as a whole, and the actual checkbox. (I think it should be just the checkbox, but this is what the mwc official demo does.) So on the demo, if you click anywhere on the item, it turns the checkbox on or off:

https://material-components.github.io/material-components-web-components/demos/list/

On our implementation, however, the *function* works when clicking anywhere on the item, but the check mark does not appear unless you click the checkbox itself. That means you can end up in a state where the dark theme is applied, yet it is not checked:

![Screenshot from 2021-03-13 10-46-20](https://user-images.githubusercontent.com/6112010/111010080-8fe99280-83e9-11eb-975a-39adf6499bdf.png)

If you refresh the page, the check appears.

PR #2060, opened 2021-03-12
#2051 Use checkbox instead of switch in more-menu 

ihongda:


sujato (commented):
It's good, except the "always show toolbar" option doesn't work.

(Also I directly pushed to this branch the theming colors for the checkbox, hope that's okay!)

ihongda:
>It's good, except the "always show toolbar" option doesn't work.
Let me check it !

>(Also I directly pushed to this branch the theming colors for the checkbox, hope that's okay!)
Of course, no problem. 😄

sujato (approved):
all good

PR #2066, opened 2021-03-13
Fix a bug of more-menu.

ihongda:
> There's a small bug affecting this.
> 
> The mwc-check-list-item has an odd UI where there are two clickable fields: the item as a whole, and the actual checkbox. (I think it should be just the checkbox, but this is what the mwc official demo does.) So on the demo, if you click anywhere on the item, it turns the checkbox on or off:
> 
> https://material-components.github.io/material-components-web-components/demos/list/
> 
> On our implementation, however, the *function* works when clicking anywhere on the item, but the check mark does not appear unless you click the checkbox itself. That means you can end up in a state where the dark theme is applied, yet it is not checked:
> 
> Screenshot from 2021-03-13 10-46-20
> 
> If you refresh the page, the check appears.
> 
> 

This PR will fix the above problems.

### DIFF
--- a/client/elements/menus/sc-more-menu.js
+++ b/client/elements/menus/sc-more-menu.js
@@ -167,17 +167,12 @@ class SCMoreMenu extends LitLocalized(LitElement) {
 
   _onThemeChanged(e) {
     const chk = e.currentTarget.shadowRoot.querySelector('mwc-checkbox');
-    chk.addEventListener('change', () => {
-      const newTheme = chk.checked ? 'dark' : 'light';
-      this.actions.changeAppTheme(newTheme);
-    });
+    this.actions.changeAppTheme(chk.checked ? 'dark' : 'light');
   }
 
   _onToolbarDisplayModeChanged(e) {
     const chk = e.currentTarget.shadowRoot.querySelector('mwc-checkbox');
-    chk.addEventListener('change', () => {
-      this.actions.changeAlwaysShowToolbarState(chk.checked);
-    });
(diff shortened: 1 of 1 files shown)


sujato:
Ahh, this is not quite right, I think I did not specify the issue clearly enough:

- **[just use the default behavior as here](https://material-components.github.io/material-components-web-components/demos/list/)**

Previously, the "click on checkbox" was separated from "click on item" so that you could activate dark theme, but the checkbox was not checked. This is no longer happening.

However! When you click on the item, it does not activate the checkbox at all. It still has a hover and ripple effect, so it seems like it should be active, but it doesn't do anything. This is bad UI, it sends the wrong signal to the user.

The default behavior of the component is that when you click anywhere on the item it toggles the checkbox *and* executes the function. **Just use the defaults**.

ihongda:
> Ahh, this is not quite right, I think I did not specify the issue clearly enough:
> 
> * **[just use the default behavior as here](https://material-components.github.io/material-components-web-components/demos/list/)**
> 
> Previously, the "click on checkbox" was separated from "click on item" so that you could activate dark theme, but the checkbox was not checked. This is no longer happening.
> 
> However! When you click on the item, it does not activate the checkbox at all. It still has a hover and ripple effect, so it seems like it should be active, but it doesn't do anything. This is bad UI, it sends the wrong signal to the user.
> 
> The default behavior of the component is that when you click anywhere on the item it toggles the checkbox _and_ executes the function. **Just use the defaults**.

Okay, Bhante, I'll adjust later :smile:

sujato (approved):
Looks good!

Issue #1774, opened 2020-11-19
Use "linear progress bar" instead of three dots.


sujato:
We currently use a three-dot animation for loading, with `sc-bouncing-loader`.

On reflection, I'd rather use a "linear progress bar". I find the bouncing loader appears somewhat randomly in the middle of the page. Since we have an app bar on every page, why not use the "linear loading" design?

https://material.io/components/progress-indicators#usage

There's a pure CSS version here, maybe it would work!

http://dev.gojko.net/web/2015/09/19/material-design-progress-pure-css.html

I've wrapped it up as a custom element:

[sc-linear-progress.zip](https://github.com/suttacentral/suttacentral/files/5566499/sc-linear-progress.zip)

PR #2083, opened 2021-03-17
Use "linear progress bar" instead of three dots. #1774

ihongda:


sujato (commented):
I can see the loader on when I do a search. One thing, the loader should be attached to the toolbar, not in the middle of the page. I was thinking it would appear at the bottom of the toolbar, underneath the context toolbar or ATB. It should not, however, *add* height to the toolbar, but rather appear in the bottom few pixels of the toolbar.


See example here:

https://kstatic.googleusercontent.com/files/b14128fc1118df8911bb9d5405a075c013d1f060e2fee2d6d78154f0d1143b41cbf1dc31d09de993bec3b18e078e8bcd4b226b245a394e70852d20d10a45f0e3

ihongda:
Yes, It's really better at the bottom of the toolbar :smile:

ihongda:
The linear-progress have attached to toolbar.

sujato (approved):
looks good!

Issue #2095, opened 2021-03-20
Improve cache-busting for new versions


sujato:
Our cache tends to be very sticky. Users commonly have a problem with getting updates to actually appear.

Can we use version numbers or something to ensure that users actually get the updated versions?

ihongda:
We can improve by modifying the webpack :)

PR #2101, opened 2021-03-22
Improve cache-busting for new versions #2095

ihongda:


sujato:
I'm not sure how to test this, but I'm happy to go ahead and merge

ihongda:
> I'm not sure how to test this, but I'm happy to go ahead and merge

Thx Bhante,

Modifying this configuration will result in a different file name per compilation than before, so that the browser will load the latest js files.

I'll test it at Staging :)

Issue #2094, opened 2021-03-19
Pali texts (and other root texts?) are not showing in search results


sujato:
Currently results for Pali search return only results in translations.

https://suttacentral.net/search?query=bilara

We should index Pali texts. Also, I think the other root texts  are also not being indexed.

PR #2102, opened 2021-03-25
Fixed: Pali texts (and other root texts?) are not showing in search results #2094

ihongda:
The code for this PR generates indexes of bilara text.

run `make index-elasticsearch` to rebuild the index.


sujato (approved):
looks good, I can see the search results locally.

PR #2123, opened 2021-04-15
Lazy load sc-toasts.js

ihongda:


sujato (approved):
This further reduces the DOM size to 410 nodes. Looks good!

PR #2124, opened 2021-04-15
Modify Chinese lookup based on feedback.

ihongda:
1. Do not include irrelevant characters
2. The first entry should be the longest cluster found in the dictionary.
3. The the second-longest, until we get to the single glyph.


sujato (approved):
looks good

PR #2131, opened 2021-04-20
Remove compact mode from sc-tiptaka.js.

ihongda:


sujato (approved):
works as expected, compact mode disabled on Home page tipitaka

PR #2136, opened 2021-04-27
Fix: Click previous/next,If there is no relevant translation, an error will occur.

ihongda:
Problem description:
[https://discourse.suttacentral.net/t/suttacentral-2021-bug-reports-here/19609/98?u=hongda](https://discourse.suttacentral.net/t/suttacentral-2021-bug-reports-here/19609/98?u=hongda)

e.g.
[http://localhost/sa87/en/analayo](http://localhost/sa87/en/analayo)

Click the `Next` Button, an error will occur.


sujato (approved):
looks good

PR #2137, opened 2021-04-28
Fix: Segmented text: Click previous/next,If there is no relevant translation, an error will occur. 

ihongda:
e.g.

[http://localhost/thig5.12/de/sabbamitta](http://localhost/thig5.12/de/sabbamitta)


sujato (approved):
ok, seems to work fine!

PR #2142, opened 2021-05-05
Add reference data to texts on user request.

ihongda:
1. No references are loaded by default.
2. References are loaded on user request via the "Reference" options in the top sheet.
3. Users may select as many sets of references as they wish.
4. User selections are persistent until the user changes it.
5. Load the `root_edition.json` to ArangoDB
6. Add API `API/root_edition`
7. Load the `available_voices` to ArangoDB
8. Add API `API/available_voices`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,7 +12,6 @@ client/data/
 /server/nextdata/
 .DS_Store
 /client/node_modules/
-package-lock.json
 client/build/
 /client/localization/pootle/prePootleFiles/
 /client/localization/pootle/generatedPoFiles/
@@ -20,4 +19,8 @@ client/build/
 generatedPoFiles/
 /pootle/log/*
 .prod.env
-environment_prod.yml
+.history
+client/.eslintrc.json
+client/.eslintrc.js
+client/.prettierrc
+.gitignore
--- a/.gitignore
+++ b/.gitignore
@@ -20,6 +20,7 @@ client/build/
 generatedPoFiles/
 /pootle/log/*
 .prod.env
+environment_prod.yml
 .history
 client/.eslintrc.json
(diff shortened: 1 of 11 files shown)


sujato (changes requested):
These items work:

- No references are loaded by default.
- References are loaded on user request via the "Reference" options in the top sheet.
- User selections are persistent until the user changes it.

But this item does not work:

- User may select as many sets of references as they wish.

Currently only the main and pts references work. The feature must work with *all* references that are available. For example for dn1 the references are here:

https://github.com/suttacentral/bilara-data/blob/published/reference/pli/ms/sutta/dn/dn1_reference.json

Thus all the references must be available at `dn1:1.1.1`:

```
bj7.2, cck9.1, csp1ed6.1, csp2ed6.1, dr9.1, km14.1, km14.2, lv1.1, lv1.2, mc9.1, ms6D_2, msdiv1, ndp6.3, pts-cs1.1, pts-vp-pli1.1, sc1, sya9.1, vri1.1, vri1.2
```

Moreover, there are the following bugs.

- the pts-cs reference is somehow changed to be just `cs`: `<a class="pts" title="Pali Text Society vol/page number." id="cs1.1" href="#cs1.1">cs1.1</a>`. Either the wrong source is used or it is being altered somehow. References must not be changed from the source.
- The "none" button does not work. It should clear all entries. None means none.

See original spec for this feature:

https://github.com/suttacentral/suttacentral/issues/1754

https://github.com/suttacentral/suttacentral/issues/1754

ihongda:
Bhante, I have fixed the related problem. Since new data is loaded into `ArangoDB`, additional execution is required:

1. `make migrate`
2. `make load-data`

Issue #2078, opened 2021-03-16
Voice speaker icon appears on suttaplexes for texts that Voice does not support


sujato:
Go here:

https://suttacentral.net/ma

click on the voice icon, get an error:

https://voice.suttacentral.net/scv/index.html#/sutta?search=ma1&lang=en

@firepick1 what's the best way for us to get a list of texts supported by Voice?

firepick1:
Bhante, we've had so much trouble with exactly this question that we've adopted an approach analogous to the SuttaCentral "published" branch.  Voice applications rely on [ebt-data](https://github.com/ebt-site/ebt-data), which is a carefully curated copy of bilara-data/published. The ebt-data repository is 1/4 the size of bilara-data. It is smaller for several reasons:

* Voice applications don't need the full Bilara git history (this is rather massive).
* Voice applications don't need the reference, html, variant trees.

The smaller repository is periodically updated from bilara-data by Github Actions. It's not a fork because we don't want the git history.

firepick1:
Actually, there is an easier way to determine what is supported:

https://github.com/sc-voice/scv-bilara/blob/main/src/auto/suidmap.json

We currently don't support the Vinaya just yet but will soon...

PR #2144, opened 2021-05-10
Fix: Voice speaker icon appears on suttaplexes for texts that Voice does not support #2078

ihongda:


sujato (approved):
looks good, Voice icons only appear where they should

PR #2155, opened 2021-05-14
Use Bilara author metadata for segmented texts.

ihongda:
The impact of this update is as follows:
1. Author info of the Suttaplex.
2. Author info of the Parallels.
3. Author info of the Sutta (segmented text).

The modification must ensure that the author information in the above list is consistent with the previous information.


sujato (approved):
I'm not sure how to test this, but it seems fine so far as i can see.

PR #2157, opened 2021-05-19
Use the Chinese system for component names.

ihongda:


sujato (approved):
The renaming conventions seem good. Note that I am away from home, so I can't actually run it to see if it works!

ihongda:
> The renaming conventions seem good. Note that I am away from home, so I can't actually run it to see if it works!

Thx Bhante, 
I will test it again before merging.

PR #2159, opened 2021-05-20
Keep navigation data in sc-data/structure.

ihongda:


sujato (approved):
navigation seems to work well

Issue #2115, opened 2021-04-07
Transliteration:  Many of the script labels are gibberish


virtualvinodh:
#### Description:

The menu items in the transliteration are simply incorrect

#### Steps to reproduce:

Any Pali/Sanskrit page with transliteration plugin

#### Expected behaviour:

The name list is just gibberish in many of the scripts. The name of the scripts is spelt using English convention and cannot be transliterated back to their native alphabets.  I would recommend retaining their English label, in case you don't how they are spelt in their own script. 

#### Screenshots:

![image](https://user-images.githubusercontent.com/1050975/113927579-f5b11a80-97ed-11eb-9b10-89cd94382d54.png)

ihongda:
Thank you for your feedback @virtualvinodh :smile:

PR #2164, opened 2021-05-25
Fix: Transliteration:  Many of the script labels are gibberish #2115

ihongda:
On @virtualvinodh's advice, the appropriate English label was kept in the language list.

深度截图_选择区域_20210525001649

### DIFF
--- a/client/elements/addons/sc-aksharamukha-converter.js
+++ b/client/elements/addons/sc-aksharamukha-converter.js
@@ -55,69 +55,90 @@ export const paliScriptsStyles = css`
     font-family: 'Mukta Malar Regular' !important;
     font-feature-settings: 'ss04', 'kern';
   }
+
   .tamil-script {
     font-family: 'Noto Sans Tamil' !important;
   }
+
   .tibetandbumed-script {
     font-family: 'Tibetan Dbu Med' !important;
     line-height: 2.3em;
   }
+
   .tibetan-script {
     font-family: 'Noto Sans Tibetan' !important;
     font-size: 110%;
   }
+
   .buhid-script {
     font-family: 'Noto Sans Buhid' !important;
   }
+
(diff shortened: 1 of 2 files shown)


sujato:
Great, the dash problem is fixed, there are a few other minor issues. 

Sometimes the first script is not Latin.

Screenshot from 2021-05-26 18-42-21

When a right-to-left script is selected (eg. Kharosthi) it reverses the direction of all the scripts:

Screenshot from 2021-05-26 18-45-17

Sometimes the script shows in an odd font: too large or too small or a weird style. Proibably it just depends on the fallback fonts a user has on their machine, but it is odd.

Screenshot from 2021-05-26 18-46-57

I guess there's nothing can be done about this. Maybe have a look at what virtualvinodh does on hos site to handle the fonts for scripts?

And, as with the direction, when a script with weird font is selected, *all* the scripts try to display in that font.

Screenshot from 2021-05-26 18-50-15

- **styles including fonts and direction should be set per script, and not inherit from the selected script.**

virtualvinodh:
@ihongda @sujato I would remove the native names and only retain the English names. You cannot back transliterate from their English names back to their native names. It doesn't work that way. 

We must either properly research their original names to use them or just drop this back-transliteration altogether.  

Another reason is that it will force the user to download all the 40+ web fonts for the minority scripts just to display the drop-down menu.  It will slowdown the site. That's why I don't provide the native names of the scripts.

V

sujato:
Great, thanks for the advice, keep it simple is best!

ihongda:
Thx @virtualvinodh 

> Another reason is that it will force the user to download all the 40+ web fonts for the minority scripts just to display the drop-down menu. It will slowdown the site. That's why I don't provide the native names of the scripts.

Indeed!

Issue #2114, opened 2021-04-07
Incorrect transliteration of /e/ and /o/ in some scripts


virtualvinodh:
#### Description:

Incorrect transliteration of /e/ and /o/ in some scripts. They are currently transliterated as short vowels.

#### Steps to reproduce:

Any Pali/Sanskrit page with transliteration plugin.

#### Expected behavior:

ISO traditionally maps /e/ and /o/ to the short vowels in the languages that have them like South Indian languages.  In others, they are to be considered long by default.  

This has an issue when converting to scripts that have both short and long vowels. The plugin (which internally I assume is calling Aksharamukha) transliterates them as short vowels in scripts that differentiate long and short /e/ and /o/. This results in incorrect transliteration.  

I would recommend replacing /e/ and /o/ with /ē/ & /ō/ before calling the transliteration module. 

#### Screenshots:

![image](https://user-images.githubusercontent.com/1050975/113927288-92bf8380-97ed-11eb-8e5a-29b7f1c9aa84.png)

Ideally, this should have long e and o.

V

sujato:
Hi, thanks for the heads up!

To be clear, you're recommending replacing *all* instances of `e` and `o` in *all* scripts?

BTW, I was with the folks from Gandhari.org the other day, and recommended Aksharamukha to them as well!

virtualvinodh:
Hallo.

//To be clear, you're recommending replacing all instances of e and o in all scripts? //

Yup. That should work 

Sinhala is a bit tricky. They traditionally used short-e and short-o for Pali.  But I don't think it's the case in Modern Sinhala. Ye dhamma rendered traditionally  ""යෙ ධම්මා"" with short /e/ gives me 1000 results but "යේ ධම්මා" using long /e/ gives me 7000 results in Google.

If it is possible, I would split Sinhala into two:  Sinahla (Traditional) and Sinhala (Modern).  Traditional uses short /e/ and short /o/ and uses conjuncts. You would need to pass 'SinhalaPali' and 'SinhalaConjuncts' in post_options to achieve this when you call Aksharamukha. And Modern is just the default output.

I would also split Thai into two. Thai (Traditional) and Thai (Karaoke). For Karaoke, you would need to pass 'ThaiTranscription' in post_options. 

These are pretty much the two edge-cases. For everything else, the replacement should work out of the box with the default settings and give you a readable text for all scripts.

V

PR #2165, opened 2021-05-26
Fix: Incorrect transliteration of /e/ and /o/ in some scripts #2114

ihongda:


sujato (commented):
I'm afraid I'm not competent to assess this, let's ask virtualvinodh once it's on staging.

PR #2166, opened 2021-05-26
Fix: Transliteration: Many of the script labels are gibberish(2)

ihongda:
Only retain the English names in the pali script dropdown.


sujato (approved):
Okay, that looks good.

PR #2177, opened 2021-06-04
Fix: Sometimes translated titles are not being used.

ihongda:
Need to execute `make load-data`


sujato (approved):
<s>Looks good, translated titles appear in all cases I have checked.</s>

Actually just found a bug. See here, all the suttas have the same title

http://localhost/sn1-nalavagga


Screenshot from 2021-06-04 17-13-46

ihongda:
Thx Bhante, I'll fix it ASAP

PR #2180, opened 2021-06-04
Fix: When changing site language, everything goes crazy.

ihongda:


sujato (approved):
Great, works smoothly now

PR #2181, opened 2021-06-04
Fix: Sometimes translated titles are not being used.(2)

ihongda:
Fixed duplicate titles in some cases.

Need to execute: `make delete-database`, `make load-data`


sujato (approved):
seems to be working properly now.

PR #2186, opened 2021-06-07
Populate `volpage` and `alt_volpage` in `text_extra_info.json`.

ihongda:
Populate `volpage` and `alt_volpage` in `text_extra_info.json` from bilara-data/reference.

Need to execute: `make delete-database`, `make load-data`

### DIFF
--- a/client/elements/suttaplex/card/sc-suttaplex-css.js
+++ b/client/elements/suttaplex/card/sc-suttaplex-css.js
@@ -260,6 +260,11 @@ export const suttaplexCss = html`
       height: 16px;
       fill: var(--sc-icon-color);
     }
+
+    .hidden {
+      visibility: hidden;
+      width: 0px;
+    }
   </style>
 `;
 
--- a/client/elements/suttaplex/card/sc-suttaplex.js
+++ b/client/elements/suttaplex/card/sc-suttaplex.js
@@ -123,11 +123,27 @@ class SCSuttaplex extends LitLocalized(LitElement) {
     return pickVolPage(this.item.volpages);
   }
 
+  get briefVolPage() {
+    const volpages = this.item.volpages.split(',');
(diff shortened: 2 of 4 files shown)


sujato (changes requested):
Looking at the latest branch, I see:

```
<span class="vol-page nerdy-row-element" title="Volume and page">
                 …
        <span class="visible"><!---->PTS MN i 1<!----></span>
        <span class="hidden"><!---->PTS MN i 1<!----></span>
</span>
```
So the HTML structure is there, but the data is still from the *old* `text_extra_info` not from the `reference` files. You can tell the difference because the old files use the roman numeral "i".

As for the markup, I was wrong, I thought `visibility: hidden` would be available to `ctrl+f,` but it's not.

Instead, we can use something like:

```
width: 0px;
    opacity: 0;
```

Use both, because it's possible that due to a rounding error a 0px element might display in some cases.

This will leave the whole long list of IDs in the DOM tree where accessibility software will read it (probably). So add the following:

```
<span class="hidden" aria-hidden="true"><!---->PTS MN i 2<!----></span>
```

ihongda:
Thanks Bhante!

The current logic is that if there is relevant information in the `reference` file, it will be updated to the `text_extra_info`:

深度截图_选择区域_20210608084832

深度截图_选择区域_20210608084944

ihongda:
I have made changes based on the feedback. In addition, if the `text_extra_info` is still not updated after the execution of `make delete-database` `make load-data`, please give me feedback in time , thx!

ihongda:
Bhante, I have made changes based on the feedback. In addition, if the `text_extra_info` is still not updated after the execution of `make delete-database` `make load-data`, please give me feedback in time , thx!

ihongda:
I found a problem, I will submit the code after the fix

PR #2201, opened 2021-06-16
Fix High LCP in mobile.

ihongda:
A newly introduced API, `/api/tipitaka_menu`, reduces unnecessary data transmission.
Delete unnecessary code, reduce references, and speed up loading.


sujato (approved):
I can't rest this locally, so I guess we'll have to push to staging and see.

PR #2212, opened 2021-06-24
fix references to 3 characteristics of existence

benmneb:


sujato (approved):
Thanks for the fixes!

Issue #2139, opened 2021-04-30
table view for parallels


sujato:
The old SC had a massive table that showed all the parallels for a nikaya in brief. It's a forbidding wall of incomprehensible data that stands like a barrier against anyone who dares to try to read suttas without years of study beforehand.

It does, however, provide a neat way of overviewing the parallels data for those who want that. maybe we can provide a similar view option on the new site?

Let's call it the table view.

What might it look like?

## navigation

The table view is available for all of the same navigation steps that currently show a suttaplex list. It is activated with an icon in the context-toolbar, which sits next to the "spacing" icon.

There is a material icon for exactly this purpose:

https://fonts.google.com/icons?selected=Material+Icons:table_view

## content

Most of the suttaplex content would be omitted: 

- blurb
- reference info
- links and cite and Voice.
- translations and texts
- vaggas and other internal headings
- translated titles (this view is for experts)

Instead we just present ID, root title, and parallels.

![Screenshot from 2021-05-04 09-20-00](https://user-images.githubusercontent.com/6112010/116945045-1324b780-acba-11eb-866c-1e8370dec8f8.png)

[ll-table.zip](https://github.com/suttacentral/suttacentral/files/6403806/ll-table.zip)

## design

Because this is aimed at quick scanning by experts, aim for information density.

- page column is screen width, as wide as possible
- HTML is just a table.

## navigation

Each ID is a link, and each link simply goes to the suttaplex for that ID.

```
<a href='/dn1'>DN 1</a>
```

Parallels are the same:

```
<a href='/t21'>T 21</a>
```

PR #2218, opened 2021-07-01
Table view for parallels #2139

ihongda:


sujato (changes requested):
Okay, so this is pretty close, but there are a few changes to make. 

1. The view is meant to be for the suttaplex list page, not the header. In other words, it should not belong in `sc-top-sheet` but in `sc-suttaplex-list`.
2. Sutta `uid` is used (eg. `da21`) and then transformed to uppercase (DA21). NEVER DO THIS ANYWHERE! IT DOES NOT WORK! **Always** use the `acronym` (or `acro`). The form of the IDs must be exactly the same as they are in the normal parallels view, in the items marked with `title="SuttaCentral ID"`. In fact if possible we should reuse that data for simplicity and reliability.
3. In my mockup I gave the table a `<caption>` but I think this is not needed, the title is already in the toolbar. So we can delete the caption.
4. The parallel items should be comma-separated.

Note, when you click on an item it table view, it takes you to the suttaplex for that view, and the suttaplex should be the *normal* suttaplex view, not the table view. This is the current behavior, and it should not change.

sujato (changes requested):
I made requests for 4 changes. Here I address each one in turn based on the update on July 8.

1. Table view is still in `sc-top-sheet`. But it does not belong there, it is a new view for `sc-suttaplex-list`. It should be **on the page** not **in the toolbar**.
2. The IDs are mostly correct now, but in some cases we are using the raw `uid`, for example on DN 1 we have “SN 41.3 , pli-tv-kd21:26 , Vb 17:21”. Do you see which one does not match? *All* IDs must follow exactly the same form that they have in `sc-parallel-list`, i.e “Pli Tv Kd 21:26”.
3. Caption is removed, that’s good.
4. Items are comma-separated, but there is a space before the comma. Comma never takes a space before.

sujato (changes requested):
Great, this is looking good now, just a few small things to attend to.

- [x]  The icon is not always visible. In particular, if I refresh the page it usually vanishes. <br>**Icon must be always present on suttaplex-list**
- [x]  The state should be persistent: once I have selected “table view”, I should always see that on the suttaplex-list. This is the normal SuttaCentral policy: change things only when user requests them, then keep them that way until user changes them back. <br>**Make table view persistent**.
- [ ] There should be a way of indicating the state on the icon. Currently when “table view” is selected, there is no visual indication. Compare the “spacing” icon, which does change. The material design “table view” icon does not have a state indication, but material icons provide various options including “outlined” and “two-tone”.  I suggest using the “two-tone” icon for this. So the default view stays the same, using the regular “outlined” icon, then when you activate table view, it swaps in the “two tone” icon. <br>**Use two-tone icon to indicate active state of table view.**
- [x] Finally, there is still one small issue with the IDs: to indicate a range, they inconsistently use either hypen (MN 38:95-103) or end-dash (SA 101–200). <br>**Always use en-dash (–) for ranges**.

sujato (changes requested):
Okay, that is looking great. I have changed the two-tone icon to make it more prominent, using accent color on the SVG instead of opacity. 

I have found one small bug. Activate the table view; then refresh the page; then de-activate table view: the two-tone icon is still visible. Now if you toggle table view, the two-tone icon remains visible. **Two-tone icon should only be visible when table view is activated.**

sujato (approved):
looks good

sujato (changes requested):
looks good, except I noticed a small bug. Consider a table like this:

http://localhost/an3-balavagga

Here, most of the suttas do not have parallels. So they are missing from the table. But they should be there. In such cases, give the basic text ID and title, and just leave the parallels blank.

sujato (approved):
Looks good

PR #2226, opened 2021-07-21
Fix: Transliteration: Fonts missing for some scripts

ihongda:


sujato (approved):
looks good for me locally

PR #2233, opened 2021-07-31
Optimize module dynamic loading code.

ihongda:
Optimize the dynamic loading of modules to increase speed and reduce network requests and avoid loading failure 
The relevant modules are:
1. sc-suttaplex.js
2. sc-action-items.js
3. sc-top-sheet-views.js
4. sc-top-sheet-toc.js
5. sc-top-sheet-parallels.js
6. sc-top-sheet-publication-legacy.js
7. sc-top-sheet-publication-bilara.js

Verification: All modules should be loaded normally as before。

### DIFF
--- a/client/elements/addons/sc-top-sheet-parallels.js
+++ b/client/elements/addons/sc-top-sheet-parallels.js
@@ -19,7 +19,11 @@ class SCTopSheetParallels extends SCTopSheetCommon {
 
   connectedCallback() {
     super.connectedCallback();
-    import('../suttaplex/card/sc-suttaplex.js');
+    import(
+      /* webpackMode: "lazy-once" */
+      /* webpackPrefetch: true */
+      '../suttaplex/card/sc-suttaplex.js'
+    );
   }
 
   show() {
--- a/client/elements/addons/sc-top-sheet-publication-bilara.js
+++ b/client/elements/addons/sc-top-sheet-publication-bilara.js
@@ -4,7 +4,8 @@ import { typographyCommonStyles } from '../styles/sc-typography-common-styles';
 import SCTopSheetCommon from './sc-top-sheet-common';
 import { store } from '../../redux-store';
 
-class SCTopSheetPublicationBilara extends SCTopSheetCommon {
+// eslint-disable-next-line import/prefer-default-export
+export class SCTopSheetPublicationBilara extends SCTopSheetCommon {
   static get properties() {
     return {
       translationTitle: {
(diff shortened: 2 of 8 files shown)


sujato (approved):
Everything seems to work ok.

Not related to this change, but sometimes it seems as if the text pages load slowly, even on my local version. I wonder if this is because we are loading the contents of the toolbar, especially sc-action-items? Might we load the we could make sure the text loads first, then the icons in the toolbar, then the content of sc-action-items?

Anyway, maybe this is a bad idea, or maybe we are doing it already, in which case ignore it!

sujato:
I'm not sure if this is relevant, but I was just clicking around, I'm getting a *lot* of the following warnings:

```
An update was requested (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update.
```

sujato:
also the following warnings:

```
{
    "moduleIdentifier": "/opt/sc/frontend/node_modules/babel-loader/lib/index.js??ruleSet[1].rules[0].use!/opt/sc/frontend/elements/sc-page-search.js",
    "moduleName": "./elements/sc-page-search.js",
    "loc": "29:0-33:35",
    "message": "`webpackMode` expected 'lazy', 'eager' or 'weak', but received: lazy-once."
}
```

ihongda:
> Everything seems to work ok.
> 
> Not related to this change, but sometimes it seems as if the text pages load slowly, even on my local version. I wonder if this is because we are loading the contents of the toolbar, especially sc-action-items? Might we load the we could make sure the text loads first, then the icons in the toolbar, then the content of sc-action-items?
> 
> Anyway, maybe this is a bad idea, or maybe we are doing it already, in which case ignore it!

I will check to see what can be optimized :smile:

ihongda:
> I'm not sure if this is relevant, but I was just clicking around, I'm getting a _lot_ of the following warnings:
> 
> ```
> An update was requested (generally because a property was set) after an update completed, causing a new update to be scheduled. This is inefficient and should be avoided unless the next update can only be scheduled as a side effect of the previous update.
> ```

Here are some optimization tips for the new version of lit, I am trying to optimize the relevant code.

ihongda:
> also the following warnings:
> 
> ```
> {
>     "moduleIdentifier": "/opt/sc/frontend/node_modules/babel-loader/lib/index.js??ruleSet[1].rules[0].use!/opt/sc/frontend/elements/sc-page-search.js",
>     "moduleName": "./elements/sc-page-search.js",
>     "loc": "29:0-33:35",
>     "message": "`webpackMode` expected 'lazy', 'eager' or 'weak', but received: lazy-once."
> }
> ```

The parameter `lazy-once` exists in the webpack document, so I don’t know why this prompt appears, I will check it.

PR #2245, opened 2021-08-14
Rewrite the breadcrumb related code.

ihongda:
1.Speed up the breadcrumb loading.
2.Reduce the chance of errors.

### DIFF
--- a/client/elements/navigation/sc-navigation-common.js
+++ b/client/elements/navigation/sc-navigation-common.js
@@ -158,11 +158,9 @@ export function setCurrentNavPosition(position) {
   });
 }
 
-async function genNavDetail(uid, navType, currentURL) {
-  const navArray = store.getState().navigationArray;
-  const url = `${API_ROOT}/menu/${uid}?language=${store.getState().siteLanguage || 'en'}`;
-  fetch(url)
-    .then(r => r.json())
+function genNavDetail(navType, currentURL, data, navArray) {
+  data
+    ?.json()
     .then(menuData => {
       if (!menuData || !menuData[0].uid) {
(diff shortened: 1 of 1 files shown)


sujato (commented):
I have got it running locally, I had to use `make run-preview-env`. It all seems fine, no error messages.

In terms of the slow response problem, I may have identified something.

- Navigate sutta > minor > kn > dhp > 1st chapter
- then the page is blank for maybe ten or twenty seconds? 
- meanwhile, my fan starts running fast, clearly something is churning in the background.
- finally the pages loads and the fan goes quiet.

Same thing happens if I navigate to here:

http://localhost/uvs1/san/nakatani

It takes maybe ten seconds to load, for only a couple of lines.

And again to here:

http://localhost/pdhp1-13/pra/cone

Once again, it took ten seconds to load (I timed it this time!), causing the fan to start up in the background.

Now with these pages, if I go back to them, they load right away. And it seems that once I have loaded a part of a text, other parts of the same text load in reasonable speed after that.

another new text

http://localhost/t210.1/lzh/taisho

This one took 13s to load.

Now loading this page, it took over 20s.

http://localhost/uv-kg1/en/rockhill

Here's the summary of js time for this page:

- 2 ms  Loading
- 286 ms  Scripting
- 1144 ms  Rendering
- 646 ms  Painting
- 1218 ms  System
- 21899 ms  Idle
- 25195 ms  Total

So yeah, 20+s idle is a big problem!

ihongda:
>So yeah, 20+s idle is a big problem!

I will try to lazy load the parallels data to see if this situation can be improved.

PR #2247, opened 2021-08-17
Configure Ngnix proxy_cache.

ihongda:


nginx/conf.d/dev.nginx.conf
```diff
@@ -1,10 +1,13 @@
+proxy_cache_path /etc/nginx/cache keys_zone=SC_CACHE:30m levels=1:2 inactive=30d max_size=1g;
```
blake-sc: I think it would be better to create a named volume for the nginx cache rather than storing the data in the container and use an /opt/ path, compare how nginx-data-volume (which should be renamed to nginx-logs) is used in docker-compose.yml.

nginx/conf.d/dev.nginx.conf
```diff
@@ -1,10 +1,13 @@
...
+
 server {
     ssl_certificate /etc/nginx/configurations/certs/server.pem;
     ssl_certificate_key /etc/nginx/configurations/certs/server.key;
+    proxy_cache SC_CACHE;
     include "/etc/nginx/configurations/sc.nginx.conf";
 
     location / {
-
+        proxy_cache_valid 5m;
```
blake-sc: No need to cache /, see below

nginx/conf.d/prod.nginx.conf
```diff
@@ -72,6 +75,7 @@ server {
     }
 
     location / {
+        proxy_cache_valid 30d;
```
blake-sc, line 78: I don't think / needs to be cached. It simply serves files anyway.

nginx/conf.d/prod.nginx.conf
```diff
@@ -37,6 +37,7 @@ map $uri $cache_control {
 
 add_header Cache-Control $cache_control;
 
+proxy_cache_path /etc/nginx/cache keys_zone=SC_CACHE:30m levels=1:2 inactive=30d max_size=1g;
```
blake-sc, line 40: We might want the cache to be bigger than 1g, that sounds close to what the all the offline builds would be, 5g  would be safe, at the moment we have 33g available on the server.

blake-sc (changes requested):

nginx/conf.d/dev.nginx.conf
ihongda: Hey, Blake, I have added the nginx-caches volume in docker-compose, but I don't know how to apply it to the nginx configuration, can you help me see :)

https://github.com/suttacentral/suttacentral/pull/2247/commits/48ea320bb21863eec632ed79ba988ba32e97388b

blake-sc: That should be fine.

blake-sc (approved):

nginx/conf.d/dev.nginx.conf
ihongda: Thx Blake :)

PR #2249, opened 2021-08-18
Add indexs to ArangoDB collections.

ihongda:


blake-sc (changes requested):
An old migration should never be modified because "make migrate" will only ever run a migration once, and won't re-run a modified migration. This results in great confusion and bugs.

Instead a new migration should be created which makes the changes. You can use an old one as a template. An example name would be "add_missing_indexes_042".

Also in ArangoDB "hash" index is now actually a persistent index, and the name "hash" is only kept for backwards compatibility reasons, it actually makes a persistent index, as such for new migrations we should use "add_persistent_index".

ihongda:
Thx @blake-sc 
Has been modified according to the problem you pointed out, the local test passed :smile:

blake-sc (approved):

sujato:
works great for me, no more lag!

PR #2250, opened 2021-08-18
I realized the wrong module was used, uwsgi_pass should use uwsgi_cache not proxy_cache.

blake-sc:
Tuned the cache settings, adding cache lock and an explicit key

For testing added a dev setting which deletes cache entries after 1 minute


ihongda (approved):

ihongda:
Local test, related directories have generated cache files.
Thx @blake-sc :tada:

PR #2268, opened 2021-09-15
Let users share state of text by extending URLs.

ihongda:
e.g.
http://localhost/sn1.6/en/sujato?layout=column
http://localhost/sn1.6/en/sujato?layout=row
http://localhost/sn1.6/en/sujato?script=Ahom
http://localhost/sn1.6/en/sujato?notes=tooltip
http://localhost/sn1.6/en/sujato?notes=sidenote
http://localhost/dn1/en/sujato?reference=pts/main
http://localhost/sn1.6/en/sujato?layout=row&notes=tooltip&script=Cham&reference=pts/main

### DIFF
--- a/client/elements/addons/sc-redux-actions.js
+++ b/client/elements/addons/sc-redux-actions.js
@@ -1,0 +1,39 @@
+/* eslint-disable import/prefer-default-export */
+import { store } from '../../redux-store';
+
+export class reduxActions {
+  static setShowHighlighting(showHighlighting) {
+    store.dispatch({
+      type: 'SET_SHOW_HIGHLIGHTING',
+      showHighlighting,
+    });
+  }
+
+  static setReferenceDisplayType(displayedReferences) {
+    store.dispatch({
+      type: 'SET_REFERENCE_DISPLAY_TYPE_ARRAY',
+      displayedReferences,
+    });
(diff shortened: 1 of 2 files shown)


ihongda:
Bhante @sujato , in this case, what is the processing logic?
https://staging.suttacentral.net/sn56.1/en/sujato?lang=pli+de

Different languages：
sn56.1/en and lang=pli+de

sujato:
Sorry I didn't answer earlier!

>Bhante @sujato , in this case, what is the processing logic?
>https://staging.suttacentral.net/sn56.1/en/sujato?lang=pli+de

The defining URL must always take precedence; that is, whatever is before the question mark. That tells us what *thing* we are dealing with. The rest tells us *how to view that thing*, but it doesn't change the essence of it.

In this case, we show `sn56.1/en/sujato`. Since we cannot (currently) show more than one translation language, what comes after the question mark is invalid and has no effect.

In fact, I would say we should avoid having a `lang` option at all for this reason. Instead, use a boolean for showing the root:

```
http://localhost/dn1/en/sujato?reference=pts/main&root=yes
```

At some point in the future, we may introduce an option for showing multiple translation languages. If that is the case, then we can apply the `lang` option.

PR #2271, opened 2021-09-28
Apply: patimokkha texts: a new model.

ihongda:


sujato (changes requested):
I'm checking the pli-tv-bu-pm to start with. 

http://localhost/pli-tv-bu-pm

- Okay, so the display is fine, except the button has no spaces between the words; it just takes `showParallelsAndDetails` and make it uppercase!
- In compact view, there is no space between the button and the suttaplex, it should have `mwc-button{margin-top: 1em}`.
- But basically, great, we have the suttaplex without parallels, that's what we want.

Alright, so next we click the button and load the rules and parallels. 

- This shows the suttaplexes with parallels but without text, which again is what we want. Good!
- One problem, the list includes suttaplexes for both the *individual rules* and the *rule categories*. **Only show suttaplexes for the rules**.
    - (hint: IDs for rules end with a number, rule categories do not.)
- However! The **rule category names should appear as a `vagga-node`**.
     - (In the Vibhangas, we already do this; for example, here the rule category "Expulsion" is a `vagga-node` : http://localhost/pli-tv-bu-vb-pj.)
- Unlike a normal `vagga-node`, however, we should **display both root and translation for each rule category name**.
    - in studying Vinaya, it is common to refer to the rules by the Pali names, so they should be present. In the Vibhangas, the rule category names are found in the toolbar. But in the patimokkhas, they are found nowhere else, so they have to be in the `vagga-node`.

Here's a mockup:

Screenshot from 2021-09-29 07-58-13

ihongda:
Thx Bhante!
>Okay, so the display is fine, except the button has no spaces between the words; it just takes showParallelsAndDetails and make it uppercase!

深度截图_选择区域_20210929172147

This may be a problem with the localization file loading. It is normal in my local area. Let me check

The others I am modifying.

sujato (changes requested):
Looking at the latest update, mostly it is great. Two minor issues.

- for the node-headings, use em-dash (—) not en-dash (–). Em-dash separates text, en-dash is a range of numbers.
- in the lists of parallels, the first one or two seem to be a bug: they are just a number, no text or anything.anything.

sujato (approved):
Great, all issues appear to be fixed.

ihongda:
orphan data is displayed in this way(The original display is a number):

 
深度截图_选择区域_20211001210233

sujato:
That looks perfect.

PR #2272, opened 2021-10-04
When the user modifies the view settings, update the URL

ihongda:


sujato (approved):
that looks great, congrats.

Issue #2314, opened 2021-10-29
ATB and title for static page essays is buggy


sujato:
On staging static pages, the titles has disappeared from the toolbar, and the ATB is not reliable. Here is an example, but all pages must be tested.

![Screenshot from 2021-10-30 09-00-56](https://user-images.githubusercontent.com/6112010/139506369-308029c8-1c84-4426-b900-000cd34e6044.png)

Issue #2313, opened 2021-10-29
The ATB for the Tipitaka essays is buggy


sujato:
Each of the three main Pitakas has an attached essay, navigable from the Home page card.

https://suttacentral.net/discourses-guide-sujato
https://suttacentral.net/vinaya-guide-brahmali
https://suttacentral.net/abhidhamma-guide-sujato

When you click on any one of these, it should take you to the relevant essay. This works fine.

But when you are on that essay, it should display the *three* essays in the Adaptive Tab Bar (ATB).

![Screenshot from 2021-10-30 08-21-18](https://user-images.githubusercontent.com/6112010/139503185-5bd51a74-a828-4cce-96fc-ed62c83cb6cd.png)

However this does not work reliably, sometimes it has the wrong items in the ATB:

![Screenshot from 2021-10-30 08-23-08](https://user-images.githubusercontent.com/6112010/139503428-53a35ec1-d6cd-42f7-93f7-4f0dab3912bd.png)

-  **When on a Tipitaka essay, the ATB must always display the three Tipitaka essays.**

sabbamitta:
>However this does not work reliably, sometimes it has the wrong items in the ATB:

Not only this. Sometimes, the items look correct, but the attached links are not.
![Screenshot from 2021-10-29 22-54-11](https://user-images.githubusercontent.com/43345786/139524083-085d11a4-115e-45ec-bd30-5859181231a4.png):

When you are in the Sutta essay, the links to Vinaya and Abhidhamma go to a page with suttaplex cards instead of the essays.

When you are in one of the other essays, clicking on the link to the Sutta essay ("Lehrreden") leads into nowhere land ("null" in the tab …):
![Screenshot from 2021-10-30 09-17-47](https://user-images.githubusercontent.com/43345786/139524234-5a3c46b1-ae5c-4693-a3a1-54c833a6e591.png)

PR #2315, opened 2021-10-30
FIx: ATB and title for static page essays is buggy

ihongda:
Related issues:
https://github.com/suttacentral/suttacentral/issues/2313
https://github.com/suttacentral/suttacentral/issues/2314

### DIFF
--- a/client/elements/sc-page-selector.js
+++ b/client/elements/sc-page-selector.js
@@ -572,9 +572,9 @@ class SCPageSelector extends LitLocalized(LitElement) {
       case 'sutta':
         return;
       default:
-        const key = `${this.currentRoute.name}Title`;
+        const key = `interface:${this.currentRoute.name}Title`;
         if (this.__resources[key]) {
-          const pageNameTitle = this.localize(`interface:${key}`);
+          const pageNameTitle = this.localize(key);
           this.actions.changeToolbarTitle(pageNameTitle);
         } else {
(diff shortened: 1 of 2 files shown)


sujato (changes requested):
There are still some bugs.

*Running branch Hongda-work in Incognito, site language English, clear site data first*.

- Click on the three essays from the Home page, all works fine.
- The problem with missing titles is fixed (#2314)
- **But!** Once on an essay page, if I refresh, the incorrect ATB comes up (i.e. it has INTRODUCTION | DONATIONS etc.)
- The same thing happens if I am on another static page, eg. `/methodology`. The ATB is correct when I navigate there, but is incorrect after refresh.
- Testing in German: change site language to `de`, German essays don't appear at all, it's just in English.

ihongda:
>Testing in German: change site language to de, German essays don't appear at all, it's just in English.

Bhante, Where is the German version of essays?
It seems that there are only these three essays at the moment:
 
`http://localhost/discourses-guide-sujato`
`http://localhost/vinaya-guide-brahmali`
`http://localhost/abhidhamma-guide-sujato`

ihongda:
>Testing in German: change site language to de, German essays don't appear at all, it's just in English.

Since the localization method has been modified, some modifications need to be made to adapt, I will open another issue.

sujato (approved):

PR #2325, opened 2021-11-15
Implement comments for segmented texts.

ihongda:
If want to test, need to manually copy the comment file to the `sc-flask` container.
e.g. `docker cp /download/sc_bilara_data/comment sc-flask:/opt/sc/sc-flask/sc-data/sc_bilara_data`
and then execute `make load-data-no-pull`

### DIFF
--- a/client/elements/styles/sc-layout-bilara-styles.js
+++ b/client/elements/styles/sc-layout-bilara-styles.js
@@ -46,6 +46,11 @@ export const plainStyles = html`
       display: none;
     }
 
+    .verse-line .translation {
+      display: block;
+      margin-left: 2em;
+    }
+
     /* Set styles for tooltip marker. First we hide the actual content. These settings ensure the beginning, i.e. the :before content, is visible and the rest is hidden. Height is important to maintain even line-height. */
     .comment,
     .variant {
@@ -74,6 +79,11 @@ export const plainPaliStyles = html`
     section,
     article {
       max-width: 720px;
+    }
+
+    .verse-line .root {
+      display: block;
+      text-indent: 2em;
     }
 
     .comment,
@@ -146,6 +156,11 @@ export const plainPlusStyles = html`
 
       grid-column: 2;
       grid-row: 1;
+    }
+
+    .verse-line .translation,
+    .verse-line .root {
+      text-indent: 2em;
     }
 
     @media only screen and (max-width: 600px) {
@@ -236,6 +251,11 @@ export const sideBySideStyles = html`
       grid-column-gap: var(--sc-size-lg);
     }
 
+    .verse-line .translation,
+    .verse-line .root {
+      text-indent: 2em;
+    }
+
     .translation {
       grid-column: 1;
     }
@@ -308,6 +328,11 @@ export const sideBySidePlusStyles = html`
 
       grid-template-columns: 60px 1fr 1fr;
       grid-column-gap: var(--sc-size-lg);
+    }
+
+    .verse-line .translation,
+    .verse-line .root {
+      text-indent: 2em;
     }
 
     .reference {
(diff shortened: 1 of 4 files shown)


sujato:
HTML is visible in comments.

Screenshot from 2021-11-27 07-46-48

sujato:
## no `data-tooltip`

There is no `data-tooltip` in the comments, and hence nothing to display on hover when "tooltip on asterisk" is set.

It should be the same as the variants:

```
<span class="variant" data-tooltip="Variant: Yo → yo ve (sya-all)">Variant: Yo → yo ve (sya-all)</span>
```

Instead we have:

```
<span class="comment">Both Bodhi and Norman implicitly accept the commentarial reading here, i.e. that this refers to “wrong thoughts”. However the identical line at <a href="/ud6.7/en/sujato#4.1">Ud 6.7:4.1</a> clearly refers to jhāna. Both also follow the commentary in reading <i lang="pi" translate="no">vidhūpitā</i> as “burning up”, but <i lang="pi" translate="no">vidhūpa</i> refers to clearing the mist, smoke, or incense.</span>
```

There is also a limitation in this approach: since this is content that will be displayed via a `:before` pseudo-element, it cannot render any HTML. So all content in the `data-tooltip` must be stripped of all HTML and be only plain text.

```
<span class="comment"  data-tooltip="Both Bodhi and Norman implicitly accept the commentarial reading here, i.e. that this refers to “wrong thoughts”. However the identical line at Ud 6.7:4.1 clearly refers to jhāna. Both also follow the commentary in reading vidhūpitā as “burning up”, but vidhūpa refers to clearing the mist, smoke, or incense.">Both Bodhi and Norman implicitly accept the commentarial reading here, i.e. that this refers to “wrong thoughts”. However the identical line at <a href="/ud6.7/en/sujato#4.1">Ud 6.7:4.1</a> clearly refers to jhāna. Both also follow the commentary in reading <i lang="pi" translate="no">vidhūpitā</i> as “burning up”, but <i lang="pi" translate="no">vidhūpa</i> refers to clearing the mist, smoke, or incense.</span>
```

sujato:
Comments are now appearing in the tooltip, good!

But the HTML has not been stripped.

Also there's some problem with the routing: the browser back button is not working properly. 

- Go to sutta with a note (http://localhost/snp4.1/en/sujato?layout=plain&reference=none&notes=sidenotes&highlight=false&script=latin)
- click on reference in sutta, this is correct: http://localhost/mnd1/pli/ms?layout=plain&reference=none&notes=sidenotes&highlight=false&script=latin
- Now the **top sheet** appears, this should never happen, top sheet should only appear when user requests.
- click "back", it goes to the **plain url**for that text (i.e.  http://localhost/mnd1/pli/ms), not to the previous text

It looks like the browser is interpreting the added URL details (?layout=plain&reference=none&notes=sidenotes&highlight=false&script=latin) as an extra page transition.

- [ ] strip HTML from comments tooltip
- [ ] make sure back button in browser always works as expected

ihongda:
> Also there's some problem with the routing: the browser back button is not working properly.

Bhante, I tried it in several browsers and it seems normal :
https://file.io/W8bVK3MxDsQo

sujato:
Okay, thanks, now it works almost as expected. The only issue that I can now see is the handling of the bookmark `#` at the end of the URL.

- Expected behavior: when a link contains a bookmark (eg `http://localhost/sn35.94/en/sujato#6.4`), clicking it will take you to the correct place in the text. If references are visible, the reference is highlighted.

But if I click the above link, it goes to:

```
http://localhost/sn35.94/en/sujato?layout=sidebyside&reference=none&notes=sidenotes&highlight=false&script=latin
```

I.e., the bookmark has disappeared.

ihongda:
> The only issue that I can now see is the handling of the bookmark # at the end of the URL.

Strangely, this problem does not exist on my local machine's chrome, edge, and Firefox browsers. I will merge them first and wait to push to staging to observe.

PR #2330, opened 2021-11-24
reduce sensitivity for toolbar appearance.

ihongda:
Modify the response time of scroll events from 500 milliseconds to 800 milliseconds.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 
 ## 2021.11.19
 
-- FIx: ATB and title for static page essays is buggy. #2315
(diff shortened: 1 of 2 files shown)


sujato:
I don't think this solves the problem. It seems that it's not the *time* but the *distance* that creates problems. 

If I scroll up just one notch on my mouse, or just a random movement of the finger on a touchscreen, then the toolbar re-appears. It should require something more definite, a longer upwards scroll before it comes. 

Can we double the distance that it takes to activate?

ihongda:
>Can we double the distance that it takes to activate?

Let me try :)

PR #2333, opened 2021-11-26
Execute the scroll event according to the scroll distance. 

ihongda:


sujato (commented):
Hmm, I still can't perceive much difference. Can you try decreasing the sensitivity again?


Is the a variable I can play with to test it out locally?

ihongda:
> Hmm, I still can't perceive much difference. Can you try decreasing the sensitivity again?
> 
> Is the a variable I can play with to test it out locally?

Bhante, This variable can be modified to reduce sensitivity:
Line 425 `if (scrollDistance !== 16) {`
Take 4 as a multiple, for example, modify it to 20 or 24

sujato:
Okay, well, it's hard to tell without testing on mobile. But I think setting it to 24 seems better? Let's try that out on staging, and test it on mobiles.

Meanwhile, part of the issue here is that different people use it differently, so I'm going to propose another option in a separate issue.

PR #2342, opened 2021-12-08
Try lift main content into light DOM.

ihongda:
The segmented text page and legacy text page now instead of using light DOM rendering.


sujato:
That's looking good, I can't see any problems!

sujato:
So the content is still under several layers of ShadowDOM. This would explain why Weava doesn't work.

Screenshot from 2021-12-10 07-29-05

ihongda:
Currently `sc-page-selector.js` and `sc-site-layout.js` need to be modified, but if these two files are modified, many related styles need to be adjusted. The `:host` indicator is invalid for lightDOM, so all relevant places are To be modified.

sujato:
> The :host indicator is invalid for lightDOM, so all relevant places are To be modified.

Okay. Well, we can also take the chance to modernize our CSS, ensure that we use `return css` rather than 

```
return html`
      <style>
```

ihongda:
> > The :host indicator is invalid for lightDOM, so all relevant places are To be modified.
> 
> Okay. Well, we can also take the chance to modernize our CSS, ensure that we use `return css` rather than
> 
> ```
> return html`
>       <style>
> ```

According to the official document:
`The styles you add to your component are automatically scoped to the shadow root and only affect elements in the component's shadow root.`

So,  if change to light DOM rendering, the style sheet needs to be moved out of `return css`

ihongda:
Bhante @sujato 
The `sc-site-layout.js`, `sc-page-selector.js`, 'sc-text-page-selector.js' has been instead of Light DOM rendering.
Everything seems to be normal,
If there is no problem, I will rewrite all static pages.

sujato:
I've tested it, it looks just fine, and Weava works! So yes, let's go ahead.

sujato (changes requested):
- The colors for the toolbar select are now fixed
- The static pages are still inside a shadowDOM.
- there are a bunch `TypeError` and `failed to fetch resource` errors

ihongda:
There are more than 20 other static pages being modified :)

ihongda:
1. Adjust the structure of `sc-site-layout.js`, pull away the static page navigation menu code to an independent component.
2. After adjusting the code structure, modifying the static page to Light DOM rendering will no longer affect other pages.

sujato (changes requested):
Everything seems good, the pages select and work just fine. A couple of details needs attending, however.


- There are some js errors being reported in the console.
- The color definitions are being imported twice, and overridden. We should avoid this, styles should only be declared as necessary, we should not duplicate code. This problem is on the production site as well, so it is not caused by the changes. But now is a chance to fix it! 
- On the Home page, some of the styles are not being applied correctly, especially the font-family. The default should be sans-serif (defined in sc-static-home.js) but it's serif in most places. There are more subtle instances as well. See, for example, the titles on the nav cards: on production they have the correct line-height, on the lightDOM page the line-height is too big. It seems the culprit is the `sc-page-selector` styles in `sc-page-selector.js`.

Screenshot from 2021-12-21 07-11-29

ihongda:
Strangely, after inspection, the color definition is indeed only loaded when the sc-site-layout component is loaded, and it is not loaded again elsewhere.

sujato (approved):
all looks good

PR #2343, opened 2021-12-08
Serve range suttas per-sutta.

ihongda:
e.g.
http://localhost/an1.71-81/pli/ms?layout=sidebyside&reference=main&notes=sidenotes&highlight=false&script=latin

Current, the above URL can be accessed normally, 
but when the user uses the URL:
http://localhost/an1.72/pli/ms?layout=sidebyside&reference=main&notes=sidenotes&highlight=false&script=latin
the relevant scriptures cannot be displayed normally.

This PR solves this problem.


sujato:
Using the links provided above, the link to an1.72 shows the correct content, but the URL auto-changes to show the range, an1.171-81. We must keep the URL showing the actual content of the page. I.e. if the sutta is an1.72, the url must be an1.72.

ihongda:
> Using the links provided above, the link to an1.72 shows the correct content, but the URL auto-changes to show the range, an1.171-81. We must keep the URL showing the actual content of the page. I.e. if the sutta is an1.72, the url must be an1.72.

It has been modified to: 
the URL do not auto-changes to show the range.

sujato (changes requested):
Okay, so the URL now works properly. But the page still doesn't work 100%. *All* of the page data must correctly represent the fact that this is the page for *this sutta*. 

Remember, we don't want something half-baked. A page served per-sutta or per-range must be exactly identical in all respects. It's just a different slice of content, that's all.

Here I am taking AN 1.72 as the example.

From top to bottom:

- [x] Page `<title>` attribute is sutta UID, not vagga: 
     - `<title>AN 1.72—Mahāsaṅgīti Tipiṭaka Buddhavasse 2500</title>`
- [x] Breadcrumbs menu must finish with the sutta UID, not the vagga, eg. 
    - `Home / Kalyāṇamittādivagga / AN 1.72`
- [x] Toolbar title is AN 1.72 not vagga name
- [x] Info icon is missing, it must be included. It is the same as normal.
- [x] Table of Contents (toc) should be removed. It is only used for pages that have an internal heading structure.
- [x] Parallels are missing, they should be the same as normal.
- [x] Main page `<h1>` heading is the sutta UID (unless there is a title), i.e. AN 1.72.
- [x] `<h2>` subheading should be removed. 
- [x] Prev/next navigation is missing. It should go to the previous and next *sutta*, not range, i.e. AN 1.71 and AN 1.73

Screenshot from 2022-01-05 17-17-27

ihongda:
Bhante, I have modified it according to the feedback, you can check it when you have time :)

sujato:
This is looking much better. Still a couple of issues:

- [ ] The "info" button isn't showing for me.
- [ ] The prev/nav works, but only so far as the vagga range goes. The logic should be, that once you have navigated to an individual sutta, it always woks forward and back for individual suttas. So an1.71 goes to an1.70, an1.81 goes to an1.82

Next thing is, the solution so far seems to not yet work with all suttas.

Here is the complete list of range suttas. We need to test that it works with all these.

- an1.1-10, an1.11-20, an1.140-149, an1.150-169, an1.170-187, an1.209-218, an1.21-30, an1.219-234, an1.235-247, an1.248-257, an1.268-277, an1.278-286, an1.287-295, an1.296-305, an1.306-315, an1.31-40, an1.316-332, an1.333-377, an1.378-393, an1.394-574, an1.41-50, an1.51-60, an1.575-615, an1.61-70, an1.616-627, an1.71-81, an1.82-97, an1.98-139
- an2.1-10, an2.11-20, an2.118-129, an2.130-140, an2.141-150, an2.151-162, an2.163-179, an2.180-229, an2.21-31, an2.230-279, an2.280-309, an2.310-479, an2.32-41, an2.42-51, an2.52-63, an2.64-76, an2.77-86, an2.87-97, an2.98-117
- an3.156-162, an3.163-182,
- an11.22-29
- sn12.83-92, sn12.93-213
- sn33.16-20, sn33.51-54
- sn35.33-42, sn35.43-51
- sn43.14-43
- sn45.104-108, sn45.110-114, sn45.116-120
- pli-tv-bi-vb-sk1-75

I have tested some of them, and here is what I find. One thing to note is that in some cases (eg. an3.165) there is no content. This is because the text is abbreviated.

- generally the an and sn seem to work.
- sn12.83 is odd: there is no content shown for the sutta, yet it does have content.
- pli-tv-bi-vb-sk1 and pli-tv-bi-vb-sk75 don't work. (In this text, there is *only* numbers one and seventy-five.)

I'm looking at dhp, and I realize that the markup is wrong. Arrg! Somehow it hasn't been made properly into range suttas. I'll fix this and then we can test it.

ihongda:
>The "info" button isn't showing for me.
The current logic is that if the language is not equal to pli, and there is relevant publication information, this button will be displayed 
e.g.
http://localhost/an1.72/en/sujato?layout=sidebyside&reference=main/bj/cck/km/mr&notes=sidenotes&highlight=false&script=latin

The following pages will not display the publication information button because the language is equal to pli :
http://localhost/an1.72/pli/ms?layout=sidebyside&reference=main/bj/cck/km/mr&notes=sidenotes&highlight=false&script=latin#1.1

ihongda:
So far, the relevant range of suttas appears to be in service :)

sujato:
> The current logic is that if the language is not equal to pli, and there is relevant publication information, this button will be displayed e.g. 

Ok, so this is a pre-existing bug! *All* texts should have some "info". I will make a separate issue for this.

Let me test some of the range texts. Most of them seem to be okay. A couple of small issues.

##  sn12.93-213 

Hmm, this is a hard one. It is in fact a range of ranges! 

Screenshot from 2022-01-31 10-43-40

The individual suttas have no content
http://localhost/sn12.93/en/sujato?layout=plain&reference=none&notes=asterisk&highlight=false&script=latin

And the "inner ranges" return an error:

http://localhost/sn12.93-103/en/sujato?layout=plain&reference=none&notes=asterisk&highlight=false&script=latin

Not sure what to do here! If it's a simple fix, great. If not, leave it for the future.

## dhp

Looks good, except that  the `<h1>` heading is the the *vagga* title. 

Remember, the `<h1>` is always the main title of *that specific thing*. If that is the *vagga* (eg. dhp1-10), then it is the `range-title`. But here, the main content is the verse, not the vagga. So the `<h1>` should not be the `range-title` but the `sutta-title`. 

In other cases this seems to be correct.

ihongda:
>The individual suttas have no content
>[http://localhost/sn12.93/en/sujato?layout=plain&reference=none¬es=asterisk&highlight=false&script=latin](http://localhost/sn12.93/en/sujato?layout=plain&reference=none&notes=asterisk&highlight=false&script=latin)

Bhante, This problem have been fixed.

sujato:
> This problem is expected to require modifying the html structure of dhp.

Ok, let's leave it as-is. We'll see how it works in production, and if there is a need we can add further changes.

sujato (approved):

PR #2345, opened 2021-12-09
Try to use Cloudflare web analytics.

ihongda:


blake-sc:
I think it would be better to let Cloudflare insert the javascript snippet, as we are proxied through Cloudflare.

Though either way, we need to update the CSP in Ngnix, I'll do this.

ihongda:
I have enabled the automatic setting of web analytics in Cloudflare :)

sujato:
Awesome, please merge this when you are satisfied that it is ready.

PR #2354, opened 2022-01-04
Identify user language on static pages.

ihongda:


sujato (changes requested):
Okay, so the URL changes properly if you change it on the Home page and then navigate to a static page.

But if you are already on a static page, eg. http://localhost/numbering?lang=en

then you change the language to German, the page content changes correctly, but the URL does not change. If you then refresh, it reverts to English.

However, if you change the language, then navigate to a different static page, the language updates correctly: http://localhost/abbreviations?lang=de

Then when you return it is still correct: http://localhost/numbering?lang=de

But if you change back to English, there is the same problem.

Screenshot from 2022-01-05 18-39-37

sujato (approved):
good!

PR #2361, opened 2022-01-13
Update fonts, use `woff2` to instead of `ttf,otf`.

ihongda:


sujato:
Hongda, sorry I haven't got around to this, I will soon!

Meanwhile, can you add a new language: Gujarati! We need to add the Noto serif and sans-serif fonts. I think I have made the other needed changes already, as per:

https://github.com/suttacentral/suttacentral/wiki/add-language-to-website

sujato:
Finally, here are the latest Skolar fonts. These are variable fonts in woff2, both serif and sans-serif.

[2022_skolar_update.zip](https://github.com/suttacentral/suttacentral/files/7954339/2022_skolar_update.zip)

There's also a nice size reduction with these fonts, all up they're about 300kb smaller than before. Better optimization I guess.

ihongda:
Bhante @sujato , the `Skolar` fonts has been updated.

sujato:
Ok, let's merge it.

PR #2386, opened 2022-02-08
Fix:Info not present on root texts.

ihongda:


sujato (approved):
Excellent, this looks great.

PR #2399, opened 2022-02-10
Fix: Deep links on legacy texts broken.

ihongda:


sujato (changes requested):
This now works *if* the "main" references are set in the "views" panel. But unlike the segmented texts, the URL does not determine these settings. So if "references" is set to "none", then we still have no deep linking.

Ideally, the URL would drive the settings for the text, as it does for the segmented texts. Thus no matter what the user previously selected, when entering the URL:

```
http://localhost/mn9/en/bodhi?reference=main#sc17
```

Then the "main" references would be selected and it would work properly.

**BUT!!!**

Most of the controls present in the URL are not relevant for the legacy texts. In fact, it is only `reference` and `highlight` that apply. But currently the legacy texts URLs display the full list of options.

```
https://suttacentral.net/mn1/en/bodhi?layout=plain&reference=none&notes=asterisk&highlight=false&script=latin
```

- We should remove `layout`, `notes`, and `script` from all legacy text URLs. 
- Then make sure `highlight` and `reference` work in the same way as segmented texts.

ihongda:
- [ ] We should remove layout, notes, and script from all legacy text URLs.

ihongda:
Bhante, I have made changes based on feedback, please check when you have time :)

sujato (changes requested):
Still not working correctly. The segmented and legacy texts are inconsistent.


- turn references off
- enter http://localhost/mn9/en/sujato?reference=none
- expands URL to include all possible options — http://localhost/mn9/en/sujato?layout=plain&reference=none&notes=asterisk&highlight=false&script=latin
     - this is useful information, it tells me what I can do!
- now enter http://localhost/mn9/en/bodhi?reference=none
- does not expand URL — we should have http://localhost/mn9/en/bodhi?reference=none&highlight=false

As well as being inconsistent, this also breaks deep linking:

- :heavy_check_mark:  http://localhost/mn9/en/bodhi?reference=main&highlight=false#sc25
-  ❌ http://localhost/mn9/en/bodhi?reference=none#sc24

The latter case should not occur.

ihongda:
Bhante, I have modified the code according to your review feedback :)

sujato (approved):
works great!

buddhist-uni:
🙏 😊 Anumodana!

ihongda:
> 

:pray:

PR #2415, opened 2022-03-13
Fix: "search:search" just says search:search in breadcrumbs.

ihongda:
`"search:hint" just says search:hint`

This is caused by the missing `search:hint` key in the relevant localization file.

e.g.
https://github.com/suttacentral/sc-data/blob/master/sc_bilara_data/root/en/site/interface_root-en-site.json
https://github.com/suttacentral/sc-data/blob/master/sc_bilara_data/translation/de/site/interface_translation-de-site.json


sujato:
Looks good :heavy_check_mark:

ihongda:
The `search:hint1 information needs to be added to the relevant localization files to solve this problem :)

PR #2416, opened 2022-03-19
Serve range suttaplexes per-sutta.

ihongda:


sujato:
This is looking good, but there is a routing bug that affects addresses with `.` in them.

All these work fine.

http://localhost/iti45?view=normal
http://localhost/mn123?view=normal
http://localhost/dhp23?view=normal

But these return a blank page:

http://localhost/ud4.3?view=normal
http://localhost/snp2.2?view=normal
http://localhost/an5.1?view=normal

They work correctly on production, so it must be a new bug.

ihongda:
> http://localhost/ud4.3?view=normal
> http://localhost/snp2.2?view=normal
> http://localhost/an5.1?view=normal

> They work correctly on production, so it must be a new bug.

Thanks, Bhante, I noticed this problem, it may be related to nginx configuration, I will post to staging today to see if I have the same problem :)

PR #2428, opened 2022-04-01
Add site footer.

ihongda:


sujato (approved):
Looks good. I've styled the content, and added some CSS to try to make it stick at the bottom of the page. It works for me, but best test it in a few browsers.

I pushed directly to Hongda-work.

PR #2434, opened 2022-04-07
URLs must be case insensitive.

ihongda:
There was a related PR before, https://github.com/suttacentral/suttacentral/pull/2305 for the text page URL, 

the current PR is for the dictionary and search page URL.


sujato (approved):
Looks good locally.

PR #2436, opened 2022-04-11
Order translations by translator.

ihongda:
Need to execute `make load-data` before testing :)

Prioritize json file:
https://github.com/suttacentral/sc-data/blob/master/additional-info/prioritize.json


sujato (approved):
This works well.

PR #2437, opened 2022-04-11
Fix:The footer is not positioned correctly on the search page

ihongda:


sujato (approved):
I can't get search working locally, so just push it to staging and we'll see

Issue #2265, opened 2021-08-30
reduce or remove need for plurals


sujato:
Handling plurals in l10n is a hassle. Even with the right software support, it's still awkward and time consuming for translators. Blake is currently working on implementing proper plural support in Bilara; meanwhile I'd like to reduce the number of plurals to as low as possible: zero is pretty low!

Most of our plurals are in the suttaplex. We can work around these by changing the design a little:

![Screenshot from 2021-08-29 08-51-03](https://user-images.githubusercontent.com/6112010/131410765-9ecc9406-dd8e-4b07-9360-04eb13be9a87.png)

This is implemented in the following branch:

https://github.com/suttacentral/suttacentral/tree/sujato-no-plurals

THIS IS A PRELIMINARY PROOF OF CONCEPT AND IS VERY HACKY!

- [ ] ensure all l10n in all languages has plurals removed 
- [ ] ensure code is written properly and not just hacked!

This not only removes the need for plurals, it is to my mind a better and cleaner design.

Let's see if there are any other instances of plurals that we can get rid of.

PR #2440, opened 2022-04-18
Reduce or remove need for plurals.

ihongda:
Related issue: https://github.com/suttacentral/suttacentral/issues/2265


sujato:
From what I can see it looks good.

PR #2447, opened 2022-05-02
Add publication web UI.

ihongda:


ihongda:
There are still some issues and features outstanding, but the current version can be used as an initial preview :)

sujato:
A few notes.

## ATB

- [ ] On the [editions overview](http://localhost/publication-editions), the ATB has PREFACE | ACKNOWLEDGEMENTS and so on, this does not apply here. In fact there is no need for the ATB on this page. 
- [ ] For the Bio data, use the sc-data/additional-info/creator_bio.json
- [ ] On the [individual editions pages](http://localhost/publication-edition/dn) the ATB has "Abbreviations" and "index", these should be removed, they were only for mockup.
- [ ] For the "frontmatter", let's use a config file in sc-data/sc_bilara_data/_publication. This is a simplified version of the config files used for the standalone files.

Here's an example:

```
{
  "publication_number": "scpub2",
  "text_uid": "dn",
  "publication_type": "web",
  "frontmatter": [
    "./matter/foreword.html",
    "./matter/preface.html",
    "./matter/general_introduction.html",
    "./matter/introduction.html",
    "./matter/acknowledgements.html"
  ]
}
```

sujato:
I'm looking at the latest version, mostly it's great. Still the styles need to be done, but I will do that soon.

meanwhile, a few small adjustments.

- on the ATB, the *first* item should always be the edition home, eg. http://localhost/publication-edition/an
- the URLs should always have the edition included. Don't just have eg. `an` because there will be many versions of `an`. It has to be the same as for the suttas,including the ID, language, and author: `an/en/sujato`.
- Also this must be included in every URL, so we might have 
    - `/edition/an/en/sujato/main`
    - `/edition/an/en/sujato/preface`
    - `/edition/an/en/sujato/acknowledgements`
    - `/edition/an/en/sujato/introduction`
- should the first item have `/main`? or something like that? Or just `edition/an/en/sujato`? hmm. 
- The items in the breadcrumbs should be shorter, let me think exactly what to have there.

sujato:
Details must be sourced from `_publication-v2.json`. Currently we have incorrect results :

```
            <h1>${this.editionDetail[0].translated_name}</h1> 
            <p class="subtitle">A translation of the ${this.editionDetail[0].root_name}</p>
            <p class="author">${this.editionInfo.publication.creator_name}</p>
```

I have changed the class names to match the JSON keys.

sujato:
I'm looking at the latest changes locally. Apart from the fact that navigation to the individual editions works, none of the issues appear to have been fixed.

sujato:
ok, most issues are fixed now, some small things remain.

- [ ] On each edition page, there is a an "updated" date. This is still just dummy text (`Updated {{05:07:2022}}`). It should record the same updated date that is used in the editions themselves. Each edition has a date recorded in the `imprint` page, eg.: `This edition | 2022-07-20 09:58:52`. This should be the same date.

 - [ ] I'm getting mixups in the navigation: the "long Discourses" sometimes says "Middle Discourses", things like that. Looks like the nav is not updating itself properly. 

Screenshot from 2022-07-29 17-36-18

- [ ] The nav has `general_introduction`, it seems the titles are fetched from the config. Hmm. We can't fetch titles from the essays, they are too long. For now, maybe just replace the underscore with a space?

- [ ] Main title is just the title of the book: "Linked Discourses", not "Linked Discourses Collection".

- [ ] in each edition, the content of the breadcrumbs and the contextual toolbar should not change. They just tell you *that* you are in this edition. The ATB tells you *where* in the edition you are.

Screenshot from 2022-07-29 17-48-20



## remaining 2-do, depending on STXnext

- [ ] image for all editions has cover for dn. we'll have to update this when the images are available.
- [ ] proper links for resources

ihongda:
> On each edition page, there is a an "updated" date. This is still just dummy text (Updated {{05:07:2022}}). It should record the same updated date that is used in the editions themselves. Each edition has a date recorded in the imprint page, eg.: This edition | 2022-07-20 09:58:52. This should be the same date.

Bhante,@sujato This time doesn't seem to exist in the config file?

sujato:
> > On each edition page, there is a an "updated" date. This is still just dummy text (Updated {{05:07:2022}}). It should record the same updated date that is used in the editions themselves. Each edition has a date recorded in the imprint page, eg.: This edition | 2022-07-20 09:58:52. This should be the same date.
> 
> Bhante,@sujato This time doesn't seem to exist in the config file?

No, it cannot, because it is dynamic.

If you're not sure where it is taken from, ask STXnext. They use it to make the HTML files.

```
<td property="dc:date">2022-07-20 10:00:59</td>
```

sujato:
Great, everything looking good, just those three outstanding issues.

One more small detail I noticed.

Each edition has a `translation_ subtitle`. Looking at http://localhost/editions for DN it is :

```
<span class="translation_subtitle">A faithful translation of the Dīgha Nikāya</span>
```
This is correct! But then if I click through to http://localhost/edition/dn/en/sujato, I have the following:

```
<p class="translation_subtitle">A translation of the <!--?lit$315014405$-->Dīghanikāya</p>
```
This is incorrect. Don't try to generate the subtitle, always get it from the config.

***

Less, urgently, we should probably ensure that in each of these cases we use a consistent markup. 

Up until now, HTML hasn't really had a proper way of defining such material. Now, however, the `<hgroup>` should be used in such cases:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup

Most information on hgroup is out of date. It should not group multiple `hX` tags, but instead group  a single `hX` tag with one or more `<p>` tags. At some point we should adopt this for our suttas as well.

Anyway, I'll make this change on a branch.

sujato:
The checked items all look good, no further issues seen.

sujato:
One small bug, the inner HTML in blurbs is escaped improperly:

Screenshot from 2022-08-12 15-52-59



- [ ] inner HTML should render as HTML

sujato:
Still seeing the same bug after updating and refreshing data.

Screenshot from 2022-08-16 19-06-21

ihongda:
深度截图_选择区域_20220816222501

It's a bit strange, both browsers on my side can display normally, I will check again :)

sujato:
Hmm. Well, if you can't locate the bug, maybe we can just leave it until we're ready to push to staging and see how.

The work on the STXnext side is proceeding well. We have a good conversion to LaTeX/PDF for one text, and are now extending that to cover all of them.

ihongda:
This bug has been fixed :smile:

sujato:
Ok yes, now it is good!

client/elements/menus/sc-menu-static-pages-nav.js
```diff
@@ -60,6 +65,12 @@ class SCMenuStaticPagesNav extends LitLocalized(LitElement) {
     }
     if (this.changedRoute !== state.currentRoute) {
       this.changedRoute = state.currentRoute;
+      this._fetchMatter();
+      this.requestUpdate();
```
agilgur5, line 69: this erroneously fetches Edition files on _every page load_ (and unnecessarily calls `requestUpdate()` after a reactive property was just changed).
the code directly below this to fetch during `editionHomeInfo` is correct, fetching only when on an Editions page and when the currently loaded Edition changes.

I've removed the erroneous extra fetch in #3724

client/elements/menus/sc-menu-static-pages-nav.js
```diff
@@ -257,6 +270,47 @@ class SCMenuStaticPagesNav extends LitLocalized(LitElement) {
...
+      console.log(error);
+    }
+  }
+
+  get shouldShowPublicationToolbarTemplate() {
+    return html`
+      ${this.staticPagesToolbarDisplayState?.displayPublicationToolbar && this.editionMatters
+        ? html`
+            <li>
+              <a href=${this.editionHomeInfo?.url}>${this.editionHomeInfo?.title}</a>
```
agilgur5, line 298: Here's the usage of `editionHomeInfo` properties, where `root_title` is never used. I've removed it in #3720

client/redux-store.js
```diff
@@ -76,13 +76,20 @@ const initialState = {
...
   toolbarPosition: {
     scrollForToolbar: true,
     fixedToolbar: false,
     toolbarAtTop: false,
   },
+  currentEditionId: '',
+  currentEditionHomeInfo: {
+    title: '',
+    url: '',
+    root_title: '',
```
agilgur5, line 91: `editionHomeInfo.root_title` appears to never be used, per [above comment](https://github.com/suttacentral/suttacentral/pull/2447#discussion_r3832589812). I've removed it in #3720

client/elements/publication/sc-publication-edition-matter.js
```diff
@@ -0,0 +1,91 @@
...
+
+  stateChanged(state) {
+    super.stateChanged(state);
+    if (this.changedRoute !== state.currentRoute) {
+      this.changedRoute = state.currentRoute;
+      this.matter = store.getState().currentRoute.params.matter;
+      if (!this.matter) {
+        return;
+      }
+      this._fetchMatter();
```
agilgur5, line 49: So this re-fetches the Edition front matter files every time the file changes, even though all files are retrieved at once. I've fixed and refactored the logic in #3725

client/elements/publication/sc-publication-edition-matter.js
```diff
@@ -0,0 +1,91 @@
...
+      this.editionFiles = await (
+        await fetch(`${API_ROOT}/publication/edition/${store.getState().currentEditionId}/files`)
+      ).json();
+      // eslint-disable-next-line no-restricted-syntax
+      for (const key in this.editionFiles) {
+        if (this.editionFiles.hasOwnProperty(key) && key.includes(this.matter.toLowerCase())) {
+          this.matterContent = this.editionFiles[key];
+        }
+      }
+      this.requestUpdate();
```
agilgur5, line 77: There's no need to `requestUpdate` here since `this.matterContent` (etc) is a reactive property already.
I've refactored this logic in #3725

PR #2462, opened 2022-05-29
Fix: back button broken.

ihongda:


sujato (changes requested):
The problem in the original issue is fixed.

However, in the Vinaya, there is a bug with the patimokkha texts.

[Monastic Law of the Dharmaguptaka School](http://localhost/pitaka/vinaya/lzh-dg-vi) >> [Monks’ Rules (1st)](http://localhost/pitaka/vinaya/lzh-dg-vi/lzh-dg-bu-pm) 

- [x] Back button from patimokkhas does not work.

This bug applies to all patimokkhas.

Note that there is another bugs on these texts!

- [ ] There is a translation, but the Yellow Brick Road says "0 English".

Screenshot from 2022-06-08 08-06-05

ihongda:
>There is a translation, but the Yellow Brick Road says "0 English".

@sujato Bhante,

This question needs to check the menu API, because the `yellow_brick_road_count` value returned by it is indeed `0`,

I'll merge the code first so I can push to staging, and check for this later.

PR #2473, opened 2022-07-01
Generate `acronym` from `super_extra_info`.

ihongda:
Generate `acronym` from `super_extra_info` and added to ArangoDB `text_extra_info` collection.

Execute `make load-data` to take effect.


server/server/common/queries.py
```diff
@@ -1385,3 +1393,11 @@ class PWA:
...
     }
 '''
+
+ACRONYM_IS_NULL_UIDS = '''
+FOR doc IN super_nav_details
+    FILTER doc.type == 'leaf' AND doc.acronym == null
+    RETURN {
+        uid: doc.uid
+    }
+'''
```
MareinK, line 1403: Missing EOF newline.

ihongda: Thx @MareinK :smile:

sujato (approved):
This is looking good. Let's merge it. Then we can test if it affects the publication API. Once we are happy with the changes, we can notify Karl and others and, if possible, delete the `acronym` field, and the `volpage` field for PTS texts. But this is not urgent.

ihongda:
> This is looking good. Let's merge it. Then we can test if it affects the publication API. Once we are happy with the changes, we can notify Karl and others and, if possible, delete the `acronym` field, and the `volpage` field for PTS texts. But this is not urgent.

The acronym field of text_extra_info in arangoDB is used to store the newly generated acronym information, so it is still reserved, so that the new acronym information can be applied without changing many places.

The volpage field is updated according to bilara references, so it is currently reserved.

PR #2490, opened 2022-07-13
Change icons SVG tag to HTML tag

### LINKED ISSUE #2487: `svg` tag on `<svg>` element is wrong
In [`sc-icon.js`](https://github.com/suttacentral/suttacentral/blob/master/client/img/sc-icon.js),  Lit's `svg` tag is used on string literals containing `<svg>` tags. According to [the documentation](https://lit.dev/docs/api/templates/#svg), that is a "common error" and it should instead be tagged with `html`.

MareinK:
Closes #2487


sujato:
tried it locally, it works just fine

PR #2502, opened 2022-08-01
use hgroup, tweak styles

sujato:
Up until now, HTML hasn't really had a proper way of defining heading material. Now, however, the `<hgroup>` should be used in such cases:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/hgroup

Most information elsewhere on the web about hgroup is out of date. It should not group multiple `hX` tags, but instead group  a single `hX` tag with one or more `<p>` tags. At some point we should adopt this for our suttas as well.

This commit makes this change in the editions project.

### DIFF
--- a/client/elements/publication/sc-publication-edition.js
+++ b/client/elements/publication/sc-publication-edition.js
@@ -194,10 +194,9 @@ class SCPublicationEdition extends LitLocalized(LitElement) {
       return html``;
     }
     return html`
-      <header></header>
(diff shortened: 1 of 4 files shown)


ihongda:
Bhante @sujato , Since the editions page has been changed to get data from the API, this PR conflicts with the existing code, so I have pushed the changes from another PR.

So this PR will be closed.

PR #2520, opened 2022-08-31
Generate Abbreviations page from data.

ihongda:


sujato (approved):
Looking good. I made a couple of small changes:

- I removed the styles from abbrTable: `nowrap` causes the table to be overwide; and `width` is not necessary, let the table be as wide as it wants. It's good if the different tables are different widths, it helps distinguish them.
- I also changed the `<caption>` wording a little bit.

- Not an error, but shouldn't we be using `return css` rather than `return html      <style>`?

ihongda:
Thx Bhante @sujato .

PR #2524, opened 2022-09-06
Format acronym by rules.

ihongda:
> >According to the latest rules, pli-tv-bi-vb in super_extra_info.json is acronym as Bi, so the new acronym for pli-tv-bi-vb-pd2-8 after conversion is equal to Bi –pd2–8
> 
> No, the new version should be Bi Pd 2–8. I'll update the original issue.

Need to execute:
`make delete-database`
`make load-data`


sujato (approved):
okay, looks good

PR #2526, opened 2022-09-08
Use static styles for CSS where possible.

ihongda:
1. Use static styles for CSS where possible.
2. Adjust the code according to the latest lit specification.


sujato:
I've checked this locally, and so far as I can tell, it's all good.

PR #2530, opened 2022-09-30
Add separate fallen-leaves data and apply to frontend.

ihongda:


ihongda:
> The fallen-leaves handles:
> 
> 

- [ ] suttaplex, including name, parallels, blurb

- [ ] > But no text or navigation. There is no overlap between the two trees: an item must be either one or the other. When building the suttaplex-list, the two lists are combined.

> 
> Thus the fallen-leaves appear:
> 
> in the suttaplex list
> 

- [ ] as an individual suttaplex on its own URL.

>

sujato (commented):
So the navigation is all good, prev/next is good, breadcrumbs are good.

But the suttaplex still needs sorting out. 

Basically the situation is similar to the patimokkha, and we should use a similar approach. The patimokkha as a whole has text but no parallels, while the individual rules have parallels but no text.

http://localhost/pli-tv-bu-pm?view=dense

In fact the old approach on SC is also a bit similar, but not exactly right:

https://suttacentral.net/pli-tv-bi-vb-pd?view=dense

So going forward, the suttaplexes here stay mostly the same:

http://localhost/pli-tv-bi-vb-pd?view=dense

Except the parallels are governed by `fallen-leaves`.

There are no parallels for bi-pd 2-8, as the parallels are listed for each rule individually.

So:

- the suttaplex for bi-pd2-8 should reference texts not parallels (remove "parallels in ancient texts")
- add suttaplexes for individual rules, bi-pd2, bi-pd3, etc. Each one of these has parallels but no text.

Use the same green button as for the patimokkha to show/hide the parallels.

***

Also note, on make load-data we are getting the following error:

```
46: Generating and loading ordering information
ERROR:root:Text uids do not match menu uids for (vi): pli-tv-bi-vb-pd2, pli-tv-bi-vb-pd3, pli-tv-bi-vb-pd4, pli-tv-bi-vb-pd5, pli-tv-bi-vb-pd6, pli-tv-bi-vb-pd7, pli-tv-bi-vb-pd8
```

ihongda:
>the suttaplex for bi-pd2-8 should reference texts not parallels (remove "parallels in ancient texts")

Bhante @sujato , whether there is no parallels in all range sutta, no only bi-pd2-8?

ihongda:
>add suttaplexes for individual rules, bi-pd2, bi-pd3, etc. Each one of these has parallels but no text.

截图 2022-10-26 21-18-45

sujato:
> > the suttaplex for bi-pd2-8 should reference texts not parallels (remove "parallels in ancient texts")
> 
> Bhante @sujato , Besides `bi-pd2-8`, are there any other similar cases?

All the cases in `fallen-leaves` should be handled the same way.

sujato:
> > add suttaplexes for individual rules, bi-pd2, bi-pd3, etc. Each one of these has parallels but no text.
> 
> 截图 2022-10-26 21-18-45

Looks good, except 

- they need to be ordered correctly.
- make sure the "aparallels" are removed from the range, as that does not have parallels.

sujato:
Ok, getting there. I just noticed that in your screenshot, the bare "parallels" suttaplexes are using the old long form of the acronym: Pli Tv Bi Vb Pd 7, whereas the "text" suttaplexes are using the correct new form, Bi Pd 1.

Apart from that it is looking good.

One tiny detail, I noticed we have `<sc-suttaplex class="">`. Is that `class` necessary?

ihongda:
截图 2022-10-31 14-14-12

1. speed up the loading of the fallen leaves suttaplex list.
2. remove `class` when unnecessary.

sujato:
excellent, looking good, let's go ahead and roll it out!

PR #2535, opened 2022-10-20
Upgrade ArangoDB to latest version.

ihongda:
In the test, it was found that the result set of the new version of ArangoDB after executing the query is different from the previous version, so it is not merged for the time being.


sujato:
Okay, but it's important to keep up to date, so we should find out the reason for the discrepancy.

ihongda:
Downgrade ArangoDB version to `3.8.0`, the search result looks normal, It will not be merged for the time being, and we will observe whether the arangoDB official has relevant instructions in the future.

PR #2561, opened 2022-11-25
style up the maps

sujato:
So this makes a few tweaks to the styles on the map static page.

- make columns responsive
- make map and columns as wide as the screen, no need to keep them in a text-sized container
- use pill design for feature items
- use sans-serif font for pills
- change name to "the Buddha's India"

I also in a separate commit corrected the names of the features. there were a lot of unnecessary spelling variations, and in some cases the variants were all wrong. I've eliminated all variants and ensured that those that remain are actually found in SC. I only checked the entries with variants, though, so there will probably be mistakes in the remaining items.

### DIFF
--- a/client/elements/static/sc-static-map.js
+++ b/client/elements/static/sc-static-map.js
@@ -33,37 +33,46 @@ export class SCStaticMap extends SCStaticPage {
     unsafeCSS(typographyStaticStyles),
     css`
 
-  article{
-    max-width: 100%
-  }
-      .columns {
-        columns: 3 480px;
-      }
+  article
+{
+    max-width: 100%;
+}
 
-      .features-section {
-        break-inside: avoid;
-      }
+.columns
+{
+    columns: 3 480px;
+}
 
-      ul{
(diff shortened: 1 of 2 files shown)


MareinK:
Hi bhante, I didn't have time to look earlier. Looking good!

I notice one potential issue: the `h3` titles at the top no longer align:

image

This is due to the omission of `display: inline-block;` compared to the earlier code. Is this intentional? If not, then the following might be a fix: (simply reintroducing the `display` property causes other issues, but with `float` and `clear` it seems good)

```
      .features-section {
        display: inline-block;
        float: left;
        clear: both;
        break-inside: avoid;
      }
```

sujato:
I'll revisit the code, thanks for pointing it out. But the 90s can keep their clearfix. :wink: 

Generally speaking, columns don't try to align, they want to flow. But I'll see if there's an elegant way to address it. As I get older and crankier, I tend to prefer doing something simple and let the machines have their way, rather than aligning every pixel.

PR #2581, opened 2022-12-19
Add publication info to eligible nav cards

ihongda:


sujato (changes requested):
Looking good, a few comments:

- [x]  they should only be visible on the full display, not on the dense view (as toggled by the "spacing" icon).
- I have made a few changes to the styling and content.
- [x]  the text content for `creator` and `translation_subtitle` should be data-driven not hard-coded.
- [ ]  when I run it locally there is a bug when I select different languages, it doesn't work properly. Anyway, the edition should only be visible in one's chosen language: don't show Chinese speakers an English edition.

sujato (approved):
excellent, all looking good

Issue #2179, opened 2021-06-04
InstantSearch with ArangoDB?


sujato:
One of the most compelling features of meilisearch is the typo-tolerant instant search.

This is a really cool feature for us, especially as it helps with people who mistype sutta titles or Pali words. Sri Lankans, for example, normally search using their own local pali spelling conventions (eg. sujatho rather than sujato). 

Can we implement instantsearch using ArangoDB?

Perhaps using just sutta titles? Or titles and blurbs, perhaps?

Meilisearch uses the open-source library from Algolia for this.

https://github.com/algolia/instantsearch.js

https://github.com/meilisearch/instant-meilisearch

sujato:
Initial implementation is on staging, here are my notes.

First thing is, it works pretty well and definitely improves search. There are some bugs.

- **partial match fail**: some words do not trigger the instant search for a partial match. If I type `brahmajalasutta` or `samannaphalasutta` I only get results when I have typed the whole word.
- **duplicate results**: having typed the whole word `samannaphalasutta`, I get three matches.
- **wrong languages**: results are in other languages, we should only get root and the user's UI language.
- **inconsistent translated title results**: mostly there are either  translated titles or root titles, which is good. But sometimes I see a translated title (in the wrong language) next to the root title.
- **no keyboard selection**: I can only go to the result with a mouse. I want to be able to use the arrow keys (<kbd>↑</kbd>, <kbd>↓</kbd>) to select a result, and <kbd>Enter</kbd> to open it.

I think it would be better if the link opened on the sutta text rather than the suttaplex. Otherwise, we make the search superfast for the user, then slow them down by making them do another click. Can we make the search result link to the translated sutta text, choosing the same text that is visible in the suttaplex `dense` view? 

Now, as to performance. It's pretty good! I'm getting results after maybe .5s. But there's a definite lag, and that will increase if there are any network problems.

The original proposal was for true client-side instantsearch, where the data is downloaded to the browser first and searched with a js library like `flexsearch` (6.5kb) or `instantsearch` (26.2kb). Can we try implementing this? I'd like to try `flexsearch`, it seems awesome, super-small, super-fast. Here it is in Chinese!

https://juejin.cn/post/7012490016853065759

***

Also, here's an example of true client-side instantsearch. I guess it should be considered a gold-star example as it is Algolia's own site! It gives very rich results instantly as you type.

https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/js/

[...]

Issue #1581, opened 2020-02-16
Alias popular texts


sujato:
We have a problem where certain well known texts are often searched for under specific names, but those are not found on SC.

The most common case is the "Metta Sutta", which is not called that in Pali, but is very commonly searched. 

I propose we make a list of such cases `alias.json`, something like:

```
{
  "uid": "snp1.8",
  "alias": [
    "metta sutta",
    "karaniyametta sutta",
    "mettasutta",
    "discourse on loving-kindness"
  ]
}
```

This could be applied in such cases as 

- metadata for SEO
- return correct text on SC's search

We could source possible aliases by asking people on discourse and by reviewing search logs.

ihongda:
@sujato  Bhante, I have implemented this feature and pushed it to staging.

![Image](https://github.com/user-attachments/assets/982f4d00-efd2-428b-b1df-5ac570350f86)

thesunshade:
Thank you!

Can you document in the wiki how other things would be added?

ihongda:
> Thank you!
> 
> Can you document in the wiki how other things would be added?

Bhante, no problem.

ihongda:
@thesunshade Bhante, I have added a new wiki about how to add alias data.

https://github.com/suttacentral/suttacentral/wiki/How-to-Add-Alias-Data-to-alias.json

thesunshade:
Thank you!

Is it necessary to have the aliases in all lower case without diacritics?

ihongda:
Capitalization is fine in principle, but currently no processing is done for diacritical marks.

PR #2596, opened 2023-01-13
Apply new search.

### LINKED ISSUE #2605: Add "filter" icon and top sheet for search results.
Currently the search results have a "visibility" icon, with label "filter". This opens a top sheet with language selections.

Change this to:

- use "language" icon (globe), the same as language selection in main menu
- label should be "languages"

Okay, now add a second icon. 

- Use the "filter" icon. This must be added to our icons file, see the code below.
- label should be "filters"

In this page we should have a list of all available filters, together with short examples.  Currently this would only be plain text. Something like the following, but this is not complete. We should make sure that *all* filters, keywords, and special functions are listed.

>**volume/page search**
>
>Find texts using legacy PTS references.
>
>- match page: `volpage:SN ii 4` , `volpage:M II 246`
>- match entire volume: `volpage:SN ii`, `volpage:M II`
>
>**author search**
>
>Filter results by author.
>
>- `author:brahmali cat`
>- `author:sabbamitta Katze`
>
>**collection search**
>
>Narow search to within a specific collection, as defined by SC IDs.
>
>- `in:dn cat`
>- `in:snp dog`
>
>**operators**
>
>- return one OR the other: `greed OR desire`
>- return one AND the other: `greed AND desire`
>
>**combine filters**
>
>- `in:mn author:sujato cat OR dog`
>
>**chinese**
>
>- space-separated keywords: 八正道 涅槃， 四神足 三摩地
>- simplified/traditional Chinese conversion: 发勤 ，观心生灭

***

```
<svg xmlns="http://www.w3.org/2000/svg" height="48" viewBox="0 96 960 960" width="48"><path d="M440 896q-17 0-28.5-11.5T400 856V616L161 311q-14-17-4-36t31-19h584q21 0 31 19t-4 36L560 616v240q0 17-11.5 28.5T520 896h-80Zm40-276 240-304H240l240 304Zm0 0Z"/></svg>
```

ihongda:
Need to import data first.

execute:

`docker-compose build sc-flask`
`make run-dev-no-logs`
`make delete-database`
`make load-data`
`make index-arangosearch `


ihongda:
Added:

1. Add suttaplex to search results.
2. Add dictionaries info to search results.

Plan:

1. Support `volpage` to search for conditions.

ihongda:
1  .Support volpage keyword search, support two modes
-  a. Search keywords have specific page numbers, such as `M II 174`, then it will display 10 results before and after page 174.
-  b. Search keywords have no specific page numbers, only a range, for example: `M II`, then it will return all results within the range.
---
-  c. Results are displayed in a `table` format.
-  d. Highlighting `volpage` in the table.

2. Chinese search supports searching with `multiple keywords` separated by spaces.

3. If input is simplified Chinese, it will be converted to `traditional Chinese` in the background during the search.

4. Support search with `or` keyword separating words.

5. Display the full name of the language in the language filter dropdown list.

sujato:
Hongda, this is all looking great. May I ask that in future, for each feature, give an example or two of how to test each new feature. I'm testing the features now, and I'll give examples here, so we are sure we are testing the same thing.

## Add suttaplex to search results.

http://localhost/search?query=mn23 

I can see the suttaplex!

## Add dictionaries info to search results.

http://localhost/search?query=atthika

Yep, the dictionary result is there. It even includes the map!

http://localhost/search?query=savatthi

##  Add languages filter dropdown list.

Hmm, not sure what is happening here.

http://localhost/search?query=rajagaha

Under "all" I get the dictionary results, then just a list of titles and metadata, no content. If I select `en` the dictionary goes away and most entries are the same, just useless. They don't even show "rajagaha". 

Searching "cat" in `en` I get a lot of results for "purifi**cat**ion" and the like. We should sort by making the whole word appear first. Sort order:

1. exact matches = `cat`
2. variants of exact matches (eg. plurals such as `cats`)
3. unrelated word that include the string `cat`

Also, we should refine the UI for this.

## Support using volpage as search criteria.

I'm not seeing any results for this.

## Support using 'author: author name' as search criteria.

also not seeing any results

## Display search results by author in table format.

also no results

ihongda:
@sujato Bhante, today I am optimizing and refactoring the code, and I will submit it today, including the above updates, and I will write some corresponding examples.

sujato:
Hongda, maybe use a checkbox next to each item, then I can just check them off.

- [X] sort whole words first
- [X] volpage search is working well
- [X] Chinese search supports searching with multiple keywords separated by spaces.
- [ ] greed or desire: this gives confusing results. I get results for `greed`, `desire` and `or`. And `or` includes things like `fact**or**`. Something should be either an operator or a result. [Google uses capitals](https://www.googleguide.com/or_operator.html) `greed OR desire`, can we do that?
- [ ] search keyword is Simplified Chinese, it will be converted to Traditional Chinese: ok wow, I didn't even know that was possible. I'm wondering, though, do we have any actual simplified Chinese? In modern translations I mean? Can a Chinese user search using simplified only?
- [ ] `author:name` works well. Can we do text search filtered by author name? `author:sujato desire OR greed`, "search in sujato translations for the words 'greed' or 'desire'".

>Is it necessary to add more complex professional modes, such as author:sujato and collection:an

Yes, I think it would be good. 

```
author:sujato AND collection:an greed OR desire
```

- search in sujato translations of an for the words 'greed' or 'desire'

I do this kind of thing a lot, and I have to use Sublime Text locally for it.

ihongda:
>  search keyword is Simplified Chinese, it will be converted to Traditional Chinese: ok wow, I didn't even know that was possible. I'm wondering, though, do we have any actual simplified Chinese? In modern translations I mean? Can a Chinese user search using simplified only?


As far as I know, the translation of author_uid = mahinda2 is in Simplified Chinese, and I am currently making further modifications, that is, regardless of whether the user enters Simplified Chinese or Traditional Chinese, the results including Simplified Chinese and Traditional Chinese will be returned.

ihongda:
2023.2.1 update:
1. Regardless of whether the user enters Simplified Chinese or Traditional Chinese, search results in both Chinese are returned.
2. Modify the logic of or search and highlight.
3. Support for complex search patterns, 
	e.g.
	a: `author:sujato and collection:an greed or desire`
	b: `author:sujato and collection:dn greed or desire or love`

sujato:
Great, looking good.

A couple of questions!

- It seems the operators `and` and `or` still do not require capitalization. However, I am not getting any unwanted results for `or` which was the case earlier. So is the plan to keep using the unpatilazed versions? I'm fine with that so long as it works (which it does!)
- Now it returns often longer passages of text, it seems to include all the entries for a given word in the text. Is that right? It seems good!
- I'm still getting some cases where the results correctly return a text, but do not show the actual text. Eg. I search `desire or love` and I get a result for AN 10.47. Now, that sutta does in fact contain the word "love". But in the search results there's no text. (see screenshot).

Screenshot from 2023-02-03 18-49-46

ihongda:
> Now it returns often longer passages of text, it seems to include all the entries for a given word in the text. Is that right? It seems good!

Currently only returns up to three related fragments, otherwise it may take up too much screen real estate.

ihongda:
> It seems the operators and and or still do not require capitalization. However, I am not getting any unwanted results for or which was the case earlier. So is the plan to keep using the uncapitalized versions? I'm fine with that so long as it works (which it does!)

Capitalization is currently not required, but I decided to be consistent with Google and use capitalization :smile:

ihongda:
> I'm still getting some cases where the results correctly return a text, but do not show the actual text. Eg. I search desire or love and I get a result for AN 10.47. Now, that sutta does in fact contain the word "love". But in the search results there's no text. (see screenshot).

This problem has been fixed by ignoring case when locating keywords.

sujato:
- [x] whole word results? 

Currently the results don't care if they return whole words or not. So if I search for "whole" for example:

Screenshot from 2023-02-07 06-50-01

It would be nice if the results returned only whole words.

## search title to return suttaplex

Currently if I search sn3.2 I get the suttaplex. It would be good to be able to return the suttaplex by searching its title. This would include:

- the translated title in either `name` or the translation text or both
- the root title

Sometimes there are several suttas with the same title. Never mind, just return them all one after the other.

## use `in:` instead of `collection:`

- [x] Can we use the filter `in:an` rather than `collection:an`? It's much shorter and handier!

##  search options UI

Here I will attempt to describe a more refined way of handling languages and other search options

- [x] by default, search in the site language + root languages

Assume my site language is English. A search for `cat` would return

- whole word results, which would be translated texts with the word `cat`, then
- both translated and root texts that include the string `cat`, such as `**cat**egory` in English or `avo**cat**` in Sanskrit.
- dictionary results (there are none in this case).

It would *not* return:

- results in any other language
- results from any static pages, etc.

Ok fine. Now, in order to refine language results, we currently use a `Languages` dropdown. This is populated with the languages in which the search term actually appears. A dropdown is not an appropriate UI for this, as it is used to select a single item, whereas we want to be able to select multiple languages.

Instead, let's reuse the top sheet element from the text pages. That way we can keep the UI consistent in different views.

### context-toolbar action-items

We would have (I think) two items. Each have an icon and a label.

- languages
- filters

Click on one of those and a top-sheet appears.

### top-sheet: languages

In the languages UI we would have a list of root and translation languages, arranged in a grid. Each item has a checkbox, so we can select as many or as few items as we like. Here is a VERY ROUGH mockup.

Screenshot from 2023-02-07 15-17-32

[language-select.zip](https://github.com/suttacentral/suttacentral/files/10671504/language-select.zip)

- Question: should we use native checkboxes or `md-` web components? Personally I think the native versions work just fine, why not use them.

### top-sheet: filters

A second top-sheet can contain the details for the various filters. Currently I think this would only be plain text. Something like the following.

>**volume/page search**
>
>Find texts using legacy PTS references.
>
>- match page: `volpage:SN ii 4` , `volpage:M II 246`
>- match entire volume: `volpage:SN ii`, `volpage:M II`
>
>**author search**
>
>Filter results by author.
>
>- `author:brahmali cat`
>- `author:sabbamitta Katze`
>
>**collection search**
>
>Narow search to within a specific collection, as defined by SC IDs.
>
>- `in:dn cat`
>- `in:snp dog`
>
>**operators**
>
>- return one OR the other: `greed OR desire`
>- return one AND the other: `greed AND desire`
>
>**combine filters**
>
>- `in:mn author:sujato cat OR dog`
>
>**chinese**
>
>- space-separated keywords: 八正道 涅槃， 四神足 三摩地
>- simplified/traditional Chinese conversion: 发勤 ，观心生灭

ihongda:
2023.2.7 changelogs:
1. Support command `list authors` to display all authors.
2. Fix:
>I'm still getting some cases where the results correctly return a text, but do not show the actual text. Eg. I search desire or love and I get a result for AN 10.47. Now, that sutta does in fact contain the word "love". But in the search results there's no text. (see screenshot).
3. Use `OR` instead of `or`.
4. Handle case when searching with `volpage:`

ihongda:
> Currently if I search sn3.2 I get the suttaplex. It would be good to be able to return the suttaplex by searching its title. This would include:
> 
> the translated title in either name or the translation text or both
> the root title

If using fuzzy search, there may be many results.

ihongda:
2023.2.8 changelog
1. Use `in:` instead of `collection:`

sujato:
> > Currently if I search sn3.2 I get the suttaplex. It would be good to be able to return the suttaplex by searching its title. This would include:
> > the translated title in either name or the translation text or both
> > the root title
> 
> Bhante, If using fuzzy search, there may be many results.

What if it were exact matches with a list of aliases?

Let's take, say the Satipatthanasutta.

Valid matches from the name files + the text `<h1>` would be:

- The Discourse on Mindfulness Meditation
- Mahāsatipaṭṭhānasutta
- Satipaṭṭhānasutta
- Great Discourse on the Establishment of Mindfulness
- Discourse on the Applications of Mindfulness

So any one of these would return the suttaplex for MN 10.

In addition, we could make a hand-curated list of aliases. It wouldn't be very large, but we could add common variants or misspellings. In this case:

- Satipatthanasutta
- Mahasatipatthanasutta
- Sathipattana
- Maha Satipattana sutta

etc. 

See here for the issue for this: https://github.com/suttacentral/suttacentral/issues/1581

Now, this search would indeed return duplicate results for MN 10 and DN 22, but that's okay, they are duplicate suttas.

On a more advanced level, this relates to instant search:

https://github.com/suttacentral/suttacentral/issues/2179

If we are able to implement instant search based on, perhaps:

- name files
- titles 
- synonym files

Then when the user types an approximate spelling, we suggest near-matches, and ultimately they just get the one they want. This also has the advantage of training users to understand how to spell things correctly!

ihongda:
2023.2.10 changelog:
Suport more flexible search modes:
e.g.
1. author:sujato in:dn greed
2. author:sujato in:dn greed OR love
3. author:sujato greed
4. in:dn sky
5. in:an sky OR flying

ihongda:
2022.2.17 changelog
1. Support for filtering search results by language
 A user's selection of a language list immediately affects current search results and will affect new search results.

---
Another idea: when the user searches, it will be fed back to the language options according to the search results. For example, if the search results contain results in three languages, then the three languages will be ticked in the search options.

ihongda:
2023.2.23 changelog:

1. Support `title:` keyword
	for example: `title:the prime net`

ihongda:
2023.2.24 changelog:
1. In Search Options, make Root Language and Site Language selected by default.
2. Remove the filter dropdownlist.

ihongda:
2023.2.27 changelog:
1. Support `in:ebt` keyword.

ihongda:
2023.3.1 changelog:
1. Add "filter" icon and top sheet for search results

client/elements/addons/sc-top-sheet-search-options.js
```diff
@@ -0,0 +1,165 @@
...
+      input,
+      label {
+        cursor: pointer;
+      }
+    `,
+  ];
+
+  get actions() {
+    return {
+      changeDisplaySettingMenuState(display) {
```
agilgur5, line 55: the `actions` method here is unused, with `reduxActions` used instead. the functions are not identical, but similar, so I'm guessing this was re-thought or re-written before the PR was made.

I've removed this unused method in #3713

agilgur5 (commented):
There was an unused `actions` method in this PR. I've removed it in #3713

PR #2611, opened 2023-03-06
Add Lulu URLs and show multiple vols for publications

### LINKED ISSUE #2609: Add Lulu URLs and show multiple volumes for publications
I have now published the books to Lulu. We need to do two things on the editions pages.

- add the correct Lulu URL for each edition
- show multiple volumes where needed

## Lulu URLs

I have added a new field `volume_lulu_url`. This contains the absolute URL for Lulu for each volume.

For example in `dn-en-sujato_scpub2-ed5-paperback_2022-02-10.json`

```
"volume_lulu_url": "https://www.lulu.com/shop/bhikkhu-sujato/long-discourses-vol-1/paperback/product-4krvj7.html",
```

Which would become:

```
<tr>
        <td>${icon.paperback} Book, paperback</td>
        <td>
          <a href="https://www.lulu.com/shop/bhikkhu-sujato/long-discourses-vol-1/paperback/product-4krvj7.html" class="external">
            ${icon.external} Print on demand
          </a>
        </td>
      </tr>
```

(BTW, it looks like the HTML formatting for this template is not good, I have corrected it here, but not in the source.)

## Multiple Volumes

In some cases, namely in KN, each edition is a single book with a single volume. For example, Snp is all printed in one book as one volume.

In the larger nikayas, we split them into multiple volumes. So DN has three volumes, for example, each of which is a separate physical book. Unfortunately, Lulu does not have a concept of "volumes", so we cannot list all three together. We must supply an individual link for each Volume on Lulu.

To do this we must introduce some logic to the `sc-publication-edition.js` template. Currently we have:

```
#printOnDemandTemplate() {
    return html`
      <tr>
        <td>${icon.paperback} Book, paperback</td>
        <td>
          <a href="https://github.com/suttacentral/editions" class="external">
            ${icon.external} Print on demand
          </a>
        </td>
      </tr>
    `;
  }
```

The logic is:

- if a publication has only one volume, make no change: just keep doing what we currently do.
- else if a publication has more than one volume, retrieve each `volume_lulu_url` from the API 
- and make a separate link
- and identify each link by volume number

```
      <tr>
        <td>${icon.paperback} Book, paperback</td>
        <td>
          <a href="https://www.lulu.com/shop/bhikkhu-sujato/long-discourses-vol-1/paperback/product-4krvj7.html" class="external">
            ${icon.external} Volume 1: Print on demand
          </a>
          <a href="https://www.lulu.com/shop/bhikkhu-sujato/long-discourses-vol-2/paperback/product-ppydw9.html" class="external">
            ${icon.external}  Volume 2: Print on demand
          </a>
          <a href="https://www.lulu.com/shop/bhikkhu-sujato/long-discourses-vol3/paperback/product-26j28g.html" class="external">
            ${icon.external}  Volume 3: Print on demand
          </a>
        </td>
      </tr>

```

Here's what it looks like:

Screenshot from 2023-03-04 17-31-19

It also needs some small CSS changes, which I have added with this PR: https://github.com/suttacentral/suttacentral/pull/2608/files

- [ ] add `volume_lulu_url`
- [ ] support multiple volumes
- [X] adjust styles

ihongda:


sujato (commented):
The styling is right, but none of the URLs work locally. If this is expected, then go ahead.

PR #2618, opened 2023-03-09
Enhance search and fix bugs

### LINKED ISSUE #2612: top sheets should close when you do a search or change filters
If you look up a search filter or sort by language, currently the top sheet remains open until you close it. Instead, leave it open until either:

- you close it
- you do a search. 

Otherwise the sheet is covering the search results. Annoying!

Same problem if you switch filters: if you open one filter, then click the other filter, it should hide the first one.

### LINKED ISSUE #2613: display search results for other languages if there are no results in user languages
Normally you get results in your language, as defined in the top sheet. But what if your search explicitly indicates that you want results in another language?

Currently if I search for `author:sabbamitta Katze` it tells me there are 7 results, but nothing is displayed. But the fact that I have defined an author and searched in German says that I want results in German, regardless of what it says on my "language" settings.

The language filter is useful if results are present in multiple languages. So if I search for "buddha" I should only get results in my language. But if I search for a German word or author, I must be looking for German results.

So how to deal with this?

How about this logic:

- look for results in languages as defined in settings
- if there are no results there, look in all languages
- if there are results, display them

In addition, show the user a notice:

>We found no results in the languages in your search language settings. But we did find results in `{german}`, `{french}`, `{saarlandisch}`, so here they are. If you want to search in these languages, add them via the “languages” icon in the toolbar.

This can be added  simply as an `<aside>` directly after `.search-result-head`. In this case, probably replace the existing `<aside>` so there is only one hint at a time. Maybe we can add other hints as well.

### LINKED ISSUE #2614: search for `dog` gets fewer results than `dogs`
https://staging.suttacentral.net/search?query=in:an%20dog

https://staging.suttacentral.net/search?query=in:an%20dogs

In such cases, show the whole word match first, then show other results. When we add more characters it should only decrease the number of results.

### LINKED ISSUE #2615: results sometimes do not include the search term
Screenshot from 2023-03-07 10-03-20

Search results must always include the search term.

ihongda:


ihongda:
2023.03.09 changelog
1. When searching and changing filters, close related top sheets.
2. Fix: results sometimes do not include the search term.

ihongda:
2023.3.10 changelog

Fix: Highlighted snippets not showing when searching for Pali.

ihongda:
2023.3.13 changelog

fix: the problem with the wrong title when clicking back to the home page from the search page.
fix: display search results for other languages if there are no results in user languages.
fix: search for `dog` gets fewer results than `dogs`.
fix: Remove extra spaces after the colon, e.g. format `in: sn` as `in:sn`
fix: A bug with search for locations.

PR #2627, opened 2023-03-14
Enhance search and fix bugs.

### LINKED ISSUE #2616: ebt filter
Add to ebts:

I'm wondering if we can simplify the list by using the groupings found in the URLs? 

So for example we have `/linked`, which includes sn, sa, sa-2, sa-3, sa-ot. Can we just add `linked` to the ebt list?

If that's not possible to do directly, we should ensure that all the relevant collections under these URLs are added. So:

- dn, da, da-ot
- mn, ma, ma-ot
- sn, sa, sa-2, sa-3, sa-ot
- an, ea, ea-2, ea-ot
- "kp", "iti", "ud", "snp", "dhp","thig", "thag"
- (all the texts listed under `dharmapadas`)
- sf
- up, d

(in addition, all Vinaya texts currently included)

ihongda:


ihongda:
2023.3.14 changelog:

1. Adjust `ebt` content.

PR #2629, opened 2023-03-20
fix: fixes bugs with scroll-based toolbar animation improves anim

Codigo-Fuentes:


client/utils/rafThrottle.js
```diff
@@ -0,0 +1,15 @@
+const rafThrottle = callback => {
```
adampotency, line 1: requestAnimationFrame should create a smoother animation than throttle and be more performant.

There is also an npm package that is similar if that is preferable:

https://www.npmjs.com/package/raf-throttle

client/elements/sc-site-layout.js
```diff
@@ -147,25 +147,16 @@ export class SCSiteLayout extends LitLocalized(LitElement) {
       this.actions.changeDisplaySuttaInfoState(false);
     });
 
-    let scrollDistance = 0;
+    // handles scroll-based universal toolbar animation on home page
```
adampotency, line 150: These two listeners were fighting with each other re setting styles on the toolbar which was causing bugs.

As far as I can tell, this is only meant to handle the toolbar animation on the homepage, so keeping the scope clear and mutually exclusive should work as a solution.

I added a comment and an early return at the top of the function if not on the home page to make this super explicit and clear. Please let me know if that is correct.

client/elements/sc-site-layout.js
```diff
@@ -147,25 +147,16 @@ export class SCSiteLayout extends LitLocalized(LitElement) {
...
     document.addEventListener(
       'scroll',
-      throttle(500, () => {
-        if (!this.toolbarPosition.scrollForToolbar) {
+      rafThrottle(() => {
+        if (!this.toolbarPosition.scrollForToolbar || this.changedRoute.path !== '/') {
           return;
         }
-        const syntheticEvent = new WheelEvent('syntheticWheel', { deltaY: 4, deltaMode: 0 });
```
adampotency, line 157: I am not 100% certain what this is meant to do but it seems to be a secondary way of rate limiting? Please let me know if there was some other purpose.

client/elements/sc-site-layout.js
```diff
@@ -174,40 +165,50 @@ export class SCSiteLayout extends LitLocalized(LitElement) {
...
     let lastScrollTop = 0;
+    // handles scroll-based universal toolbar animation on pages other than the homepage
     document.addEventListener(
       'scroll',
-      throttle(500, () => {
-        if (!this.toolbarPosition.scrollForToolbar) {
+      rafThrottle(() => {
+        // if settings specify that scrolling is not the basis for toolbar visibility
+        // or if the current route is the home page return
+        if (!this.toolbarPosition.scrollForToolbar || this.changedRoute.path === '/') {
```
adampotency, line 174: move all the early return conditions to the top to make the control flow very clear.

sujato (commented):
Running it locally it seems good. I'm not quit sure how to merge it so I'll leave that to Hongda.

I did have one question: is there a recommended way to make adjustments to animation if required?

More specifically, is it possible to make the "scroll up" animation a little less sensitive? On a desktop it shows the toolbar after one notch on a scroll wheel. I think it should be possible to scroll up a little bit before showing the toolbar. (I know, this is the same as the example I gave you, but still, if it could be tweaked …)

client/elements/sc-site-layout.js
ihongda: Works great, this is something I've always wanted to improve, thanks :pray:

ihongda: :+1:

client/elements/sc-site-layout.js
ihongda: Thanks for the improvement ：）

ihongda (approved):
Thanks, now it's smoother than before and the code is much clearer :+1: :pray:

PR #2669, opened 2023-05-16
Enhance: show verse number in suttaplex nerdy row.

### LINKED ISSUE #2626: show verse number in suttaplex nerdy row
Some texts list verse numbers, which are commonly used to reference things. They are displayed in the Dhammapada, but not in other texts. 

>The Dhammapada is the exception because the verse number is also the sutta number. Since it already works well, we should make sure to exclude it from these changes so that it does not show duplicate numbers.

Screenshot from 2023-03-13 18-00-08

We should display verse numbers whenever they are available. These are listed as `vns` in the `reference` data. They should show up in the "nerdy-row" after other details.

- Show the first and last verse numbers within scope
- Make sure to use en-dash (–) as separator! 

Screenshot from 2023-03-13 18-05-58

ihongda:


sujato:
I'm running the latest version, have updated data, but I am not seeing the verse numbers in snp.

ihongda:
Ok, Bhante, Let me push it to` staging` to see if it works

PR #2670, opened 2023-05-16
Enhance `sc-badge.js`

### LINKED ISSUE #2655: add badges on suttaplex for "aligned" and "annotated"
On the suttaplex it is not immediately obvious which suttas are "aligned", (i.e. root and translation segments aligned via Bilara), except by the fact that it has translation language + root language mentioned.

Nor is there any indication of where notes AKA annotations are available.

Let's solve these by putting a badge on the suttaplex to indicate these things. 

Something like this.

badge-mockup

ihongda:


sujato:
There's a small bug where sometimes the "annotated" badge is displayed, but there are no annotations.

http://localhost/ud-bodhivagga?view=dense

ud1.1 has a badge, but no annotations.

We can see why by looking at the file:

https://github.com/suttacentral/bilara-data/blob/published/comment/en/sujato/sutta/kn/ud/vagga1/ud1.1_comment-en-sujato.json

There's just an empty JSON for the note. This was a problem on the front end as well, but I fixed it with CSS.

We should do both of these things:

- the "annotations" badge should be triggered only by the presence of actual annotations, not an empty file
- we should delete blank comment segments and files from Bilara-data in the relevant GA.

PR #2676, opened 2023-05-23
Sujato add map

sujato:
This adds the static map image to /map.

- [X] add image
- [X] modify html and css
- [x] localize text strings

@ihongda can you do the localization, and then merge? Thanks!


ihongda:
Ok, Bhante, let me check.

PR #2696, opened 2023-06-20
Upgrade material components.

ihongda:
execute:
`docker-compose build sc-frontend`
`make run-dev-no-logs`


sujato:
A couple of bugs. Some of these are bugs in the old implementation too!

- <s>When you click the search icon, there is a white space around the icons</s>
- [x]  when clicking to change language, the menu resizes: it should stay the same size, at least horizontally and probably vertically. 
- [x] There's no ripple effect on the more-menu-button. Odd, i can see the ripple component there, but it's not working.

Also, there's no mouse scrolling on the Languages menu. This applies currently too. It's tricky to scroll down the list of languages. On my machine, the only way is to click on the "Choose your language" header, then scroll with the keyboard. Not good UI!

I see that it is already set to `overflow-y: scroll` but for some reason it's not working.

- -[x] we should be able to just scroll with the mouse/trackpad/touchscreen (Compare the Youtube implementation, where they get it right.)
- [x] there should also be a scrollbar. Make the scrollbar *always* appear on the language chooser.
- -[x] Also make the "Choose your language" header fixed, so that the languages scroll under it. That way there is always a visible way to return to the normal menu view.

sujato:
I think this is merged already?

ihongda:
yes, i will close this PR.

PR #2734, opened 2023-08-03
Use `oramasearch` to implement instant search.

ihongda:
1. Use `flexsearch` to implement instant search.
2. Only return data relevant to site language.
3. Capture keyboard up and down key events。
4. Only return `ebt` related data.


sujato:
Okay, I have it running locally. The basic functionality works, but some improvements are needed.


- [x] Avoid duplicating responses

    - entering mn22 gives two results, one for the Pali title, one for the translation. There should be one response, with both Pali title and translation.  
- [x] Link to sutta text, not suttaplex
    - The results should take you straight to the segmented sutta text without further links.

- [x]  partial matches are lacking

   - it only gives whole word responses: thus mn1 only returns mn1 (not mn11, etc.). It should return anything starting with or containing mn1, then narrow it down as you type.

 -  fuzzy matching is lacking
       - similar to the problem of partial matches. If you omit diacriticals or make a mistake when typing, you get no matches.

The last two items are essential for instant search, the whole point is that people can get feedback as they type, whereas now you only get a result when you type the complete thing without mistakes.

Looking at flexsearch more closely, it seems that it does not support fuzzy search. The README mentions it, but an open thread says it is not available.

https://github.com/nextapps-de/flexsearch/issues/118

If we cannot do typo tolerant fuzzy searching with flexsearch, maybe we should look for an alternative? <s>Maybe look at instantsearch.js?</s> It seems that instantsearch.js can only be used with Algolia? I think?

https://www.algolia.com/doc/guides/building-search-ui/what-is-instantsearch/js/

Or minisearch:

https://lucaong.github.io/minisearch/

gustavopch:
Another alternative: https://github.com/oramasearch/orama

sujato:
Thanks! Do you have any experience with these? Any thoughts or recommendations?

gustavopch:
Unfortunately not. I just know that Orama (previously called Lyra) looks quite feature rich and well maintained.

sujato:
Oramasearch looks great actually, super small ([3.9kb](https://bundlephobia.com/package/@orama/orama@1.2.0)), performant, simple API, typo tolerance, stemming … @ihongda what do you think, should we try it out?

ihongda:
I have reimplemented instant search using oramasearch, oramasearch is fast and easy to use.
Thanks @gustavopch

sujato:
Ok, checking oramasearch now, it's looking great. We need to refine a few things.

- **updating this after further consideration!**

Let's emulate Github's search interface as far as seems useful. Here is a mockup.

- the user has written `satipa` in the search field
- the input field has an autocomplete based on the names of sutta titles. The autocomplete suggestion is highlighted with `<mark>`. (This behavior emulates browser address bars)
- the search term is highlighted with `<b>` in the results.
- user can press enter or click on the input field to search this term, taking them to the normal search results page.
- or else the user can select one of the other two search options, which apply existing search filters.
    - `in:ebt` is always present
    - Then there is an optional item added when a user is in a navigation context. Here they are at `/mn` (or `/pitaka/sutta/middle/mn`) so `in:mn` is added.
- user can also click one of the "jumpto" options which takes them directly to the text or the collection.
- In addition to the names themselves, the jumpto items are visually distinguished between the text and collections, in two ways:
     - the "books" icon at the left: for texts, this is the same icon as used in the suttaplex results for suttas; for collections this will be something that looks like a collection; maybe a pile of books
     - the text at the right has `jump to text` and `jump to collection`
- As always, we should use the acronym for display rather than the UID. However, search should work for both, so we may need to add the acronyms to the search data. So if someone searches `mn2` or `mn 2` they get the same results. 
- Each element of the search results items should be wrapped in a span with appropriate class (eg. `<span class='root-title'>`) for styling, etc.
- at the bottom of the page is a "search tips" link. This leads to a new static page (which does not yet exist). This lists the search options available via the filters on the search results bar, and any other information useful for searching. (I'll create this page soon.)

Screenshot from 2023-08-17 18-57-51

*Notes:*

- currently we have "class=selected" on search items. We should try to avoid adding/removing classes as much as possible, prefer using :focus, :hover, etc. pseudo-elements if possible. That way there is no chance to get a mismatch.
- Obviously the emojis should be replaced with proper icons!

## what of the current search page?

In this new design, the current search input in the top bar is completely removed. When you click the search icon, it opens the oramasearch dialog.

Let's keep this behavior consistent. 

Look at a classical search results page, eg. http://localhost/search?query=satipatthana

The page stays mostly the same, except:

- there is a new search input at the top of the page , not in the toolbar, which contains the search term.
- underneath is the results heading (we can restyle this)
- the search hint can be removed, it's no longer necessary
- this search input handles full text search only

This is mockup shows the idea:

Screenshot from 2023-08-17 19-16-28

PR #2741, opened 2023-08-26
Refresh home design

sujato:


sujato:
Tips for merging. With following two we can simply overwrite the old versions

`client/elements/styles/sc-colors-light.json`
`client/localization/elements/home_en.json`

Issue #2739, opened 2023-08-18
home page title has annoying flash of subtitle


sujato:
Reproduce:

- start on Home page
- scroll down one notch
- scroll back up one notch
- subtitle flashes into view right away, appearing above the main title as it transitions

You can see it clearly if you watch this frame by frame.

[Screencast from 2023-08-18 18-28-52.webm](https://github.com/suttacentral/suttacentral/assets/6112010/20334060-6e5a-42c7-a81e-830888b852b3)

Desired behavior:

Subtitle never overlaps with main title. It only become visible *below* the main title.

Notice that it does not have to change visibility by transforming, as it does now. Another option would be to animate the opacity. Or perhaps both?

PR #2759, opened 2023-09-24
Sujato m3 version3

sujato:
This update fixes many issues with the instant search.

- **remove all js code that is messing with tab order** 
- **make sure all results are `<a>` tags** 

At last the tab order works, because the results are plain, semantic HTML. Tabbing through results:

[Screencast from 2023-09-24 18-38-49.webm](https://github.com/suttacentral/suttacentral/assets/6112010/5ce696f6-7423-48fb-8c92-a44effafc38e)

In addition:

- make html simple and consistent by removing unnecessary spans, etc.
- make class names clear and consistent
- style `ebs` search to look like search, to distinguish search from goto.
- replace incorrect icons and remove unnecessary icons from sc-icons.js: 
    - `search_gray` was wrong style and unnecessary, we already have a search icon
    - `gotolink` which was just a right arrow, and in addition, was a corrupted file.
- organize CSS
- style instant nav results
- add instant-nav description
- make container width consistent with or without scroll bar
- add scroll bar styles for chrome

In addition:

- on Home page, revise styling and animation to fix bugs #2739

## to do

- [x] make sure **all** material components use SC fonts. *Never* let components display Roboto.
- [x] make sure **all** material components work with dark theme.
- [x] review `search-suggestion-link`

```
<a class="search-suggestion-link" href='' @click=${e => this.#gotoSearch(e, item.uid, this.searchQuery)}>
```

Here the `href` is inert, I added it only to make the tab order work. The `instant-nav-link` is formed differently. Should these be consistent? I don't know what the best practice is here.

```
<a class="instant-nav-link" href=${this.#generateURL(item)}>
```


ihongda:
> review search-suggestion-link

I've included the relevant URL, but the click event still needs to be retained to perform some additional actions.

PR #2785, opened 2023-10-10
Sujato instant top close

sujato:
With this I have adjusted a variety of issues with display of instant search sheet.

- show search icon in md text field (fix reversion, typo in js)
- ensure text fields and nav items are evenly aligned (fix style reversion)
- use `max-height: calc(100vh - 160px)` to ensure the sheet footer  is always visible (needs testing!)
- reduce size of footer a little
- adjust display of nav items for better responsiveness
- make text suggestion more like md text field, especially on long queries


ihongda:
>use max-height: calc(100vh - 160px) to ensure the sheet footer is always visible (needs testing!)

I tested it and it seems to be fine. Let me publish it to staging and try it.

PR #2903, opened 2023-12-12
Fix docker-compose network warning

ghost:
Update docker-compose.yml to set the "external" property of the default network to true

### Previous behaviour
```
suttacentral git:(master) ✗ DOCKER_DEFAULT_PLATFORM=linux/amd64 make run-preview-env
docker-compose -f docker-compose.yml down --rmi local --volumes
WARN[0000] network default: network.external.name is deprecated. Please set network.name with external: true 
[+] Running 10/0
```

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -101,8 +101,8 @@ services:
       
 networks:
   default:
-    external:
-      name: nginx-proxy
+    name: nginx-proxy
+    external: true
 
 volumes:
   db-data-volume:


ihongda:
Thank you @ishaanv 🙏🏼

ihongda:
It may be due to version reasons. After merging the code, execute the docker-compose command and the prompt

> networks.default value Additional properties are not allowed ('name' was unexpected)

So I temporarily restored the original code. 🙏🏼

ghost:
Ah right. Sorry for the inconvenience. Which version of docker-compose are we using in production?

Issue #2623, opened 2023-03-12
"live" filter UI


sujato:
One key feature of search is to be able to refine results. 

One way to help would be to offer "live" filter selection. Currently filters are added by hand. 

What if users could do a search, then think, "hmm, too many results, what is in `mn`?" 

Then click the `mn` filter and only `mn` results are shown. We would have to give some thought to the UI, how to show the filters. Hmm.

thesunshade:
My issue https://github.com/suttacentral/suttacentral/issues/2717 seems to be a duplicate of this one.

PR #2921, opened 2023-12-27
Search results live filter prototype.

### LINKED ISSUE #2717: On larger search results, display a breakdown of what books the results are in
#### User story
Someone searches for a rather common term, e.g. "rāga". They get 2610 results:
image

They need some way to narrow things down. At this point they may decide that the sutta they were looking for is in a particular book, or simply that they would be interested in a result that came from a particular book. For example, they think that they might want a pithy verse, so they would like to narrow it down to the Sutta Nipāta.

Currently there is no way for them to know the distribution for their results or an easy way to narrow them (without investigating the "filter" icon).

#### Feature description

Add a break down of results per book. Clicking on one of those book names would add the necessary filter to the search and re-run the search.

The DPR has something like this (see top right of screen):

image

#### Issues:
* Should there be a threshold of x results before this kind of interface is presented?
* Would every single book that has a result be represented? This could lead to having dozens of books with just one result each.

#### Pre milestone planning check:
- [ ] Small enough to completed in a milestone.
- [ ] Dependencies marked
- [ ] No external dependencies block the PBI from being completed.
- [ ] Details are understood by dev team to decide if the PBI can be completed.

#### Done check:
- [ ] Produced code for presumed functionalities
- [ ] Project builds without errors
- [ ] Peer Code Review performed/pull request approved x2
- [ ] Project deployed on the stage environment identical to production platform
- [ ] Feature is tested against acceptance criteria
- [ ] Feature ok-ed by Product Owner (moving to Closed on the Board)
- [ ] Refactoring completed
- [ ] Any configuration or build changes documented (readme, etc...)

ihongda:
Related issues:
https://github.com/suttacentral/suttacentral/issues/2623
https://github.com/suttacentral/suttacentral/issues/2717

### DIFF
--- a/client/elements/sc-page-search.js
+++ b/client/elements/sc-page-search.js
@@ -161,6 +161,7 @@ class SCPageSearch extends LitLocalized(LitElement) {
   #searchOptionsTemplate() {
     return html`
       <div class="search-options">
+        ${this.#searchResultFilterTemplate()}
         <label>
           ${this.localize('search:matchPartial')}
           <md-switch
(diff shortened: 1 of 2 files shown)


thesunshade:
@ihongda I just realized I can checkout your branch and try this out using my local build. Amazing! Thank you for making this prototype. 

This works very well to refine searches. I'm just wondering if we want to start out with a full on GUI for this, or if it should be more tightly integrated with the filter system.

@buddhist-uni, Have you tried this yet?

image


image

The branch is kind of out of date on other things, but it's very easy to test this specific feature.

image

ihongda:
Hmm, it's a prototype, there might be a better place to put these filters, like a top sheet?

PR #2963, opened 2024-01-26
Show included languages in search results.

ihongda:


client/elements/static/sc-static-search-options.js
```diff
@@ -0,0 +1,245 @@
...
+      input,
+      label {
+        cursor: pointer;
+      }
+    `,
+  ];
+
+  get actions() {
+    return {
+      changeDisplaySettingMenuState(display) {
```
agilgur5, line 106: the `actions` method here is unused, with `reduxActions` used instead. the functions are not identical, but similar, so I'm guessing this was re-thought or re-written before the PR was made. it also might have been copy+pasted from the `TopSheetSearchOptions` component, which has [similar unused code](https://github.com/suttacentral/suttacentral/pull/2596#pullrequestreview-4891703571)

I've removed this unused method in #3713

PR #3023, opened 2024-02-22
Clicking outside of "views" overlay doesn't always close it.

### LINKED ISSUE #3001: Clicking outside of "views" overlay doesn't always close it.
#### Description:
If the overlay of the views menu (or any of them) is open you can click outside of it to close, but only if you click in the body of the sutta. This is especially relevant to the views menu because changing some options could easily reduce the size of the sutta area so much that it pulls the footer into the full clickable area.
2024-02-13 09_00_56-general-index xlsx — LibreOffice Calc

#### Steps to reproduce:

* Make sure first that you have "side by side" view turned on
* Go to the last section of [DN16](https://suttacentral.net/dn16/en/sujato?lang=en&layout=plain&reference=none&notes=none&highlight=false&script=latin#dn16:6.27.0)
* Open "views" menu
* Change to "plain" view from side by side view.

Now because the sutta text is shorter, the footer will have been pulled up.

* Try clicking on a non-link area of the footer. 

This will not close the views overlay. But it should.

### LINKED ISSUE #3004: Headings need `word-break: break-word` to prevent forced wide pages
#### Description:

Some title text is so large that it won't fit on a small screen. Instead of the words breaking and wrapping (or even getting a horizontal scroll bar) they keep their full width and force left justified text off the screen, etc.

All that is necessary is `word-break: break-word`.

Of course having soft hyphens in all of these long titles would be ideal. But for now it is impossible to read these pages on mobile and this change would fix that for now.

#### Steps to reproduce:

* https://suttacentral.net/tha-ap392/pli/ms?lang=en&layout=sidebyside&reference=none&notes=asterisk&highlight=false&script=latin#msdiv67

image

#### Expected behaviour:
Titles should break and wrap:
image

### LINKED ISSUE #3005: Verse segments have large gap when references are activated
#### Description:
When references are activated there is a large gap between verse lines and the ref numbers appear on their own line:
image

#### Steps to reproduce:

* go to any sutta with verses and turn on refs, e.g. [Kp5](https://suttacentral.net/kp5/en/sujato?lang=en&layout=plain&reference=main&notes=none&highlight=false&script=latin)

#### Expected behaviour:
Refs should appear at the start of the line just like any other translation:
image

#### Solution
Add
```css
.verse-line {display:block}
```

And remove
```css
.verse-line .translation {display:block}
```

image

image

ihongda:
1. Clicking outside of "views" overlay doesn't always close it. #3001.
2. Headings need word-break: break-word to prevent forced wide pages #3004.
3. Verse segments have large gap when references are activated #3005

### DIFF
--- a/client/elements/sc-page-selector.js
+++ b/client/elements/sc-page-selector.js
@@ -481,24 +481,22 @@ export class SCPageSelector extends LitLocalized(LitElement) {
 
   _changeRoute(location) {
     const [route, params] = this.router.match(location.pathname);
-    if (params) {
-      if (params.categoryId) {
-        params.categoryId = params?.categoryId.toLowerCase();
-      }
-      if (params.langIsoCode) {
-        params.langIsoCode = params?.langIsoCode.toLowerCase();
-      }
-      if (params.authorUid) {
-        params.authorUid = params?.authorUid.toLowerCase();
-      }
-      if (params.suttaId) {
-        params.suttaId = params?.suttaId.toLowerCase();
-      }
-      if (params.word) {
-        params.word = params?.word.toLowerCase();
-      }
+    const normalizedParams = this._normalizeParams(params);
+    this.actions.changeRoute(route, normalizedParams, location.pathname);
+  }
+
+  _normalizeParams(params) {
+    if (!params) {
+      return;
     }
-    this.actions.changeRoute(route, params, location.pathname);
+    const keys = ['categoryId', 'langIsoCode', 'authorUid', 'suttaId', 'word'];
+    const normalized = {};
+    keys.forEach(key => {
+      if (params[key]) {
+        normalized[key] = params[key].toLowerCase();
+      }
+    });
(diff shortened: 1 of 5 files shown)


client/elements/sc-page-selector.js
```diff
@@ -626,40 +624,30 @@ export class SCPageSelector extends LitLocalized(LitElement) {
...
-        break;
-      default:
+    const titleMap = {
+        'search': this.localize('interface:searchResults'),
+        'define': this.localize('interface:dictionaryResults'),
+        'home': 'SuttaCentral',
+        'navigation': '',
+        'suttaplex': '',
+        'sutta': '',
+        'publicationEditionMatter': '',
```
agilgur5, line 634: Hmm, this logic is actually different from the previous `switch` statement as there are no longer early returns. These components already set their own title, while this sets it to an empty string. It shouldn't be set here at all -- I've fixed this in #3726

PR #3041, opened 2024-02-26
add missing localization and remove title attributes

thesunshade:
## Localization

* bottomsheet help text + padding css:

image

* Buddhism.net text in footer
* Buddhism.net card on homepage:

image

Note: When I did this I also changed the subhead and text. The subhead was too long for the design. And I felt that the text was just lifted from their homepage, so I made it fit better in this context. Here is the original if you want to keep it:

> making the teachings of the Buddha accessible  
> At Buddhism.net, we hope to help you 1) learn Buddhism, 2) develop a practice, and 3) discover teachers and communities. Buddhism.net is cross-sectarian. We focus on early Buddhism, the common root of all modern forms of Buddhism. We aim to offer you a solid foundation in early Buddhism, which will be immensely helpful to you regardless of which form of Buddhism you adopt.

My new version appears in the screenshot.

## title attributes

I removed the code for the following from `home` as they are titles that add no significant value and are useless on mobile. This is what they look like in the json file:

```
  "home:2": "SuttaCentral promotional video 2020",
  "home:16": "Visit SuttaCentral Voice",
  "home:21": "Visit BuddhaNexus",
  "home:24": "Visit SuttaCentral forum",
  "home:30": "Visit Bilara translation webapp",
```

## Warnings:

These changes will require accepting a new PR on `bilara-data`

The old Bilara card on the home page was hard coded in, in the sense that the template for that card is called ` #bilaraCardTemplate() {`.

The code might be more robust if all the card templates were simply renamed to `#card1`, `#card2~ etc.

Because I recycled segments `home:31` and `home:31` , in cases where the homepage has been translated, that text will be wrong. I guess somehow that has to be dealt with.

### DIFF
--- a/client/elements/addons/sc-bottom-sheet.js
+++ b/client/elements/addons/sc-bottom-sheet.js
@@ -1,6 +1,7 @@
 import { LitElement, html, css } from 'lit';
 import { LitLocalized } from './sc-localization-mixin';
 import { icon } from '../../img/sc-icon';
+import { unsafeHTML } from 'lit/directives/unsafe-html.js';
 
 export class SCBottomSheet extends LitLocalized(LitElement) {
   static properties = {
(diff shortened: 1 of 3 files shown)


ihongda:
Thanks @thesunshade 🙏🏼

PR #3042, opened 2024-02-26
Remove Unused Numpy Dependency

khemarato:
The numpy requirement doesn't seem to be used anywhere, and since this is a fairly large import, its best to not include it unless its really needed.


ihongda:
Thanks @buddhist-uni 🙏🏼

PR #3056, opened 2024-02-28
Allow proper debugging of the Python Server

khemarato:
This PR (hopes to eventually) add support for debugging the python server with the VSCode Debugger tools.

### DIFF
--- a/docker-compose.debug.yml
+++ b/docker-compose.debug.yml
@@ -1,0 +1,44 @@
+version: '3'
+
+services:
+
+  sc-flask:
+    command: bash docker/entrypoint.debug.sh
+    env_file:
+      - server/env/.dev.env
+    volumes:
+      - ./server/:/opt/sc/sc-flask
+      - ./client/:/opt/sc/frontend
+    ports:
+      - '5000:5000'
+      - '5679:5679'
+
+  sc-nginx:
+    volumes:
+      - ./client/:/opt/sc/static
+    command: bash entry-point.dev.sh
+
+  sc-arangodb:
+    env_file:
+      - server/env/.dev.env
+    ports:
+      - '8529:8529'
+
+  sc-swagger:
+    ports:
+      - '8080:8080'
+
+  sc-frontend:
+    command: npm run dev
+    volumes:
+      - ./client/elements:/opt/sc/frontend/elements
+      - ./client/utils:/opt/sc/frontend/utils
+      - ./client/img:/opt/sc/frontend/img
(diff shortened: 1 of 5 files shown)


docker-compose.debug.yml
```diff
@@ -0,0 +1,44 @@
...
+  sc-flask:
+    command: bash docker/entrypoint.debug.sh
+    env_file:
+      - server/env/.dev.env
+    volumes:
+      - ./server/:/opt/sc/sc-flask
+      - ./client/:/opt/sc/frontend
+    ports:
+      - '5000:5000'
+      - '5679:5679'
```
khemarato, line 14: We will need to expose a port for the debugger

docker-compose.debug.yml
```diff
@@ -0,0 +1,44 @@
+version: '3'
+
+services:
+
+  sc-flask:
+    command: bash docker/entrypoint.debug.sh
```
khemarato, line 6: Point to a new debug entrypoint

nginx/conf.d/prod.nginx.conf
```diff
@@ -59,6 +59,16 @@ server {
     }
 
     include "/etc/nginx/configurations/sc.nginx.conf";
+    
+    location /api/ {
```
khemarato, line 63: Moved in from sc.* because the dev (/debug?) config will have to be different.

nginx/conf.d/dev.nginx.conf
```diff
@@ -12,4 +11,8 @@ server {
         proxy_set_header Upgrade $http_upgrade;
         proxy_set_header Connection "Upgrade";
     }
+    location /api/ {
```
khemarato, line 14: This is broken and I'm not sure why.  Any ideas @ihongda?

server/docker/entrypoint.debug.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/env bash
+cd server
+echo "Waiting for arango to start"
+python wait_for_arango.py
+python manage.py migrate
+pip install debugpy
+cd ..
+touch /tmp/.done.info
+python server/wsgi.py # eventually I want this to start a debugpy server
```
khemarato, line 9: The main difference here is that I'm trying to call the wsgi.py server file directly with python instead of going through the uwsgi program.  This will be necessary to use the python debugger, but I haven't yet figured out how to hook up nginx to this 'raw' python server.

khemarato (commented):
NOT WORKING YET.  Help appreciated :)

Issue #2943, opened 2024-01-12
Site page titles should have the actual page title first in the title before "SuttaCentral" instead of after


thesunshade:
#### Description:
When you have multiple site pages open, it becomes impossible to tell what they are because they all start with "SuttaCentral"

#### Steps to reproduce:

Open up enough site pages in your browser so that titles begin to be cut off.

Here is what it looks like with just a few:

![2024-01-12 15_26_46-SuttaCentral—ပင်မစာမျက်နှာ — Mozilla Firefox](https://github.com/suttacentral/suttacentral/assets/82448383/b6e48d36-bc77-49df-a516-ebca5b26f4bf)

So far no problem. But open any more and you see this:
![2024-01-12 15_27_39-SuttaCentral—ပင်မစာမျက်နှာ — Mozilla Firefox](https://github.com/suttacentral/suttacentral/assets/82448383/e8dce7e9-35e6-46c6-a3dd-8c05cf7c72d2)

Tabs are indistinguishable.

#### Expected behaviour:

The actual page title should come first, not the site title. So instead of 
```
SuttaCentral—introduction
```

It should have 
```
Introduction—SuttaCentral
```

Btw, why are the page titles lower case?

PR #3059, opened 2024-02-29
Put unique page title of static pages first in tab title

thesunshade:
This is an extremely minor change to achieve this issue: https://github.com/suttacentral/suttacentral/issues/2943


ihongda:
Thanks @thesunshade

Issue #2972, opened 2024-01-30
`list authors` output needs to show the translator ID, not their short name


thesunshade:
#### Description:

For example, `list authors` shows this:
![image](https://github.com/suttacentral/suttacentral/assets/82448383/d4b5c545-9869-4523-b469-a0e8e778df46)

That would lead you to the conclusion that `author:Алокананда` would work. But it doesn't. It must be `author:alokananda`

And it's not just non-latin names.

![image](https://github.com/suttacentral/suttacentral/assets/82448383/16f2d3fb-75f4-4984-a412-27d1d39e801e)

No way from looking at this to tell that you need to search for `author:cowell-rouse`

#### Expected behaviour:

The first column of the table should be the author ID, not whatever the short form of the authors' name is used now.

thesunshade:
The author list feature is really great, btw.

PR #3074, opened 2024-03-06
Put UIDs in list authors table

thesunshade:
This addresses https://github.com/suttacentral/suttacentral/issues/2972

I also changed the list so that hovering over either item in the list (uid and name). It's not possible to wrap a whole row in `<a>` tags. 

Ideally I think this would be better as flexbox cells. Then it could treat whole rows as a link. 

I wrapped the UIDs in `<code>` tags. Not sure if that is the best thing to do. The point is to emphasize that that is the part that is used in the filter.

I have also added instructions and table headers.

All has been tested on my local build.

Note: For some reason the columns are not aligned properly. I'm not sure what is causing that since I didn't change any css. Perhaps it is related to the changes in the `volpage:` table?


ihongda:
The display of new modifications in mobile browsers needs to be improved.

Maybe we could just keep showing the author's full name?

thesunshade:
> The display of new modifications in mobile browsers needs to be improved.

I agree.

> Maybe we could just keep showing the author's full name?

Well, people need a way to see the UID so they can use it in the `author:` filter.

I think the problem with the mobile is mainly because the cells wrap. This was needed on the `volpage:` table but not here. If I remove the 100% width it improves:
image

ihongda:
screenshot_2024-03-08_01-10-53

I made some modifications, using div instead of table, and made responsive optimization.

screenshot_2024-03-08_01-12-07

thesunshade:
Looks great! Thanks!

Issue #3067, opened 2024-03-04
Make the search urls more human readable by substituting spaces with `+` signs


thesunshade:
#### User story
Jackie wants to share a search with someone. So they copy the url and paste it into an email. They get something like this:
```
https://suttacentral.net/search?query=volpage:DN%20ii%2055
```
It's a bit hard to read. Lets clean it up.

#### Feature description
Before writing to the URL, substitute `+` signs for the spaces. That would change the above url to this:
```
https://suttacentral.net/search?query=volpage:DN+ii+55
```
Do the same for other things as well
```
https://suttacentral.net/search?query=author:sujato%20kassapa%20OR%20moggallana
becomes
https://suttacentral.net/search?query=author:sujato+kassapa+OR+moggallana
```
This would require that the string be processed both when writing to as well as reading from the URL bar.

PR #3101, opened 2024-03-25
replace spaces in url Search params with `+` sign

thesunshade:
I believe this PR will accomplish the following issue: https://github.com/suttacentral/suttacentral/issues/3067

I have tested it on my local version. I honestly can't say how robust this is, but I think it might be fine.

EDIT: I have also added 2 localization segments unrelated to the search urls.

@buddhist-uni , do you have any thoughts?

### DIFF
--- a/client/elements/navigation/sc-navigation.js
+++ b/client/elements/navigation/sc-navigation.js
@@ -253,8 +253,8 @@ export class SCNavigation extends LitLocalized(LitElement) {
     document.dispatchEvent(
       new CustomEvent('metadata', {
         detail: {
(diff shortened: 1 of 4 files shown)


thesunshade:
So, ChatGPT is suggesting the following:

```python
from urllib.parse import unquote_plus

query = request.args.get('query', None, type=str)
if query:
    query = unquote_plus(query)
```

What do you think?

khemarato (commented):
Nice. Looks reasonable to me 😊

khemarato:
> ChatGPT is suggesting

This is unnecessary as the proper decoding is already happening elsewhere in the server framework.

PR #3120, opened 2024-04-04
add more localization support

thesunshade:
Adds more localization support for menus and titles.

Will need to be accompanied by the acceptance of a corresponding pr in `bilara-data`: https://github.com/suttacentral/bilara-data/pull/3386


ihongda:
Thanks @thesunshade 👍🏼

PR #3142, opened 2024-04-16
Add dpd word lookup

cittadhammo:
I have added the dpd to Suttacentral pali lookup. I have used the same method as in [The Buddha's Words](https://thebuddhaswords.net/home/index.html) website

See https://github.com/digitalpalidictionary/dpd-db/tree/feature/db-relations/tbw for explanation

It looks good 

dpd

I have not used the arango database, but simply the json file from the link above. I don't know what is the best approach, loading js or feeding the data in the database and making request. I have not used much db in my programming life.

These file could be fetched automatically from dpd repo when building Suttacentral for prod.

The code needs to be cleaned, and I don't know if you want to keep the old lookup function. Do you want to display the result of the old lookup function as well?

Anyway, at least it is a proof of concept and can be tested and talked about ;-)


cittadhammo:
FYI @theSunShade, @bdhrs

- need to accommodate other language (revert to old pali lookup if other than english selected ?)

thesunshade:
Wow! that was fast. This is exciting.

I'm getting the following error when I try to view a sutta:

image

and this is what I see in the console:
image

cittadhammo:
@thesunshade I think you need to rebuild the front end 

```
docker-compose build sc-frontend
```
then
```
make run-dev
```

thesunshade:
> I think you need to rebuild the front end

Hmm. I just did that and I'm still getting the same error. As far as I can tell I'm using the correct branch (I see your changes)

@buddhist-uni do you have time to test this PR?

khemarato:
After you rebuild the docker container did you ctrl+c and restart `make run-dev` to pick up the new image?  It's working on my machine... (Except that it's always returning English definitions, even when e.g. Portuguese lookup is selected)

cittadhammo:
> Except that it's always returning English definitions, even when e.g. Portuguese lookup is selected

Yes, as far as I understand, the dpd is only in English ? I can revert to the old lookup function if another language than English is selected.

thesunshade:
> did you ctrl+c and restart `make run-dev` to pick up the new image?

I did not! That fixed it. I obviously still don't have a correct mental model of how this whole thing works.



> Yes, as far as I understand, the dpd is only in English ?

I believe that the expectation is that when the lookup is selected to be one of the other languages, like Portuguese then it will use the existing Portuguese dictionary.

And since those dictionaries are analogous to the NCPED that is the existing English dictionary on the production site, I think the lookup logic will have to be the existing lookup logic. Not sure if that makes sense. Basically if the language selected is English, then use the DPD, deconstructior, etc, but for any other language, use the existing code.

ihongda:
Ven. @cittadhammo  Thank you for your efforts in spreading the teachings of Buddha 🙏🏼 
Sàdhu! Sàdhu! Sàdhu!

I tested the code and it works fine 👍🏼 .

In addition, would it be better to move the `/files/dpd` directory to `/client/elements/lookups` ?

bdhrs:
@ihongda As DPD is a work in progress, every month on full moon Uposatha I will have a dictionary update of the relevant JSON files. How would you like to handle that? Please let me know the repo and location so I can automate the pull request.

ihongda:
> @ihongda As DPD is a work in progress, every month on full moon Uposatha I will have a dictionary update of the relevant JSON files. How would you like to handle that? Please let me know the repo and location so I can automate the pull request.

Thank you for your efforts in spreading the teachings of Buddha 🙏🏼

Currently, sc-data repo has a `dpd` directory for placing related files:

https://github.com/suttacentral/sc-data/tree/main/dpd

The data here is converted and used to import into the database.

Issue #2733, opened 2023-07-31
Replace Pali lookup with new system from digitalpalidictionary (DPD)


sujato:
Our Pali lookup was originally created by Blake based on the then best appraoch. A new approach that is far more accurate has been created by the Digital Pali Dictionary project.

We should investigate how this works and see if it can be applied to SC and Bilara.

<details><summary>related but deleted files</summary>
<p>

https://github.com/digitalpalidictionary/dpd-db/blob/main/tbw/tbw_exporter.py

https://github.com/digitalpalidictionary/dpd-db/tree/main/tbw

https://github.com/digitalpalidictionary/dpd-db/tree/main/tbw/output

</p>
</details>

thesunshade:
Starting to look at this issue, it appears that /suttacentral/client/elements/lookups/sc-lookup-pli.js` is the code that needs to be modified on SC.

There is this file on the TheBuddha'sWords repo that I believe is used for the definition lookup: 

https://github.com/thebuddhaswords/BW2/blob/main/js/dpd_deconstructor.js

thesunshade:
So I wanted to dig in and see what might actually happen when we change this. So I went to [MN111](https://suttacentral.net/mn111/en/sujato?lang=en&layout=sidebyside&reference=none&notes=none&highlight=false&script=latin) and clicked on `puthupañño`

![image](https://github.com/suttacentral/suttacentral/assets/82448383/c420c2de-e7fc-4e6c-a842-3bd1f4ee71cf)

This breaks the word in two and gives links to two separate sc definition pages: [puthu](https://suttacentral.net/define/puthu) and [pañña](https://suttacentral.net/define/pa%C3%B1%C3%B1a)

I'm assuming that we would be using similar data to that found [here](https://github.com/thebuddhaswords/BW2/tree/main/js) that includes the deconstructor data that splits compounds in [dpd_deconstructor.js](https://github.com/thebuddhaswords/BW2/blob/main/js/dpd_deconstructor.js), the inflected to the non-inflected (aka headword) in [dpd_i2h.js](https://github.com/thebuddhaswords/BW2/blob/main/js/dpd_i2h.js) and finally the definition in [dpd_ebts.js](https://github.com/thebuddhaswords/BW2/blob/main/js/dpd_ebts.js)

[...]

PR #3166, opened 2024-05-09
Rewrite Pali lookup according to `DPD`.

ihongda:
Thanks to @cittadhammo for contributing the code. 🙏🏼 

1. Prioritize using DPD way for lookup, If no relevant entries are found, use the old method to search again.
2. In text pages, Use DPD dictionary instead of NCPED dictionary.
3. On the Pali word definition page, add DPD dictionary content. e.g.
image

need to execute `make laod-data` to load latest DPD dictionary.

Related issue: 
https://github.com/suttacentral/suttacentral/issues/2733


cittadhammo:
This is interesting, I don't quite understand how it works between the different functions in the `client/elements/lookups/sc-lookup-pli.js`

there are : 

- `lookupWord(wordToLookup)` that returns  `this.lookupWord_old(wordToLookup)`
- `lookupWord_old(wordToLookup)` that uses `this._lookupWord(word)`
- `_lookupWord(word)` that does a lot...

If they are happy together, I assume it is all working ! Sadhu !

By the way, if you could enlighten me about the underscore _functions(), I was wondering what was the difference with them.

ihongda:
@cittadhammo 🙏🏼 Good morning.

image

Only when the word cannot be matched using DPD, `this.lookupWord_old(wordToLookup);` will be executed.

`lookupWord_old(wordToLookup)` will split the given word according to the rules to determine whether there is a matching entry.

Adding an underscore generally means that the method is private.

cittadhammo:
Good, thank you. This is what I thought as well, that the function with underscore were only used in the file itself.

I can see the "if" statement now, ;-) thank you for pointing it out!

Issue #2932, opened 2024-01-07
Filters should be possible at any place in search string


thesunshade:
Currently, filters are only parsed when they appear at the beginning of a search string.

## User Story

Say someone does a search for `meditation`. Of course they get too many results, so they think, "Hmm, why don't I limit my search to the Samyutta Nikaya."

So they add `in:sn` to the end of their search string to get `meditation in:sn`. It's normal to add things to the end when something is already in the search field, but also one may have the natural language thought, "I want to look for all occurances of the word `meditation in:sn`". However they get no results because it should have been entered as `in:sn meditation`.

## Expected behaviour

Filters should be parsed no matter where they exist in the search string. This also includes `author` filters. For example, `in:dn meditation author:sujato` fails but `in:dn author:sujato meditation` does not.

thesunshade:
https://github.com/suttacentral/suttacentral/issues/2623 may also relate to this.

mtnadi:
I've got it working https://github.com/mtnadi/suttacentral/tree/nadi-issue-2932, however before I open a PR, it only handles `in:` and `author:`.  I ran `make test-server` fine.

Do we want it to do more first?  Are there other tests to run?

thesunshade:
Oh wow! Thanks.

The `volpage` and `ref` don't need to be combined with anything else. I don't have much experience using `title`.

I think that `in` and `author` are the only ones that really need it.

mtnadi:
Okay, that's good (for now!) as `title` seems to have more issues around it...!  Shall open PR for now.

thesunshade:
This is working live on the site now. 

Thanks so much for this!!!

Issue #2970, opened 2024-01-30
Change `author:` fiter to `by:`


thesunshade:
With this I'm not proposing any functional change. I think it would be beneficial for a few reasons:

* It's easier for everyone to type
* It's easier for non-English speakers
* It takes up less space in the urls.
* Authorship is kind of an odd thing with EBTs. Using `by` would fit the English phrase of "translations by so and so" as well as "manuscripts created by such and such organization"

thesunshade:
This does create a small problem with the command `list authors`. 

My idea is that the `list` command should be available for all filters and the parameter should always be the exact text of the filter. For example
```
list by
list in
list for
```

khemarato:
There should be an overlap period where both are supported for a while before deprecating `author`  Agreed that a warning would be nice during that period, informing people about the change

thesunshade:
I guess my thought is that it's still early enough in the life of the new search that there may not be a lot of people using the author filter. So supporting both makes the code more complicated (I'm assuming) and that just has to be uncoded later. Since I hope we will have error reporting anyway, I thought that would be a simpler solution.

ihongda:
The relevant changes have been pushed to Production.

PR #3189, opened 2024-06-14
Fix: handle author and collection filters anywhere in search string

mtnadi:
refs #2932

### DIFF
--- a/server/server/search/instant_search.py
+++ b/server/server/search/instant_search.py
@@ -1644,24 +1644,19 @@ def extract_query_conditions(param):
 
 
 def extract_rest_keywords(result, param):
+    """
+    Note:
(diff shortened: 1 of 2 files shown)


mtnadi:
b80fb6a23 will hopefully make #2970 easier, but I'm not sure if things were hard-coded for performance?

thesunshade:
> I'm not sure if things were hard-coded for performance?

I'm not 100% sure, but I believe that the code just never got refactored to include the constants. Using them is good.

ihongda:
Thanks @mtnadi 🙏🏼

thesunshade:
@mtnadi Thank you so much for this. I know that without this fix there would be many times people get zero results because of not using the exact format. It's really valuable. Thanks!!

PR #3236, opened 2024-08-04
Bump client element localization files

khemarato:
This is a completely automated set of changes pulling in the latest strings.


khemarato:
@ihongda - This is just a simple automated string pull. Can it be merged?

Issue #2308, opened 2021-10-22
Keyboard shortcuts


sujato:
## idea

Can be cool, right? Hit <kbd>ctrl</kbd> + <kbd>k</kbd>, a dialog pops up with some common actions: search maybe? Or something else?

There's a thing called <ninja-keys> that does this, and it's lit!

~https://ssleptsov.github.io/ninja-keys/~
https://github.com/ssleptsov/ninja-keys

Are there good use cases for this, or is it just a fancy dream? :thinking: 

## others

Some uses out there:

- slack: unread, channels
- notion: search in your content
- linear.app: search, support, feedback, invite, log out, create workspace etc., jump to things, lots of other stuff
- oh yeah, and now, right here on Github! 
     - https://twitter.com/hey_sergei/status/1453392498756292629

## for SC

We should look to solving an existing problem. I don't think access to search is a big deal; it's there in the toolbar on every page. 

With our current navigation, however, I can see a few problems that this might address.

## cost 

Before starting, let's consider the *cost*. To make this useful we must pay the following costs:

- developer time, and consequently *not* doing other things
- clients must download and execute js.

To use it, people must:

- know it exists
- know the shortcuts
- do the following steps:
    - activate the shortcut
    - select the command
    - execute the command

So to justify it, we have to be confident that it offers a substantial improvement over what is already there.

## history

Readers might often navigate between several suttas, and it can be hard to keep track. Perhaps we could remember them for them?

- when a user arrives at a sutta, store the URL.
- keep a list of, say, the last ten visited suttas
- when the keyboard is activated, show the list.

thesunshade:
> Sounds like this would mesh well with an instant search kind of thing

In Joplin, the command palette and instant search are the same interface, but they have two shortcut keys. ctrl p for instant search and ctrl shift p for command palette. The only difference is that the command palette searches start with a colon.

thesunshade:
I'd like to revisit this.

Re-reading from the top it sounds like although the title of the issue is "Keyboard shortcuts," what you are really talking about is a command palette. 

Personally I think that hot keys (e.g. pressing `v` to scroll through the three view options) would be more useful and more understandable. I know not everyone is a dedicated keyboard shortcut user, but the concept is ubiquitous. 

A command palette is a much newer thing. Also, I think it is well suited for apps that have hundreds of different commands and the user may not be able to memorize shortcuts for all of them (or even where they sit in a menu structure). 

[...]

PR #3237, opened 2024-08-04
Add keyboard shortcut: "V" to toggle between TextViews

khemarato:
This PR (built on top of https://github.com/suttacentral/suttacentral/pull/3236 so please ignore the extra string changes) takes a first stab at one of the keyboard shortcuts in https://github.com/suttacentral/suttacentral/issues/2308


client/elements/addons/sc-top-sheet-views.js
```diff
@@ -183,6 +183,14 @@ export class SCTopSheetViews extends LitLocalized(LitElement) {
       label {
         color: var(--sc-on-tertiary-primary-text-color);
       }
+      
+      code.key {
```
khemarato, line 187: This is what this looks like in both light and dark themes:

Screenshot from 2024-08-04 14-29-08

Screenshot from 2024-08-04 14-29-18

Happy to change this styling to conform to whatever style guidelines you have.

client/elements/addons/sc-top-sheet-views.js
```diff
@@ -330,6 +338,31 @@ export class SCTopSheetViews extends LitLocalized(LitElement) {
     super.connectedCallback();
     this._loadScToasts();
     this._fetchReferenceDisplayType();
+    document.addEventListener('keydown', this._handleKeydown.bind(this));
+    window.setTimeout(() => {
+      this.show();
+      this.hide();
+    }, 2000);
```
khemarato: This timeout is absolutely necessary to make it work, but I can't for the life of me figure out why.  Without this `show/hide`, the toast pops up showing that the key press was registered, but the layout doesn't actually change.  Any ideas, @ihongda ?

client/elements/addons/sc-top-sheet-views.js
```diff
@@ -330,6 +338,31 @@ export class SCTopSheetViews extends LitLocalized(LitElement) {
...
+  }
+
+  // This componenet is _never_ disconnected the way the code is written today (2024-08-04)
+  // Still, adding this in case some day this component is properly unloaded.
+  disconnectedCallback() {
```
khemarato, line 345: I was expecting this callback to be called when navigating from a sutta page to e.g. the homepage, but it wasn't called.  It seems that the top sheets are never unloaded?

client/elements/menus/sc-action-items.js
```diff
@@ -304,6 +304,13 @@ export class SCActionItems extends LitLocalized(LitElement) {
     this.#displaySearchOptionsButtonStateChange();
     this.scSiteLayout = document.querySelector('sc-site-layout');
     this.#setToolBarLanguage();
+    document.addEventListener('keydown', this._handleKeydown.bind(this));
+  }
+
+  _handleKeydown(e) {
+    if (e.key === 'Escape') {
+      this.#closeAllTopSheetsOfTextPage();
```
khemarato, line 312: While I'm here, I also added a keyboard shortcut to close all the top sheets: `esc`

client/localization/elements/viewoption_en.json
```diff
@@ -35,7 +35,7 @@
   "viewoption:sideBySide": "Side by side",
   "viewoption:sidenotes": "Show as sidenotes",
   "viewoption:textView": "Original text and translation",
-  "viewoption:textViewDescription": "Some texts on SuttaCentral are “segmented”; the original text and translation are approximately matched line by line. For such texts, see both root text and translation.",
+  "viewoption:textViewDescription": "Some texts on SuttaCentral are “segmented”; the original text and translation are approximately matched line by line. For such texts, see both root text and translation. Press <code class=\"key\">V</code> to toggle between the views.",
```
khemarato: The other string changes were pulled in automatically by the build (from where?) except for this one which I changed to get the helptext updated.  What is the correct way to update this string, @sujato ?

khemarato (commented):
cc @thesunshade

thesunshade:
Thanks!!!!!

I am able to break it, though. 

1. Go to an alligned sutta, e.g. MN 5.
2. Cycle through the views by pressing `v`.
3. Click on the parent vagga in the breadcrumb trail
image
4. click on a new aligned translation, e.g. MN 6
5. Cycle through the views by pressing `v`.

**What doesn't happen:**
* the actual view does not change (meaning that the text itself does not change)

**What does happen**
* the snack bar announces the change
* the params in the url change

**What probably happens**
* the radio buttons change

HOWEVER… once you click on the "views" menu, then the change is applied and the new view is shown. After that, the `v` toggle works just fine.

Note, the `esc` key seems to work immediately, which probably makes sense. Opening the "info" panel doesn't get the view toggling working, which probably makes sense.

client/elements/addons/sc-top-sheet-views.js
khemarato: Thanks for playing with it @thesunshade ! Your bug is certainly the same one this is attempting to work around. It will definitely need a real fix before this can merge. That's waiting on @ihongda's knowledge: what gets hooked up only after the top panel shows on a given sutta?

client/localization/elements/viewoption_en.json
thesunshade: I think this has to be updated in `bilara-data` and then go through all its magic.
I'm wondering if this help text should get its own new segment.

client/elements/addons/sc-top-sheet-views.js
ihongda: > This timeout is absolutely necessary to make it work, but I can't for the life of me figure out why. Without this `show/hide`, the toast pops up showing that the key press was registered, but the layout doesn't actually change. Any ideas, @ihongda ?

Bhante 🙏🏼 

This part of the code can be placed in the `firstUpdated` event of lit.
>Called after the component's DOM has been updated the first time, immediately before [updated()](https://lit.dev/docs/components/lifecycle/#updated) is called.

In the firstUpdated event, we should be able to remove the `setTimeout`.

client/elements/addons/sc-top-sheet-views.js
```diff
@@ -330,6 +338,31 @@ export class SCTopSheetViews extends LitLocalized(LitElement) {
...
+    super.disconnectedCallback();
+    document.removeEventListener('keydown', this._handleKeydown);
+  }
+
+  _handleKeydown(event) {
+    switch (event.key) {
+      case 'v':
+      case 'V':
+        var idx = this.textViewArray.findIndex(item => item.textView === this.selectedTextView);
```
ihongda: Using `const` or `let` instead of `var` is more in line with the specification 🙏🏼

client/elements/addons/sc-top-sheet-views.js
ihongda: Yes, this component is not actively uninstalled from the DOM.

client/localization/elements/viewoption_en.json
khemarato: @thesunshade - Can you take the lead on this upstream change? :pray:

thesunshade: Sure. Just give me a poke when the shortcut is live if I don't notice.

client/elements/addons/sc-top-sheet-views.js
khemarato: What are you talking about @ihongda? You mean put the addEventListener to the firstUpdated callback instead of the connectedCallback?  Cause that doesn't work.  I'm telling you that there's something that's only getting hooked up after the top sheet actually _opens_ (firstUpdated is called before that).  There's some function I need to call here (or some other logic somewhere else that I need to change?) to make sure that `actions.chooseSegmentedSuttaTextView` actually does what it's supposed to.

khemarato: @ihongda - Okay. I figured it out. `sc-text-bilara.shouldRestoreUserSettings` is `false` by default and is suppressing the user-selected setting until the user opens the control panel, I assume to allow the url params to take effect without changing the user's settings.  In this case, the correct thing to do (?) is to set shouldRestoreUserSettings to `true` after the user hits one of the keyboard shortcuts.  Will update the PR with that logic.

client/elements/addons/sc-top-sheet-views.js
```diff
@@ -332,6 +340,27 @@ export class SCTopSheetViews extends LitLocalized(LitElement) {
...
+  }
+
+  _handleKeydown(event) {
+    switch (event.key) {
+      case 'v':
+      case 'V':
+        let idx = this.textViewArray.findIndex(item => item.textView === this.selectedTextView);
+        if (idx >= 0) {
+          idx = (idx + 1) % this.textViewArray.length;
+          this.#setTextBilaraPageIsTextOptionsMismatchSavedSettingsState();
```
khemarato, line 358: @ihongda - this is the magic function call I needed to make it work.

ihongda: 👍🏼

ihongda (approved):
Thanks, Bhante 🙏🏼

thesunshade:
My suggestion would be to hold off on the interface changes for now. There are some folks who will start localizing things immediately and it's probably good to have further discussions on the best way to notify users of the hot-keys.

khemarato:
The string change here will get overwritten on the next `make` right?

thesunshade:
> The string change here will get overwritten on the next `make` right?

Yes, sorry. I was speaking more towards HongDa. I do believe that your change will be lost.

It's a discussion to be held probably in the issue, but I'm not even sure what Bhante had in mind. There will be a few that won't have an easy place to note in the existing interface.

BTW, I forgot to test what happens when you type a `v` when entering something in the search field. Did you?

khemarato:
Thanks @thesunshade - That was indeed getting captured. Somehow I thought it would automatically ignore `input` keypresses.  Anyway, that's now fixed and I've removed the string change. :)

PR #3249, opened 2024-08-19
Add Toasts for changing ref/notes Display Types

khemarato:
This PR prepares the ground for adding keyboard shortcuts for these settings, in that those keyboard shortcuts will need to trigger toasts announcing the changes to the user.


client/localization/elements/viewoption_en.json
```diff
@@ -12,11 +14,13 @@
...
   "viewoption:lookup": "Lookup",
   "viewoption:lookupDictionaryEnabled": "The {lookupDictionary} lookup dictionary is now enabled.",
+  "viewoption:mainRefsEnabled": "The main references are now enabled.",
+  "viewoption:mainRefsDisabled": "The main references are now disabled.",
   "viewoption:none": "None",
   "viewoption:noteDescription": "Where available, view comments on the translation and variant readings for root texts.",
-  "viewoption:noteDisplayTypeAsterisk": "Tooltip on asterisk",
-  "viewoption:noteDisplayTypeNone": "None",
-  "viewoption:noteDisplayTypeSidenotes": "Display as sidenotes",
+  "viewoption:noteDisplayTypeAsterisk": "Notes and variants in segmented texts will be shown as tooltip on asterisk.",
```
khemarato: @thesunshade - It turns out that these three strings are unused, so I decided to reuse them.  Happy to not reuse them and instead use new strings, but the new strings will definitely need to use the "Asterisk" "None" "Sidenotes" naming convention.

thesunshade: The only problem with reusing them is that if they have already been translated, there is no automatic alert to tell translators that they need to re-translate.

We also need to be a bit cautious in declaring something unused simply by searching for a key in the code. Sometimes the keys are dynamically constructed. I fell into that trap in the past. That may not be the case here, but in general I recommend not reusing keys.

thesunshade: Wait… aren't those keys/values used for the radio button labels? They have been translated into 22 languages.

image

thesunshade: Yes. it looks like those are used here:

https://github.com/suttacentral/suttacentral/blob/0fed0c638b5b51fb322fcf81fa3356a482d35ba2/client/elements/addons/sc-top-sheet-views.js#L475

khemarato: Compare `viewoption:sidenotes` to `viewoption:noteDisplayTypeSidenotes`  Which string is actually being used? 😉

thesunshade: Yeah, I'm not sure. But the second one is for sure being used. So we can't recycle those keys.

khemarato: I am sure.

358981898-ad127b85-ee77-4773-92f7-ee0e47dde0a6

This string says "Display" not "Show" Also, if you put a breakpoint on the line you pointed out above, you'll see what values `displayType` can take: e.g. `sidenotes`, never `noteDisplayTypeSidenotes` 😊

thesunshade: Yeah, my rule is just don't recycle the keys.

Is there some pressing reason we need to?

thesunshade: Of course cleaning up unused strings is a good thing. But we did a cleanup recently so I'm inclined to just let things be for now.

khemarato: The only pressing reason to clean them up is I just did the work to make sure these were unused. Better to save future us from having to do that work again is all.

thesunshade: Sadhu!

Once everything is up and running I can go through and delete the translations of the strings too.

khemarato: Great!  And I'll use a unique string name for these :smile:

ihongda:
Thanks, Bhante @khemarato 
I also updated the related viewoption key in Bialra-data.

PR #3250, opened 2024-08-29
Use <kbd> instead of code.key for shortcut icons

khemarato:
Thanks Bhante for pointing out the semantically correct HTML to use for keyboard letters.


ihongda:
Thanks, Bhante @khemarato 🙏🏼

PR #3252, opened 2024-08-31
Add N shortcut for notes

khemarato:
Continues to make progress on issue #2308 by adding the shortcut for cycling though notes display modes.  This PR is heavily inspired by https://github.com/suttacentral/suttacentral/pull/3237


thesunshade:
@khemarato I tried this out locally but it seems to have the same problem as the previous shortcut did originally.

It will only work after opening and closing the views panel at least once.

So to recreate

1. navigate to a aligned sutta
2. press <kbd>n</kbd>
3. toast displays but the note view doesn't change
4. Open and close view panel
5. press <kbd>n</kbd>
6. Works as expected

ihongda:
Thanks, Bhante @khemarato @thesunshade .

Issue #3255, opened 2024-09-03
Opening one of the panels should put the whole panel in focus


thesunshade:
#### Description:

This is what you see when you open the parallels panel:
![image](https://github.com/user-attachments/assets/76a19e99-4b76-4ea0-91ee-800c9247d5e3)

In order to bring the panel into focus the user has to press tab twice. Note the blue border:

![image](https://github.com/user-attachments/assets/137dd51f-1801-49a2-af2a-88eee2d06d7e)

Opening the info panel takes four tabs.

This will become more relevant when there are keyboard shortcuts to open the panels.

PR #3271, opened 2024-09-12
Add I O P Shortcuts

khemarato:
Continues work on #2308 by adding shortcuts to open the top tool panels when viewing a text.

cc @ihongda and @thesunshade


thesunshade:
@ihongda I have tested on my local build and this works.

Ideally this issue would also get fixed: https://github.com/suttacentral/suttacentral/issues/3255 That would allow much better keyboard navigation.

Thanks @khemarato !!

PR #3272, opened 2024-09-12
Only one tab to focus on newly opened panels

### LINKED ISSUE #3255: Opening one of the panels should put the whole panel in focus

khemarato:
Fixes #3255 (and compliments #3271)

Unfortunately, the panel itself isn't focusable. I could focus on the first child within the panel that is focusable, but unfortunely that is sometimes "below the fold", causing the i.e. Info panel to scroll down.

The solution here is to instead immediately focus on the last of the action buttons so that one hit of the tab key will always focus on the next focusable element within the panel.

cc @ihongda and @thesunshade

### DIFF
--- a/client/elements/menus/sc-action-items.js
+++ b/client/elements/menus/sc-action-items.js
@@ -336,12 +336,16 @@ export class SCActionItems extends LitLocalized(LitElement) {
 
   #showSuttaInfo() {
     const isSegmentedText = !this.suttaMetaText;
+    let infoPanel = null;
     if (isSegmentedText) {
-      this.scSiteLayout.querySelector('#bilara-sutta-info').show?.();
+      infoPanel = this.scSiteLayout.querySelector('#bilara-sutta-info');
     } else {
-      this.scSiteLayout.querySelector('#sutta-info').show?.();
+      infoPanel = this.scSiteLayout.querySelector('#sutta-info');
     }
+    infoPanel.show?.();
     this.shadowRoot.querySelector('#btnInfo')?.classList.add(this.activeClass);
+    infoPanel.setAttribute?.("tabindex", "0");
+    infoPanel.focus?.();
   }
 
   #onBtnViewCompactClick(e) {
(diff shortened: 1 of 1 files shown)


thesunshade:
> The solution here is to instead immediately focus on the last of the action buttons so that one hit of the tab key will always focus on the next focusable element within the panel.

So do you mean that if you open the info panel that the info panel icon comes into focus? I think if you do that the user still has to tab through the other icons/buttons to get the whole panel in focus. That's still better than nothing. I don't mean to question you, but you are sure there is no way to focus the whole panel?

And yes, focusing on the first focusable item in the info panel won't work. It should ideally be the whole panel so that the arrow keys can scroll the panel.

Thanks for working on this!

khemarato:
You're welcome to check out the code and fiddle with it 😊  I'm sure there are some tricks I didn't try

thesunshade:
> I'm sure there are some tricks I didn't try

Oh, I'm sure you know far more tricks than I do. In fact on this I am probably totally trickless. In that regard, would trying to explain why it doesn't work be of any use? I realize that you may already have invested as much time on this as you care to.

khemarato:
Ah, okay, I have to add `tabindex="0"` to the panel to make it focusable.  Okay, okay. I can do this the "right" way... if you insist...

thesunshade:
> if you insist...

I didn't want to pressure you, but in the back of my mind I was thinking, "I'm sure he can figure this out. If anyone can it's him."

khemarato:
okay, @thesunshade - How's this? :smile:

thesunshade:
I can't get the iop to work at all now. This is what I have checked out:
image

Also, just clicking on the icons directly doesn't open with the panel focused.

Am I doing something wrong?

khemarato:
> can't get the iop to work at all now.

Of course. These are two different PRs on two different branches.

> just clicking on the icons directly doesn't open with the panel focused.

Are you sure? Use the arrow keys and see what scrolls. Where does a single Tab take you? There won't be a UI indication of what is focused.

thesunshade:
> There won't be a UI indication of what is focused.

Ah, that's what threw me. It does work for info and parallels. However with views, I have to hit tab once to get the whole thing in focus. Then I can use the left/right arrows to pan.

khemarato:
Okay, so for the settings menu there's a `section` inside the toplevel element which does the horizontal scrolling, so for that one we'll have to reach inside its shadow DOM to focus the right thing.  Updated. Thanks for testing it!

thesunshade:
I can confirm this works for me on my local build.

Thanks!

@Hongda this is ready to merge as far as I can tell.

ihongda:
Merged!
Thanks @khemarato @thesunshade  🙏🏼

PR #3274, opened 2024-09-16
Add T Shortcut for toggling PTS Refs

khemarato:
By [popular demand](https://discourse.suttacentral.net/t/experimental-hotkeys/35779/17?u=khemarato.bhikkhu), here is a shortcut key (`t`) for toggling PTS References.

Continues work on #2308.

### DIFF
--- a/client/elements/addons/sc-top-sheet-views.js
+++ b/client/elements/addons/sc-top-sheet-views.js
@@ -731,6 +731,13 @@ export class SCTopSheetViews extends LitLocalized(LitElement) {
         this._showToast(this.localize('viewoption:mainRefsDisabled'));
       }
     }
+    if (selectedReferenceDisplayType === 'pts') {
+      if (checked) {
+        this._showToast(this.localize('viewoption:ptsRefsEnabled'));
+      } else {
+        this._showToast(this.localize('viewoption:ptsRefsDisabled'));
+      }
+    }
     this.actions.setDisplayedReferences(references);
   }
 
--- a/client/elements/addons/sc-top-sheet-views.js
+++ b/client/elements/addons/sc-top-sheet-views.js
@@ -373,6 +373,13 @@ export class SCTopSheetViews extends LitLocalized(LitElement) {
     }
   }
 
+  // T Toggles the "pTs" references
+  _handleTKeydown() {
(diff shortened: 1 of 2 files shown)


khemarato:
Unfortunately, my local Docker got borked with an update (the docker-compose.yml files are incompatible with the new Docker 4, maybe?) so I can't test this change locally :(

@thesunshade - Does this need a toast? Anything else?

thesunshade:
> Does this need a toast?

Unfortunately it does. There are plenty of texts that have no PTS numbers, so there is a good chance that there will be no visual indication of anything changing without a toast.

Are you able to do that? I'm more than happy to test.

BTW, Should I be updating my Docker? I have never updated anything in my Ubuntu setup. Kind of afraid to now.

khemarato:
Yeah, I can get that added. Then please test it.

If it ain't broke... (speaking of which, the CI build here has been broken for a long time...)

khemarato:
Okay. Toast added.  Does my code work on the first try, @thesunshade ? :sweat_smile:

thesunshade:
> Does my code work on the first try,

🥳

First and a half try. I saw your update while I was testing.

PTS toggle with toast works.

ihongda:
Thanks @khemarato @thesunshade 🙏🏼 

The CI build failed because of the docker version upgrade. If the source code is modified to adapt to the new docker version, the server will fail because docker is not updated. I will see how to deal with it later.

khemarato:
@thesunshade - This PR contained a string change  Can you handle the required bilara change? 🙏🏻

thesunshade:
Are you talking about the two new key/value pairs [here](https://github.com/suttacentral/suttacentral/pull/3274/commits/d9ce9f87b5fb3ffc7bfe26d8ff684cceef3629e5)

It might be better for @ihongda to work on since it requires changes on two different `bilara-data` branches and I always seem to mess them up.

khemarato:
> the two new key/value pairs [here](https://github.com/suttacentral/suttacentral/pull/3274/commits/d9ce9f87b5fb3ffc7bfe26d8ff684cceef3629e5)
> 
> It might be better for @ihongda to work on since it

Yes, I mean those two new strings. @ihongda ?? Can you make sure those get added before you push this to staging?

PR #3345, opened 2024-12-28
Navigate word lookup in sc-bottom-sheet.js using arrow keys

transducer:
For me the approach mentioned under help for navigating the bottom sheet using HTML access keys doesn't work. This PR should maintain the original functionality using HTML access keys and also provide functionality to navigate respectively left or right one word on the pressing of ← and →. I think this is more discoverable.

suttacentralarrows
_Goes right one word on pressing right and left one word on pressing left._

updated help display
_Updated help display._

Please let me know if this is a desired implementation or not and if I should add a unit test. And if desired maybe a good idea to also add a keydown handler for <kbd>p</kbd> (for previous word) and <kbd>n</kbd> (for next word)?

Also need to update the [translations files found here](https://github.com/search?q=repo%3Asuttacentral%2Fsc-data+bottomsheet%3Ahelp&type=code).

## Summary by Sourcery

New Features:
- Navigate between words in the bottom sheet using left (<kbd>←</kbd>) and right (<kbd>→</kbd>) arrow keys.

### DIFF
--- a/client/elements/addons/sc-bottom-sheet.js
+++ b/client/elements/addons/sc-bottom-sheet.js
@@ -2,6 +2,7 @@ import { LitElement, html, css } from 'lit';
 import { LitLocalized } from './sc-localization-mixin';
 import { icon } from '../../img/sc-icon';
 import { unsafeHTML } from 'lit/directives/unsafe-html.js';
+import { ignorableKeydownEvent } from '../sc-keyboard-shortcuts';
 
 export class SCBottomSheet extends LitLocalized(LitElement) {
   static properties = {
@@ -408,6 +409,10 @@ export class SCBottomSheet extends LitLocalized(LitElement) {
     this.lookup = {};
     this.defineURL = '';
     this.chinesePunctuation = '，,!！?？;；:：（()）[]【 】。「」﹁﹂"、‧《》〈〉﹏—『』';
+    this._keydownHandlers = {
+      'ArrowLeft': this._handleArrowLeftKeydown.bind(this),
+      'ArrowRight': this._handleArrowRightKeydown.bind(this),
+    };
   }
 
   render() {
@@ -433,12 +438,16 @@ export class SCBottomSheet extends LitLocalized(LitElement) {
                     <ul>
                       <li>
                       ${this.localize('bottomsheet:help5')} =
+                        <kbd>←</kbd>
+                        ${this.localize('bottomsheet:or')}
                         <kbd>Alt</kbd>
                         +
                         <kbd>n</kbd>
                       </li>
                       <li>
                       ${this.localize('bottomsheet:help6')} =
+                        <kbd>→</kbd>
+                        ${this.localize('bottomsheet:or')}
                         <kbd>Alt</kbd>
                         +
                         <kbd>b</kbd>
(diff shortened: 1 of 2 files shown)


thesunshade:
Interesting. Without your PR, alt n does nothing for me on FF and alt b triggers the browser's bookmark interface. So for sure that needs to be fixed.

Generally we seem to avoid having two ways of doing things, so I doubt Bhante @sujato will want to do  arrow keys _and_ alt b/n.

There has also been talk of using the arrow keys for navigating to the previous and next *suttas*. In fact the [browser extension](https://discourse.suttacentral.net/t/a-browser-extension-for-suttacentral/36074) does just that.

Thank you for working on this! Bhante @sujato is traveling atm so it may take a little bit for him to respond.

transducer:
Thank you. I agree a single way of doing things would be best, though in this case some keyboards don't have (easy to use) arrow keys so then it would make sense to support both arrow keys and b and n. Though in this case arrow keys possibly already have the intended purpose of moving between suttas.

For now the access keys don't work then for both of us for some reason. According to my LLM it should be accessKey in React JSX, not lower case. I will try tomorrow and possibly send a different PR. (Update today: this was a hallucination and doesn't matter.) An alternative would be only b or n keydown (without the access key modifier).

I'll wait for reply on how to proceed here.

Op za 28 dec. 2024 19:17 schreef thesunshade ***@***.***>:

thesunshade:
Just as an FYI, here are all the existing hotkeys:
image

transducer:
Thanks for the hot key reference, these are indeed the same as defined in sc-top-sheet-view.js.

I notice now that actually the backward access key (<kbd>Alt</kbd>+<kbd>b</kbd>) does work in Firefox, only the forward (<kbd>Alt</kbd>+<kbd>n</kbd>) does not. The <kbd>Ctrl</kbd>+<kbd>n</kbd> ([in my case the modifier is <kbd>Ctrl</kbd>+<kbd>⌥ Opt</kbd>](https://en.wikipedia.org/wiki/Access_key#Access_in_different_browsers)) I think conflicts in Firefox with other behaviour. In Chrome it actually does work for me.

Since [I don't see access keys used anywhere else in the SuttaCentral applications](https://github.com/search?q=org%3Asuttacentral%20accesskey&type=code), they are quite hard to discover and can conflict with other browser behaviour, maybe best to not use access keys and introduce new regular hot keys for navigating the word lookup? But what keys to use if the arrow keys and <kbd>n</kbd> (toggle notes view) are already taken? Perhaps just like in vi the <kbd>h</kbd> and <kbd>l</kbd> keys for going left and right one word, though this is not so discoverable. Or maybe overload the arrows keys that when the sc-bottom-sheet-view.js is open they are used for navigating the word lookup instead of the suttas? (Though I don't think the latter is technically feasible if the capturing of arrow keys comes from an extension, but if the hotkeys are all defined in the main application it should be possible to arrange it like that.)

jhanarato:
Hey @thesunshade, you commented on this. Do you think it is still relevant or can I close it?

thesunshade:
No idea! Would be good if it could get resolved

transducer:
@thesunshade I changed the keybindings to <kbd>,</kbd> and <kbd>.</kbd>, since they are not taken yet.

thesunshade:
Just a point of consideration… `,` and `.` don't always appear side by side on non-English keyboards.

transducer:
@thesunshade good point. <kbd>b</kbd> and <kbd>n</kbd> was my first thought, but <kbd>n</kbd> is already taken to toggle the notes view. <kbd>[</kbd> <kbd>]</kbd> are not on every keyboard. Maybe <kbd>h</kbd> (left) and <kbd>l</kbd> (right)? They are the vim convention and plain letters on every keyboard. I have changed it to that.

Issue #3369, opened 2025-03-09
Test and refactor data loader


jhanarato:
About two weeks ago the server was having difficulties:

https://discourse.suttacentral.net/t/http-500-error-when-trying-to-read-ven-sujatos-trans-of-mn-137/38285

While the urgent matter is the performance issues, I think we'd do well to refactor first. By refactoring I'm following Martin Fowler's definition:

> Refactoring (noun): a change made to the internal structure of software to make it easier to understand and cheaper to modify without changing its observable behavior.

> Refactoring (verb): to restructure software by applying a series of refactorings without changing its observable behavior.

Automated testing is complementary to refactoring, so as I go along I can create a suite of unit tests. Higher level system tests might need to be in place to get started, otherwise you don't know what you're going to break.

To enable that I have a SuttaCentral instance running in VirtualBox:

https://github.com/jhanarato/sc-testing/blob/main/docs/How-To-Install-Suttacentral-In-VirtualBox.md

So the game plan: 

1. Write some high level system tests and/or characterisation tests to provide some safety.
2. Refactor in small steps, committing regularly.
3. Build a suite of unit tests as the refactoring progresses.
4. Examine the data load process with knowledge gained to see where performance issues might lie.

As it happens, I still seem to have permissions for the suttacentral repository. I can do this work on a branch, or else make a fork. Up to you!

Nice to be back,

Ajahn J.R.

thesunshade:
Glad to have you back!

I know nothing of the work that needs to be done on this count, but I am happy to say Sadhu! and do whatever I can. 

BTW, I don't think Bhante Sujato monitors the issues closely, so if you need his attention best to reach out directly.

jhanarato:
Hey @thesunshade, pretty sure I know who you are on D&D. 

The work is just a series of small improvements, avoiding changes to the behavior. The near future will involve reading the code to figure out what the behavior is, adding tests via `pytest` to avoid introducing bugs.

Back in the day I worked for a library systems company and learned what happens when you refactor without tests. Basically you introduce a series of silly bugs and your coworkers look at you askance. I mentioned this recently to a software developer who stayed with us at Kusala. He said, "Oh yeah, everyone goes through that" Rather comforting.

[...]

PR #3390, opened 2025-03-26
A start on running the data load process in pytest

jhanarato:
I've managed to get the first stage of `make data-load` running in pytest. Running `make test-data-load` runs the test.

There's a test for the whole `arangoload.run()` function that is skipped. It doesn't pass due to a failure in `copy_localization()`. We may need the `sc-frontend` container available for this stage.

Given that when the test passes it will slow down the tests unacceptably, we may want to skip the test module in the `make test-server` target.

## Summary by Sourcery

Introduces a pytest-based test suite for the data load process. Adds a new make target to run the data load tests in a docker container. Implements tests for setting the database name, checking the base directory, and collecting data.

Tests:
- Adds initial tests for the data load process, covering database name setting, base directory verification, and data collection.
- Marks the end-to-end data load test as skipped due to a failure in the copy_localization function.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,6 +62,9 @@ test:
 test-server:
 	@docker exec -t sc-flask pytest -s server/
 
+test-data-load:
+	@docker exec -t sc-flask pytest -v server/data_loader/
+
 test-api:
 	docker compose run --entrypoint "python /opt/sc/api-tester/run-tests.py" sc-api-tester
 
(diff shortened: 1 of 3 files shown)


ihongda:
Bhante, nice work, thank you!

jhanarato:
Thanks!

Just a note that this relates to issue #3369

PR #3425, opened 2025-05-15
Increase test coverage for TextInfoModel 

jhanarato:
More work relating to Issue #3369 

Test coverage for the `TextInfoModel` class is at a point where I'm happy to start refactoring. I've also added some type hints, though there is more to go. There's a small change to the docker configuration to allow access to ArangoDB's web interface when running `make prepare-tests`


ihongda (approved):

ihongda:
Thanks Bhante!

PR #3450, opened 2025-06-02
Testing and refactoring of textdata.py

jhanarato:
I've continued to beef up my tests for the `textdata.py` module and I think they're pretty solid.

Refactoring has finally begun. There are very few changes to behavior, though there are some small changes to logging of missing data in the HTML files.

The old `PaliPageNumbinator` class is finally gone.

As expected, there does not seem to be much of a change in terms of performance as it still does what it did before.

This is continued work for issue #3369


ihongda (approved):
Good code @jhanarato , Thanks!

Issue #3418, opened 2025-05-08
Remove workaround for "font smearing"


jhanarato:
In my investigations into the performance of `make load-data` I came across some odd code:

```py
def is_bold(self, lang, element):
    if element.tag in {'b', 'strong'}:
        return True
    return lang in {'lzh', 'ko', 'jp', 'tw'} and element.tag in {'h1', 'h2', 'h3', 'h4', 'h5', 'h6'}
```

This is used to scan every legacy text and written as a set to the `unicode_points` collection in ArangoDB:

```
unicode_points/en

{
  "normal": "\t\n !\"#%&'()*+,-./0123456789:;=>?ABCDEFGHIJKLMNOPQRSTUVWXYZ[\\]_`abcdefghijklmnopqrstuvwxyz{|}~ §©ÑÜáâæéêíïñóôüĀāăĕĪīŇŋōŚśŪūəɛɦɪʃʌʒˌ̣̄ӕḍḥḷṁṃṅṆṇṚṛṢṣṬṭẓạấ ‍–—―‘’“”…→√《》一三中交佛修典冊出分十卷受員四國基大委子學官年度律戒所提揵教文方日會月有權正毾氀氍法流版研究站經綖綩網翻菩華蔣藏術覴譯金院際電頁鼓ﬁﬂ︰（）",
  "bold": " '(),-.0123456789:?ABCDEFGHIJKLMNOPRSTUWY[]abcdefghijklmnopqrstuvwxyz§ñĀāīūḷṃṇṭ–—’“”…",
  "italic": " !&'(),-.0123456789:;?ABCDEFGHIJKLMNOPQRSTUVWXY[]abcdefghijklmnopqrstuvwxyzñĀāīōŚśū̄ḍḥḷṁṃṅṇṛṣṭ–—‘’“”…"
}
```

That's the first time I've looked at the English record. Looks like it's broken.

I can't see anything that reads from the collection so I contacted Bhante Sujato. His response:

> the reason is probably to avoid using faux bold and italic fonts in certain languages. When a browser smears a bold font it normally looks kind of bad, but it's especially probl;ematic for CJK, where the smear can completely obscure the character. There's supposed to be a CSS property font-synthesis to forbid this, but when we built it support for that was I think FF and Safari only. Just checking now, it was introduced to Chrome in 2022. So our workaround can be deleted, and we can rely on font-synthesis: none to do our work for us!

[...]

PR #3459, opened 2025-06-09
Remove unicode points and refactor unsegmented text titles.

### LINKED ISSUE #3454: Investigate logging of missing titles for unsegmented texts
After refactoring the `load_html_texts` code I get a large number (hundreds?) of errors logged when I run `make load-data`.

`ERROR:data_loader.textdata:Could not find title for text in file: /opt/sc/sc-flask/sc-data/html_text/si/pli/sutta/an/an11/an11.358-405.html`

This occurs only on my `refactor-load-html-texts` branch, not the main branch.

The text itself is here:

https://suttacentral.net/an11.358-405/si/zoysa?lang=en&reference=none&highlight=false

And this is the source HTML file:

https://github.com/suttacentral/sc-data/blob/main/html_text/si/pli/sutta/an/an11/an11.358-405.html

Here's a minimal example for the problem:

```html
<html>
<body>
<header>
<h1>11.358–405</h1>
</header>
</body>
</html>
```

On my refactoring branch, in `UnsegmentedText` we have the following method:

```py
    def title(self) -> str:
        root = self._root
        header = root.select_one('header')
        if not header:
            return ''

        h1 = header.select_one('h1')
        if not h1:
            return ''

        if self._lang_uid == 'lzh':
            left_side = h1.select_one('.mirror-left')
            right_side = h1.select_one('.mirror-right')
            if left_side and right_side:
                return right_side.text_content() + ' (' + left_side.text_content() + ')'

        return regex.sub(r'[\d\.\{\} –-]*', '', h1.text_content(), 1)
```

When the language is not Chinese, and the `<header>` and `<h1>` tags are present, that last line is executed.

It does not use the standard Python `re` library for regular expressions, instead it pulls in this dependency:

https://pypi.org/project/regex/

And a quick test shows that the last line does indeed return an empty string:

```py
    def test_example_of_regex_in_question(self):
        assert regex.sub(r'[\d\.\{\} –-]*', '', '11.358–405', 1) == ''
```

So I've turned up some odd behavior, if not an actual bug. 

How should these cases be handled?

jhanarato:
I've addressed issue #3418 by removing the process that extracts characters and stores them in the `unicode_points` collection. I've left the `ArangoTextInfoModel.update_code_points()` method alone for now as I don't have test coverage for the class. It isn't called though.

On my machine this avoids 10 seconds of CPU time, reducing it from 45 seconds to 35 seconds. This is our first big optimization!

The `UnsegmentedText` class has been refactored to use simple functions and a dataclass containing the extracted details.

I've also fixed #3454 - we no longer log missing titles if they're just empty strings. The output of `make load-data` should no longer print hundreds of errors for this case.

### DIFF
--- a/server/server/data_loader/tests/test_textdata.py
+++ b/server/server/data_loader/tests/test_textdata.py
@@ -255,7 +255,7 @@ class TestTextInfoModel:
         add_html_file(sutta_path, html)
         text_info.process_lang_dir(language_path, base_path, files_to_process)
         assert caplog.records[0].levelno == logging.ERROR
-        assert caplog.records[0].message == f"Could not find title for text in file: {str(sutta_path)}"
+        assert caplog.records[0].message == f"Could not find title in file: {str(sutta_path)}"
 
     def test_logs_missing_title_when_there_is_no_h1_tag(
                 self, text_info, sutta_path, language_path, base_path, files_to_process, caplog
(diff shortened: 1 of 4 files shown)


ihongda (approved):

ihongda:
Thanks @jhanarato 🙏🏼

jhanarato:
Yay!

Issue #3463, opened 2025-06-11
`ChangeTracker` use in `load_html_texts()` forces update of every unsegmented text


jhanarato:
This is the beginning of the `load_html_texts()` function:

```py
def load_html_texts(change_tracker: ChangeTracker, data_dir: Path, db: Database, html_dir: Path):
    print('Loading HTML texts')

    force = change_tracker.is_any_function_changed(
        [textdata.TextInfoModel, textdata.ArangoTextInfoModel]
    )

    if force:
        print('This might take a while')
        db['html_text'].truncate()
```

In my (limited) testing, `force` is always `True`. This means that every time we run this function **every** unsegmented text is updated. By setting `force = False` and running `make load-data` with a populated database we reduce the time taken in `load_html_texts()` from 54 seconds to 0.6 seconds.

jhanarato:
Originally the `ChangeTracker` class just checked file modification times. However, Blake added support for changes to python functions: https://github.com/suttacentral/suttacentral/commit/ce1caa9fa5c21b7086e1ecdf373ef199807ac57b

He updated the functionality later on: https://github.com/suttacentral/suttacentral/commit/f8b8808e8ab0de083454c79efbabe1c68d2eca45


And Ayya Vimala made use of this functionality passing a list of classes: https://github.com/suttacentral/suttacentral/commit/5044e823c834701d7e4789245f536c471c47a26d

I can write some tests to find out why this always indicates a change and the forced update of all html texts, but we really should just remove it as it is unnecessary now.

jhanarato:
I've removed the update forcing and have run a few tests locally.

We rely on the load-data stage to update the last modified time of each file:

```py
printer.print_stage("Updating mtimes")
change_tracker.update_mtimes()
```

Because of this, after this change `make load-data` will still update every unsegmented text the first time it is run. Subsequent runs will skip any unchanged files, which is likely to be everything. Looking at the commits in `sc-data` we typically load a small batch once in a while, which should not take long.

jhanarato:
Running `make run-dev` and `make load-data` skipped unchanged files the first time. Not sure what the behavior will be for the first run in production.

jhanarato:
While investigating #3486, it turns out that the `force` check will be `False` after the 3rd run due to the way `any()` short circuits at the first `True` result.

For this reason it might not have given us such an improvement in performance.

PR #3465, opened 2025-06-12
Refactoring and fix for forced updates

jhanarato:
Sorry to combine these changes. There's quite a bit of refactoring here, but do take a look at the last commit.

There I've removed the forcing of updates as per issue #3463. This is a material change to how loading unsegmented texts works. It should be much faster, less than a second on my machine when there are no updates to process. We do need to check that any new changes to the unsegmented texts are indeed processed.


ihongda (approved):
👍🏼 Thanks Bhante @jhanarato

jhanarato:
Thanks @ihongda for the merge! Let me know how you go on CPU usage.

jhanarato:
This relates to issue #3369

PR #3492, opened 2025-07-29
Fixed title and style issues on publication pages.

ihongda:


client/elements/publication/sc-publication-edition.js
```diff
@@ -211,6 +211,19 @@ export class SCPublicationEdition extends LitLocalized(LitElement) {
...
   }
 
+  _createMetaData() {
+    const description = `${this.editionDetail[0].root_name} — ${this.editionInfo.publication?.creator_name}`;
+    document.dispatchEvent(
+      new CustomEvent('metadata', {
+        detail: {
+          pageTitle: description,
+          title: description,
+          description,
```
agilgur5, line 221: `this.editionDetail[0].blurb` could be used as a proper `description` instead of duplicating the title. I've done that in #3729

client/elements/publication/sc-publication-edition.js
agilgur5, line 1: The `PublicationEditionMatter` pages should also have proper metadata -- I've added similar code in #3729

PR #3505, opened 2025-09-05
Refactored the code for generating publication download links, Adjust the styles of the comments and variants

ihongda:


client/elements/publication/sc-publication-edition.js
```diff
@@ -224,6 +279,17 @@ export class SCPublicationEdition extends LitLocalized(LitElement) {
     );
   }
 
+  async getGitHubDirectory(repoUrl, path = '', branch = 'main') {
```
agilgur5, line 282: this function was added twice - once [on line 182](https://github.com/suttacentral/suttacentral/pull/3505#discussion_r3744228975) and once here. I removed this second one in #3712

client/elements/publication/sc-publication-edition.js
```diff
@@ -176,39 +179,91 @@ export class SCPublicationEdition extends LitLocalized(LitElement) {
...
 
-  #computeFileUrlByType(publicationType) {
-    let discoursesName = this.editionDetail[0].translated_name
-      ?.replace('Collection', '')
-      .trim()
-      .replace(/\s+/g, '-');
-
-    if (this.editionUid === 'pli-tv-vi') {
-      discoursesName = this.editionInfo.publication.translation_title.trim().replace(/\s+/g, '-');
+  async getGitHubDirectory(repoUrl, path = '', branch = 'main') {
```
agilgur5, line 182: this function was added twice - once here and once [on line 282](https://github.com/suttacentral/suttacentral/pull/3505#discussion_r3744227584). I removed the second one in #3712

agilgur5 (commented):
the function `getGitHubDirectory` was duplicated in this PR. I've removed the duplicate in #3712

server/server/api/views/views.py
```diff
@@ -1761,3 +1762,182 @@ def fetch_full_path_by_uid(self, db, uid):
...
+                'size': item.get('size', 0),
+                'download_url': item.get('download_url'),
+                'html_url': item.get('html_url'),
+                'path': item['path']
+            }
+            filtered_contents.append(formatted_item)
+
+        return filtered_contents
+
+    def _get_recursive_contents(self, owner, repo, path, branch, file_extensions):
```
agilgur5, line 1930: the `_get_recursive_contents` method is never used. I've removed it in #3721

server/server/api/views/views.py
```diff
@@ -1761,3 +1762,182 @@ def fetch_full_path_by_uid(self, db, uid):
...
+            parsed = urlparse(url)
+            if parsed.netloc.lower() != 'github.com':
+                return None
+
+            path_parts = parsed.path.strip('/').split('/')
+            return (path_parts[0], path_parts[1]) if len(path_parts) >= 2 else None
+        except:
+            return None
+
+    def _get_directory_contents(self, owner, repo, path, branch, file_extensions):
```
agilgur5, line 1890: the `_get_directory_contents` method is only used in `_get_recursive_contents`, which is never used, per [above comment](https://github.com/suttacentral/suttacentral/pull/3505#discussion_r3812981970).
I've removed them both in #3721

server/server/api/views/views.py
```diff
@@ -1761,3 +1762,182 @@ def fetch_full_path_by_uid(self, db, uid):
...
+            path = request.args.get('path', '')
+            branch = request.args.get('branch', 'main')
+
+            repo_info = self._parse_github_url(repo_url)
+            if not repo_info:
+                return {'error': 'Invalid GitHub repository URL'}, 422
+
+            owner, repo = repo_info
+
+            api_url = f"https://api.github.com/repos/{owner}/{repo}/contents"
```
agilgur5, line 1826: uwahhh, there's no validation on this to check for the `suttacentral/editions` repo only. that means that any attacker or bot could use this SC API to check the contents of any repository, allowing them to hide their tracks, IP address, etc. this is particularly relevant since the API is unauthenticated and allows anyone to use it

I've allowlisted this in #3722 

Since only `suttacentral/editions` is used and allowlisted, the entire `owner` and `repo` parameters could be removed as well. The `branch` is similarly never changed. So it could be a few specific paths instead.
Currently, this is written far too broadly, making it easy to be abused. The opposite should be true, lock it down as much as possible.

But honestly, for its original purpose ([forum thread](https://discourse.suttacentral.net/t/editions-pdf-links-not-working/34034/36?u=agilgur5)), there's no need for an API method at all. I'd recommend either a symlink or pulling the `last_run_date` file -- see the "Future Work" section of #3722

PR #3531, opened 2025-12-16
Add SuttaCentral Express to Related Projects

benmneb:
Adds https://suttacentral.express links to the Related Projects in the site footer and the home page cards.

Note the "real logo" (@thesunshade :eyes:)

Also:

This 03ee3a68ff4c0bfc9f311b8d82aa91292ab462b7 adds styling to account for the overflow on small desk/laptop screens (1121px+ wide) due to another project card, so ie this:

Screenshot 2025-12-16 at 4 27 59 pm

is now this:

Screenshot 2025-12-16 at 4 28 43 pm

As you can see, this adds a new line of cards below 1564px screen width, but fixes all text overflow issues from the header getting squished (these happen between 1121px and 1467px screen widths).

There is no translations because I can't do them without help from a machine...


thesunshade:
se

Here is a higher res image that matches better the others on the page. Only 9kb

benmneb:
not sure why docker 404s in the CI here. is this a problem?

Issue #3536, opened 2025-12-28
Searching in SERP sends two requests


jhanarato:
#### Description:

When submitting a query from the SERP the instant search API receives two requests. This does not happen when the page is loaded, only when enter is pressed or the search icon is clicked. Given that each search can take over 3 seconds to complete this may be causing significant load on the server.

#### Steps to reproduce:

* Open the "Network" tab in Firefox DevTools and filter on POST.
* Type a search into the Algolia based search to bring up the SERP.
* The network activity now shows one request to `/api/search/instant`
* Assuming the search field is filled, hit enter.
* The network activity now shows three requests, all the same.
* With the same search query, click the magnifying icon.
* Another two identical requests are shown.

#### Expected behaviour:

As the requests are identical, there should be one request in each case:

* We submit a search request via the Algolia search bar.
* On the SERP we press enter.
* On the SERP we click the magnifying glass.

#### Investigation

We have two components that interact:

* `sc-page-search.js`
* `sc-auto-complete-list.js`

In page search we have a component `SCPageSearch`. The input field is defined as:

```js
  get searchResultHeadTemplate() {
    return html`
      <md-filled-text-field
        id="search_input"
        type="search"
        label="${this.localize('search:inputSearchTerm')}"
        minlength="2"
        maxlength="100"
        @keypress=${this.#keypressHandler}
      >
        <md-icon slot="trailing-icon" @click=${this._startSearch}> ${icon.search} </md-icon>
      </md-filled-text-field>
      ${
        extractSelectedLangsName(
          store.getState().searchOptions.displayedLanguages,
          this.localize('autocomplete:change'),
          true
        )
      }
      ${this.#searchOptionsTemplate()}
    `;
  }
```

Note the keypress and click are hooked up asynchronously here. These seem to work ok.

`SCPageSearch` also defines this callback:

```js
  shouldUpdate(prevProps) {
    if (prevProps.has('searchQuery')) {
      this.#startNewSearchWithNewWord();
    }
    if (prevProps.has('lastSearchResults')) {
      this.#populateList();
    }
    return true;
  }
```

Really, `shouldUpdate()` is intended to inform the component whether to update. Here we are hitting the search API which is not a good idea. 

An overview of the Lit lifecycle can be found here:

https://lit.dev/docs/components/lifecycle/

It looks like `shouldUpdate` is used this way elsewhere in the code and appeared before @ihongda took over. 

[...]

PR #3549, opened 2026-01-16
Fix issue 3536

jhanarato:
The double-search issue #3536 appears to be fixed. I've extracted the code for fetching search results into a separate controller and tested it as well as I can. We only fetch a new response from the search API if the URL or body of the request change. This should really be handled via the Lit update lifecycle but that will be quite a bit of work if we take it on. For now we seem to be good.

It all seems worthwhile as it will greatly reduce the strain on the server.


client/elements/sc-page-search.js
```diff
@@ -11,14 +11,15 @@ import './addons/sc-error-icon';
 import './addons/sc-progress';
 import { icon } from '../img/sc-icon';
 import { store } from '../redux-store';
-import { LitLocalized } from './addons/sc-localization-mixin';
+import { LitLocalized} from './addons/sc-localization-mixin';
```
khemarato, line 14: Should probably keep the style consistent with the existing imports

client/elements/sc-page-search.js
```diff
@@ -11,14 +11,15 @@ import './addons/sc-error-icon';
...
 import { API_ROOT } from '../constants';
 import { dictionarySimpleItemToHtml } from './sc-dictionary-common';
 import { SCPageSearchStyles, searchResultTableViewStyles } from './styles/sc-page-search-styles';
 import { SCUtilityStyles } from './styles/sc-utility-styles';
 import { dispatchCustomEvent } from '../utils/customEvent';
 import { reduxActions } from './addons/sc-redux-actions';
 import { extractSelectedLangsName } from './addons/sc-functions-miscellaneous';
+import {SCSearchController} from "./sc-search-controller";
```
khemarato, line 22: ditto

client/elements/sc-search-controller.js
```diff
@@ -0,0 +1,72 @@
+// The SCSearchController class and it's helpers only exist because
```
khemarato, line 1: its*

client/elements/sc-page-search.js
agilgur5: I'd recommend just running `prettier` in dev and CI so style is automatically consistent

client/elements/sc-page-search.js
agilgur5: also the file path is inconsistently double-quoted here too (on top of the brace spacing) -- `prettier` can handle both of those

client/elements/sc-page-search.js
agilgur5: - [ ] I see `prettier` was already added to the toolchain and pre-commit in https://github.com/suttacentral/suttacentral/pull/1846, but it hasn't been added [to CI](https://github.com/suttacentral/suttacentral/blob/63434a1db6acd4206654bdd7df97269007821c88/.github/workflows/deploy.yml#L13) (which also [failed on this PR](https://github.com/suttacentral/suttacentral/actions/runs/21067095523/job/60587002740) but was still merged -- this can be set to a requirement for merge in the GH settings).
  - **EDIT**: WIP in #3572 et al
- [x] I also see the `make test` step was commented out in CI in e79e98fbfafb3cfdeed8a9971022ae705e18d700 (https://github.com/suttacentral/suttacentral/pull/3396) but was never added back in.
    - **EDIT**: ✅ this is being added back as part of #3613 / #3617 
- [ ] `make test` only runs server tests [currently](https://github.com/suttacentral/suttacentral/blob/63434a1db6acd4206654bdd7df97269007821c88/Makefile#L59)
- [x] a separate `make lint` and `make lint:fix` (auto-format) tasks could be good to have as well
    - [x] `ruff`, `flake8`, `pylint`, `black`, `isort` as some possible tools for the Python codebase
      - since we're using `uv` in #3557, would make sense to use [`ruff`](https://github.com/astral-sh/ruff), backed by the same org
    - **EDIT**: ✅ `ruff` and corresponding `make` targets set-up in #3656 
    
Wrote those up as 'to-dos' above as I think they'd be quite helpful for consistency and stability and I may well tackle those

Issue #3556, opened 2026-02-09
Volume mounting for `sc-flask` can be replaced by Dockerfile script


jhanarato:
Having spent some time configuring a `uv` environment for the `sc-flask` docker image, I've come across some configuration problems. The `Dockerfile-flask` file sets things up for us and creates the directory `/opt/sc/sc-flask` in our docker image. Our `Makefile` build process then introduces several docker compose files:

- `docker-compose.yml`
- `docker-compose.dev.yml`
- `docker-compose.test.yml`
- `docker-compose.prod.yml`

If we run, say, `make run-dev` we use `docker-compose.yml` and `docker-compose.dev.yml`.

These configure the `sc-flask` service like so:

### `docker-compose.yml`

```yml
services:
  sc-flask:
    build:
      context: ./server/
      dockerfile: docker/Dockerfile-flask
    container_name: 'sc-flask'
    labels:
      application: 'sc-flask'
    env_file:
      - server/env/.base.env
    environment:
      - FLASK_APP=server/app.py
    volumes:
      - socket-volume:/opt/sc/sockets
      - ./server/:/opt/sc/sc-flask
      - working-storage-volume:/opt/sc/storage
      - sc-data-volume:/opt/sc/sc-flask/sc-data
    expose:
      - '5000'
    depends_on:
      - sc-arangodb
```

### `docker-compose.dev.yml`

```yml
services:
  sc-flask:
    command: bash docker/entrypoint.dev.sh
    env_file:
      - server/env/.dev.env
    volumes:
      - ./server/:/opt/sc/sc-flask
      - ./client/:/opt/sc/frontend
    ports:
      - '5001:5000'
```

### `docker-compose.test.yml`

```yml
services:
  sc-flask:
    volumes:
      - ./server/:/opt/sc/sc-flask
      - ./client/:/opt/sc/frontend
    env_file:
      - server/env/.test.env
    command: bash docker/entrypoint.test.sh
```

### `docker-compose.prod.yml`

```yml
services:
  sc-flask:
    restart: always
    command: bash docker/entrypoint.prod.sh
    env_file:
      - server/env/.prod.env
    volumes:
      - ./server/:/opt/sc/sc-flask
      - ./client/:/opt/sc/frontend
    ports:
      - '5001:5000'
```

In each case we mount a volume like so: `./server/:/opt/sc/sc-flask`

So whatever we created in `/opt/sc/sc-flask` in our Dockerfile is now replaced by the new volume.

[...]

Issue #809, opened 2018-03-26
Documentation: Sphinx


SamaneriVimala:
Original ticket Creator:
malgorzata.maksimczyk

*Why*

to support future development and understanding of what was coded so far
*AC*# Documentation: Sphinx
*Technical subtasks (implementation details)*# Core task as in AC

Jira ID: SUCE-17
Jira Creation Date: 23/Aug/17 1:52 PM
Jira Epic: 
Jira Priority: Medium
Jira Status (at import): To Do

Comments:
 
14/Sep/17 2:28 PM
malgorzata.maksimczyk
reestimate 5->2
13/Feb/18 12:24 PM
malgorzata.maksimczyk
read me should be enough. read me is part of go live activities

PR #3557, opened 2026-02-09
Use uv

jhanarato:
I've added support for dependency management with `uv`. If we now rebuild the `sc-flask` image a virtual environment is created and added to the path. 

This is a fairly major change to how we do things so please have a good look at this before deploying in production.


jhanarato:
Also, before deploying you'll want to make sure that the `sc-flask` image is fully rebuilt. Perhaps run 

`docker compose down --rmi all sc-flask` 

Hope I've got that right.

server/docker/Dockerfile-flask
```diff
@@ -1,8 +1,20 @@
...
-COPY requirements.txt ./
-RUN pip install -r requirements.txt
-COPY *.* ./
+
+# Create a new virtual environment with uv
+RUN mkdir -p /opt/sc/venv
+WORKDIR /opt/sc/venv
+COPY pyproject.toml ./
+COPY uv.lock ./
+COPY .python-version ./
```
agilgur5, line 14: There's a shorthand for [copying to the same directory](https://docs.docker.com/reference/dockerfile/#source-1):

```suggestion
COPY pyproject.toml uv.lock .python-version ./
```

server/docker/Dockerfile-flask
```diff
@@ -1,8 +1,20 @@
...
+RUN uv sync --locked
```
agilgur5, line 15: You'll want to change the `Makefile` as well:
https://github.com/suttacentral/suttacentral/blob/63434a1db6acd4206654bdd7df97269007821c88/Makefile#L48

and should remove the `requirements.txt` in this PR or a follow-up

server/docker/Dockerfile-flask
```diff
@@ -1,8 +1,20 @@
...
+ENV PATH="/opt/sc/venv/.venv/bin:$PATH"
+
+# The application source code will be here when
+# we mount it in the docker compose files.
```
agilgur5, line 19: This comment and removing the `COPY *.* ./` seems to contradict your plan in #3556. Not sure if this was a prior iteration before you had opted for removing volume mounts?

agilgur5 (commented):
This is just a very quick initial look through (it's ~2am here and I am a bit exhausted after a Super Bowl party 😅) -- I'll take a deeper dive later this week, but I left a few small suggestions below

server/docker/Dockerfile-flask
jhanarato: Oh right, missed that one. 

Also, the `api-tester` container depends on the same `requirements.txt` file so I've left it there for now.

server/docker/Dockerfile-flask
jhanarato: That's right. When we remove the volume mounts we'll update the `Dockerfile` to copy exactly what we need in the image. The comment is there to describe the current state of affairs.

server/docker/Dockerfile-flask
agilgur5: I was gonna mention that you could put [a comment](https://pip.pypa.io/en/latest/reference/requirements-file-format/) `# TODO: remove this once uv migration is done` at the top of the `requirements.txt` to indicate this in the interim, but it's already removed in #3562 (https://github.com/suttacentral/suttacentral/issues/3561#issuecomment-3896316919) so you got to that before I could even mention it!

✅  Can mark this as "resolved" now (there should be a button in the top right)

server/docker/Dockerfile-flask
agilgur5: ✅  Completed in #3562, can resolve this thread now

server/docker/Dockerfile-flask
agilgur5: > Also, the `api-tester` container depends on the same `requirements.txt` file so I've left it there for now.

`api-tester` seems to have [its own separate `requirements.txt`](https://github.com/suttacentral/suttacentral/blob/6cb1b4e269b264bce85ef532ae672215ee4c1e08/server/api-tester/requirements.txt), no? ❓ 

**EDIT**: `api-tester` is being followed up on in #3573

server/docker/Dockerfile-flask
agilgur5: ✅  This has been updated in #3562 so is outdated now and can be resolved.

That kind of makes sense for a lack of `COPY *.* ./` in this PR, although I'd be concerned if prod depended on that line/those files existing in the container (idk). This is technically a "breaking change" to the container image as such and so I would have left it as-is to make sure nothing breaks in the interim between the two PRs.
But nbd now and didn't seem to break anything! 🙏

server/docker/Dockerfile-flask
```diff
@@ -1,8 +1,20 @@
...
-COPY *.* ./
+
+# Create a new virtual environment with uv
+RUN mkdir -p /opt/sc/venv
+WORKDIR /opt/sc/venv
+COPY pyproject.toml ./
+COPY uv.lock ./
+COPY .python-version ./
+RUN uv sync --locked
+ENV PATH="/opt/sc/venv/.venv/bin:$PATH"
```
agilgur5, line 16: ✅  This and the `RUN mkdir -p /opt/sc/venv` lines et al above was the main last bit I hadn't quite gone through as I didn't think it should be necessary (`uv` should do that automatically), but wasn't 100% sure.
It looks like you've removed these lines in #3562 so indeed they were not necessary 👍

server/pyproject.toml
```diff
@@ -0,0 +1,98 @@
...
+    "pyenchant==3.2.2",
+    "pygments==2.13.0",
+    "pyjwt==2.6.0",
+    "pykakasi==2.2.1",
+    "pyparsing==3.0.9",
+    "pyppeteer==1.0.2",
+    "pyrsistent==0.19.2",
+    "pytest==7.2.0",
+    "pytest-flask==1.2.0",
+    "pytest-sugar==0.9.6",
```
agilgur5, line 61: - [x] Some of these (I highlighted a few, there might be more) should be moved to [dev dependencies](https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies) now that we're using a package manager that supports the separation
  - **EDIT**: ✅ Moved dev dependencies into a separate group in #3596

server/pyproject.toml
```diff
@@ -0,0 +1,98 @@
...
+    "sortedcontainers==2.4.0",
+    "soupsieve==2.3.1",
+    "sphinx==5.3.0",
+    "sphinxcontrib-applehelp==1.0.2",
+    "sphinxcontrib-devhelp==1.0.2",
+    "sphinxcontrib-htmlhelp==2.0.0",
+    "sphinxcontrib-jsmath==1.0.1",
+    "sphinxcontrib-qthelp==1.0.3",
+    "sphinxcontrib-serializinghtml==1.1.5",
+    "sphinxcontrib-websupport==1.2.4",
```
agilgur5, line 82: - [x] I vaguely mentioned the Spinx deps in https://github.com/suttacentral/bilara-data/pull/4739#issuecomment-3694488270, but I'm pretty sure they're [no longer used](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+sphinx&type=code) or were never fully implemented from https://github.com/suttacentral/suttacentral/issues/809.

    - [x] And I'd suggest having dev docs as plain markdown ([as I described in dev chat before](https://discourse.suttacentral.net/chat/c/-/12/5193)) within relevant directories anyway for ease of use and access.
        - **EDIT**: ✅ dev docs in directories were started as part of https://github.com/suttacentral/suttacentral/pull/3575 with a [`server/README.md`](https://github.com/suttacentral/suttacentral/blob/652f7a0b45fb87cfafdf80f6417901c6c0c0aa29/server/README.md)
    
    If a separate docs site were used for dev docs, I'd still suggest splitting out these deps as either dev deps or into a separate `pyproject.toml`.

    But right now it seems like they can just be removed. I removed them in a branch of mine but I hadn't gotten to testing that it didn't break anything - i.e. if Sphinx is a sub-dependency of something or not. Although with the `uv.lock` that should be easier to detect though!
    
    - **EDIT**: ✅ Sphinx deps removed in #3578 

- [x] And there might be other unused deps too 😅 
  - **EDIT**: ✅ Rest of unused deps removed in #3587, #3588, #3591

server/pyproject.toml
```diff
@@ -0,0 +1,98 @@
+[project]
+name = "sc-server"
+version = "0.1.0"
+description = "The backend python implementation for SuttaCentral.net"
+readme = "README.md"
+requires-python = "==3.11.3"
```
agilgur5, line 6:
```suggestion
requires-python = ">=3.11"
```

I'd suggest allowing any newer Python too (see also https://github.com/astral-sh/uv/issues/7429#issuecomment-2353092246), though nbd really since this is primarily internal code and not a published package

agilgur5 (commented):
Just added some follow-up comments: 
- mostly been resolved in #3562 already
- and jotted down some TODOs both related and unrelated to `uv`

server/docker/Dockerfile-flask
agilgur5: It seems to have been added back in in https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2870800331 though?
**EDIT**: see the comments there for follow-up

server/pyproject.toml
agilgur5: ✅ Moved dev dependencies into a separate group in #3596

server/pyproject.toml
agilgur5: - ✅ Sphinx deps removed in #3578 
- ✅ other unused deps removed in #3587, #3588, #3591 
- ✅ dev docs in relevant dirs started as part of https://github.com/suttacentral/suttacentral/pull/3575 with a [`server/README.md`](https://github.com/suttacentral/suttacentral/blob/652f7a0b45fb87cfafdf80f6417901c6c0c0aa29/server/README.md)

server/pyproject.toml
jhanarato: I've just pinned all the dependencies to avoid breaking something. But yes, we should at least use the latest patch level for Python 3.11

agilgur5: For the Python dependencies (outside of Python itself), we can now switch to ranges because the lockfile keeps things pinned. SemVer ranges should allow us to auto-update some deps when necessary (e.g. security patches) while keeping them all compatible with each other (which is otherwise much harder when there are no ranges)

jhanarato: Seems like a good idea to me.

agilgur5: #3589 will already make this easier, and then once all my Python dep clean-up PRs have been merged, I can switch the remaining ones to at least accept patch versions, if not minor bumps too

agilgur5: ✅ Using SemVer specifiers for all deps in https://github.com/suttacentral/suttacentral/pull/3661

Issue #3561, opened 2026-02-13
Tidy build configuration for sc-flask


jhanarato:
While reviewing the docker configuration I've noted the following:

### We mount the `.server/` directory from the host machine

This is covered by issue #3556 

### There are three entrypoint scripts

- `entrypoint.dev.sh`
- `entrypoint.test.sh`
- `entrypoint.prod.sh`

Apart from a blank line, these are identical. We should have a single entrypoint script which can live in the root of the build context. 

### Entrypoint is specified in docker compose

Each of the three docker compose files for `dev`, `test` and `prod` specify a command for the sc-flask service. e.g. `docker-compose.dev.yml`:

```yml
services:
  sc-flask:
    command: bash docker/entrypoint.dev.sh
    env_file:
      - server/env/.dev.env
    volumes:
      - ./server/:/opt/sc/sc-flask
      - ./client/:/opt/sc/frontend
    ports:
      - '5001:5000'
```

With a single entrypoint script we can remove this from docker compose and add it to the `Dockerfile` as a `CMD` or `ENDPOINT`.

### `docker` directory not pulling it's weight

If we move the new entrypoint script into the build context we can do the same for `Dockerfile-flask` and rename it to `Dockerfile`, then remove the `docker` directory.

### `pytest.ini` is empty

Delete it.

jhanarato:
We also need to remove the `requirements.txt` file and `install-requirements` in the Makefile

jhanarato:
I'm not sure, but we might be able to remove the `FLASK_APP` environment variable in `docker-compose.yml`:

```yml
services:
  sc-flask:
    build:
      context: ./server/
      dockerfile: docker/Dockerfile-flask
    container_name: 'sc-flask'
    labels:
      application: 'sc-flask'
    env_file:
      - server/env/.base.env
    environment:
      - FLASK_APP=server/app.py
```

jhanarato:
We should also remove the version attribute:

```
WARN[0000] /home/jr/Code/suttacentral/docker-compose.yml: the attribute `version` is obsolete, it will be ignored, please remove it to avoid potential confusion 
```

jhanarato:
And fix this:

```
WARN[0000] networks.default: external.name is deprecated. Please set name and external: true 
```

agilgur5:
Pretty much agree 100% with the recommendations. 👍 Some of them I was planning on doing myself too.

[...]

PR #3562, opened 2026-02-16
Tidy build configuration

### LINKED ISSUE #3556: Volume mounting for `sc-flask` can be replaced by Dockerfile script

jhanarato:
This PR covers issues #3556 and #3561.

I've not closed #3556 yet, I still have work to do to setup "watch" with docker compose:

https://docs.docker.com/compose/how-tos/file-watch/

That's for the next PR.

I've removed our `.venv` from the PATH environment variable I added to work with uv. Now we can rely on it looking for the `.venv` in the current or nearest parent directory when we use `uv run`.

This work is a stepping stone. Now we can turn our application code into an installable package.


server/server/pytest.ini
agilgur5, line 1: I might recommend a [`.gitattributes` file](https://docs.github.com/en/get-started/git-basics/configuring-git-to-handle-line-endings) to standardize line endings between OSes. The diff here says "Whitespace-only changes", which, given this file is empty, might be because of that? (although could be something else that GH is not displaying well)

Probably a one-liner would suffice:
```
# Set the default behavior, in case people don't have core.autocrlf set.
* text=auto
```

server/Dockerfile-flask
```diff
@@ -0,0 +1,19 @@
+FROM python:3.11.3
+COPY --from=ghcr.io/astral-sh/uv:0.9.30 /uv /uvx /bin/
+
+RUN apt-get update && apt-get install -y pngnq && apt-get install -y python3-enchant
```
agilgur5, line 4: You didn't change this line, but there is a short-hand for installing multiple packages as well:

```suggestion
RUN apt-get update && apt-get install -y pngnq python3-enchant
```

agilgur5: Also it looks like `python3-enchant` / [`pyenchant`](https://github.com/pyenchant/pyenchant) / `enchant` is [no longer used](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+enchant&type=code) (was previously introduced in https://github.com/suttacentral/suttacentral/pull/2636)

Not sure if [`pngnq`](https://github.com/stuart/pngnq) is still used or a sub-dependency of something or not

- [x] Would make a separate PR to remove one or both
  - **EDIT**: ✅ Removed `pyenchant` in #3588 
  - **EDIT**: ✅ `pngnq` is removed in https://github.com/suttacentral/suttacentral/pull/3585#discussion_r2969404316

api-tester/Dockerfile-api-tester
agilgur5, line 1: IMO, I'd leave `api-tester` inside of `server` since it's a E2E tester _for_ the `server` if I'm understanding correctly.

server/Dockerfile-flask
```diff
@@ -0,0 +1,19 @@
...
+
+WORKDIR /opt/sc/sc-flask
+
+# Setup virtual environment with uv
+COPY pyproject.toml uv.lock .python-version ./
+RUN uv sync --locked
+
+# Copy over remaining project files.
+COPY entrypoint.sh uwsgi.ini ./
+COPY server/ ./server
```
agilgur5, line 17: Largely orthogonal to this PR, but it might make sense to rename the inner directory to `src/`, which is a common name and helps alleviate the current confusion of same named sub-directories `server/server/` 😅 (to `server/src/` instead)

agilgur5 (commented):
A few tiny comments below (the `api-tester` move being the main one), but this LGTM!

Nice work! 🙂

server/Dockerfile-flask
agilgur5: ✅ Oh I just saw #3564 which covers this

jhanarato: Yep, that's done and cued up on my local repository

server/Dockerfile-flask
jhanarato: Sure, we can take a look at that later

api-tester/Dockerfile-api-tester
jhanarato: Addressed in issue #3573

server/Dockerfile-flask
jhanarato: I'm not sure myself, so I didn't remove it. 

The way we're going, I'm not sure we need a full-fat debian install for `sc-flask`

server/server/pytest.ini
agilgur5: ✅  a `.gitattributes` file was verbatim added in https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2904365272 / https://github.com/suttacentral/suttacentral/commit/196c80725db812a87723b0374b0868dee4b04d20

api-tester/Dockerfile-api-tester
agilgur5: ✅  Moved back in https://github.com/suttacentral/suttacentral/pull/3575/commits/31d3ddaf5cb4c6b1ffff47dff6fbb09661490fb1

server/Dockerfile-flask
agilgur5: ✅ Removed `pyenchant` in #3588 

I'll leave `pngnq` to you to remove!
- **EDIT**: ✅ it was removed in https://github.com/suttacentral/suttacentral/pull/3585#discussion_r2969404316

PR #3565, opened 2026-02-22
fix: correct assignment of viewCompose in SCTextBilara class

ihongda:


agilgur5:
Nice catch!

For reference, this fixes a bug [discovered by](https://discourse.suttacentral.net/t/suttacentral-bug-reports/27763/291?u=agilgur5) @sabbamitta

client/elements/text/sc-text-bilara.js
```diff
@@ -787,7 +787,7 @@ export class SCTextBilara extends SCTextCommon {
     this.viewCompose = `${this.chosenNoteDisplayType}_${this.chosenTextView}`;
     if (!this.bilaraTranslatedSutta && this.bilaraRootSutta) {
       if (this.chosenNoteDisplayType === 'sidenotes') {
-        viewCompose = 'sidenotes_root';
```
agilgur5, line 790: This should be caught by a linter as assigning an undeclared variable 🤔 

- [x] Similar to https://github.com/suttacentral/suttacentral/pull/3549#discussion_r2778027843, it looks like `eslint` is not being run. I started fixing this up and realized there isn't even an ESLint configuration (`.eslintrc` / `eslint.config.js`), despite all the [ESLint dependencies and plugins](https://github.com/suttacentral/suttacentral/blob/9b3974700379052f18d60af1ca3b3bc6f013d132/client/package.json#L77). I'll work on that right now
  - ✅ ESLint is configured in #3572

PR #3568, opened 2026-02-26
clean(make): remove outdated `extract-static-strings` command

agilgur5:
## Summary

Remove outdated `make rebuild-static-pages` command that uses the no longer existent `npm run extract-static-strings` script.

## Details

- it was removed from the `package.json` in 3182f7e50d05331b3b7cf884d6d55d022fba7fac (https://github.com/suttacentral/suttacentral/pull/2035), so `make rebuild-static-pages` will always error and doesn't otherwise do anything
  - this `make` command is [the only remaining place](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral%20extract-static-strings&type=code) that uses it
  
## Verification
- Found to be unused by search and historical analysis (see above)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -100,9 +100,6 @@ rebuild-frontend:
 bundle-analyzer:
 	docker compose run sc-frontend npm run build --report
 
-rebuild-static-pages:
-	cd client && npm run extract-static-strings
-
 run-preview-env:
 	@make clean-all
 	@make create-network


jhanarato:
As I understand it, the script was written directly in `client/package.json` in commit 54dd265c6d13a7b56d1b169e28517812572c22a6:

```json
"extract-static-strings": "babel --plugins babel-plugin-macros elements/static-templates/ --out-dir elements/static"
```

The script was removed from `client/package.json` in 3182f7e50d05331b3b7cf884d6d55d022fba7fac but not the `Makefile` which still refers to the script.

Trying to run the `Makefile` command fails:

```commandline
❯ make rebuild-static-pages 
cd client && npm run extract-static-strings
npm error Missing script: "extract-static-strings"
npm error
npm error To see a list of scripts, run:
npm error   npm run
npm error A complete log of this run can be found in: /home/jr/.npm/_logs/2026-04-02T04_48_12_474Z-debug-0.log
make: *** [Makefile:105: rebuild-static-pages] Error 1
```

So the command is broken and should be removed.

Makefile
```diff
@@ -100,9 +100,6 @@ rebuild-frontend:
 bundle-analyzer:
 	docker compose run sc-frontend npm run build --report
 
-rebuild-static-pages:
```
jhanarato, line 103: This command is broken and should be removed.

jhanarato (approved):
The command is broken and should be removed.

PR #3570, opened 2026-02-26
clean(deps): remove unused Babel syntax proposal plugins

agilgur5:
## Summary

Remove unused Babel syntax proposal plugins in the `client`, their deps and config

## Details

- these syntaxes are not used in the codebase:
  - `do-expressions`: [`do {`](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+%22do+%7B%22&type=code) (and `async do {`)
  - `pipeline-operator`: `|>`
    - the only time these characters appears in the codebase is in [a regex](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+%7C%3E&type=code), which is unrelated
  - `function-bind`: `::<function call>`
    - all the [`::` usage in the codebase](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+%3A%3A&type=code&p=1) is for CSS pseudo-selectors
  - `throw-expressions`: `throw` in the middle of an expression
    - `throw` is only used as its own expression [in the codebase](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+throw+language%3AJavaScript&type=code&l=JavaScript)
  - `decorators`: `@<function>`
    - [`@`](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+%40+language%3AJavaScript&type=code&l=JavaScript) is only used in CSS media queries (`@media`) and animations (`@keyframes`), Lit event listener bindings (e.g. `@click=`), package names (e.g. `@babel/`), regex, and some occassional JSDoc comments
 - they appear to have been added in https://github.com/suttacentral/suttacentral/pull/2161, but I guess never used?
    
 - remove the devDeps in `package.json` and `package-lock.json`
 - remove them from the Vite and Webpack Babel configs
 
 ## Verification
 
 - Found to be unused via searches (see above)
 - `npm test` still passes
 - `npm run lint` (#3572) has no syntax errors without these plugins
 
 ## Future Work
 
 - [ ] The Vite, Webpack, and ESLint Babel configs can all probably be unified as a single `babel.config.js` used by all 3, unless there are necessary differences between them
[...]

PR #3571, opened 2026-02-27
clean(deps): remove redundant Babel transforms

agilgur5:
## Summary

Remove a few Babel transforms from the client that are redundant as they are already part of `preset-env`, their deps and config

This resolves two of the "Future Work" todos I left in #3570.
~It also may well merge conflict with that PR because the lines that change are adjacent, so merge that one first please and then I can update this one. Marking this as draft for now as such~
- **EDIT**: It did indeed conflict. I've now rebased on top and fixed those, so this is now ready

## Details

- [`@babel/plugin-transform-class-properties`](https://babeljs.io/docs/babel-plugin-transform-class-properties), [`@babel/plugin-transform-private-methods`](https://babeljs.io/docs/babel-plugin-transform-private-methods), and [`@babel/plugin-transform-modules-commonjs`](https://babeljs.io/docs/babel-plugin-transform-modules-commonjs) are all already included in `@babel/preset-env` as of ES2022 and so do not need to be explicitly installed
  - you can also see they're all in `@babel/preset-env`'s dependencies [in](https://github.com/suttacentral/suttacentral/blob/9b3974700379052f18d60af1ca3b3bc6f013d132/client/package-lock.json#L2043) [the](https://github.com/suttacentral/suttacentral/blob/9b3974700379052f18d60af1ca3b3bc6f013d132/client/package-lock.json#L2072) [`package-lock.json`](https://github.com/suttacentral/suttacentral/blob/9b3974700379052f18d60af1ca3b3bc6f013d132/client/package-lock.json#L2060)
    - Note that while the diff has some removals in `package-lock.json`, that's because there are actually two different versions in the deps due to different version pins [in the `package.json`](https://github.com/suttacentral/suttacentral/blob/f1a4c0e1045693907b917e3725ad6e397b294a34/client/package.json#L56). So this removes one of the two

- additionally, `@babel/plugin-transform-modules-commonjs` is independently [unused](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+transform-modules-commonjs&type=code) in the codebase as it's not configured anywhere

## Verification

- Found to be redundant as per search and docs (see above)
- `npm test` still passes
- `npm run lint` (#3572) has no new syntax errors without these plugins


jhanarato:
Nice work!

When @ihongda has checked #3570 we can look at this.

agilgur5:
> It also may well merge conflict with that PR because the lines that change are adjacent, so merge that one first please and then I can update this one. Marking this as draft for now as such

> When @ihongda has checked #3570 we can look at this.

I've rebased on top and fixed the small conflicts. Marked this as ready for review now

ihongda:
Thanks @agilgur5 ❤️

Issue #3564, opened 2026-02-18
Restructure `sc-flask` Python code for packaging.


jhanarato:
### Existing setup

The `sc-flask` docker container contains a number of Python files under `suttacentral/server/server/`. These are called directly by the `Makefile` and `uswgi` command.

### Create an installable package

We can turn this Python code into a wheel build and install it in the container.

This will likely involve the following steps:

- Add the required `[build-system]` and `[tool.uv.build-backend]` sections to the `pyproject.toml` file.
- Add a README.md file. This is usually required by the build process and should contain details for working with the Python source code inside the container and within the developer's native environment.
- Update the `uwsgi` configuration to use the installed `server` package.
- Update the `Makefile` so that all commands use installed `server` package.
- Move the `server` package into a `src` folder.
- Move the test code out of the `src` folder into `test`.
- Update `Makefile` to use the `test` folder.

jhanarato:
During my investigations yesterday I discovered our packages are within the `server/server/` folder. There is some odd stuff happening with paths, in particular our `uswgi` configuration point to the module `server.wsgi:app` but then the `wsgi.py` script does this:

```py
import os
import sys

dir_path = os.path.dirname(os.path.realpath(__file__))
sys.path.append(dir_path)

from app import app  # isort:skip


if __name__ == '__main__':
    app.run(debug=True)
```

If we want to install the SC packages in the `.venv` we need to treat the directories `api`, `common`, etc as packages and the remaining files, `app.py` etc as scripts.

jhanarato:
Also note that `conftest.py` contains something similar:

```py
myPath = os.path.dirname(os.path.abspath(__file__))
sys.path.append(myPath)
```

agilgur5:
> * Move the `server` package into a `src` folder.

> within the `server/server/` folder

Oh just saw this issue -- I suggested the same `server/src/` structure in https://github.com/suttacentral/suttacentral/pull/3562#discussion_r2839235161 too

agilgur5:
I think most of this is good to do in any case, but this one:

> Create an installable package

is not strictly necessary. Since there is already a `pyproject.toml`, it is already understood as an application (which it is). There's no need to make it a library, unless you wanted to install it elsewhere. I actually wouldn't recommend it unless necessary, as it creates some complications (e.g. needing to use an editable install during dev etc).

[...]

PR #3574, opened 2026-02-27
Issue 3564

jhanarato:
This is my first attempt at a stacked pull request. Pull request #3562 should be merged before this one.

Most of the changes here are files moving from the `server/server` directory into `server/src` and `server/tests`. A number of Python files have had their imports modified, perhaps more than necessary as PyCharm has helpfully removed unused imports as part of the refactoring.

Our `Makefile` retains the same targets with changes to the locations of the tests run by pytest and running the Flask manage commands via the installed project rather than stepping into the required source folder.

Two `__init__.py` files have been removed as they are unneeded and reliance on the current working directory has been reduced. `conftest.py` is an example of where I have removed the weird path setting that was previously required. It should be noted that there are still some places where this still happens, such as the migrations package. We can tackle these further down the track.

The `Dockerfile-flask` file has several changes, and is greatly improved in terms of caching.

`docker compose watch` is now configured for `sc-flask` and hitting `w` while running `make dev` will watch the source files for changes and rebuild automatically. 

I've been able to use my PyCharm IDE integrations to run and debug a number of commands, though I've not tried to bring Flask up yet. It should certainly be doable. 

I've also added a `README.md` file with some examples of how to work with the Python source both natively and in the container.


agilgur5 (commented):
For future reference:

> # Issue 3564

It's best to name your PRs with the _specific_ action(s) you are doing in the PR, i.e. based on their content, rather than meta-content like an issue number (#3564 was also not properly referenced).
That way one can tell what the PR is about at a glance; with an issue number title, I have no clue what it means without clicking into it.
Similarly, if the title is overly generic (e.g. "refactor x"), then it is hard to distinguish from other PRs at a glance. Something more _specific_, like "Modify `server` directory structure and modules" is more understandable at a glance as I'd realize it has renames and moves involved

> `docker compose watch` is now configured

Also this should close #3556

> This is my first attempt at a stacked pull request. Pull request #3562 should be merged before this one.

I suggest using draft PRs until the first one is merged and the [base branch of this PR changed](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request), so that the stacked PRs are not as easy to accidentally merge

Since this has essentially been moved to #3575, I'll review the actual diff/contents there

Issue #2600, opened 2023-02-11
review PWA


sujato:
The PWA function is a few years old now and we have made many changes. We should carefully review it and ensure we are using up-to-date best practices. In addition, check the following specific issues:

- there's a bug where the offline version shows duplicate icons in the toolbar
- several users report not being able to get it working.
- there seems to be a bug with the app icon in the manifest

https://discourse.suttacentral.net/t/offline-site-not-working/28083/6

https://discourse.suttacentral.net/t/suttacentral-bug-reports/27763/12

thesunshade:
There is a duplicate issue here https://github.com/suttacentral/suttacentral/issues/2685 that I am going to close.

This is the only information in it, posted by Bhante Sujato:

*****

It has been several years since we looked at the PWA functionality. We should do a careful review to ensure it is working as well as possible.

    investigate modern PWA standards and ensure compliance
    bug test on as many devices as possible
    review performance on low-end devices and networks
    review strategy for downloads, consider optimizing

See several discussions on Discourse, such as:

https://discourse.suttacentral.net/t/an-inconvenient-pwa-offline-behavior/29289

agilgur5:
Jotting some notes down here after a quick glance after discussion with Ven. thesunshade:

- [ ] Some of the deps are a bit outdated, like the [`workbox-webpack-plugin`](https://github.com/suttacentral/suttacentral/blob/6acc2e161b8b1399d6bc12ba4e5243781b6f0685/client/package.json#L106), which has some CVEs right now in its transitive deps (and is pinned) and doesn't match the version of `workbox-sw`
    - but they were up to date at the time of this issue, so may not be the cause of this
- [ ] Since PWAs will inevitably lose cached data from eviction, all graceful fallbacks should be reviewed. It's possible the PWA is working as it is supposed to, but the fallbacks are not so graceful, causing confusing UX
- [ ] The [Persistent Storage API](https://web.dev/articles/persistent-storage) that I mentioned [on the forum](https://discourse.suttacentral.net/t/sutta-pitaka-app-work-in-progress-feedback-and-collaboration-welcome/40496/19?u=agilgur5) should be used to (attempt to) prevent auto-eviction
    - if all else fails, an offline wrapper app might be worthwhile per the forum post

thesunshade:
Thanks for that! Bhante @sujato is this still something you would like to have examined/fixed?

PR #3575, opened 2026-03-01
Refactor project structure and update build configuration

ihongda:


agilgur5:
> _No description provided._

This is https://github.com/suttacentral/suttacentral/pull/3574 but to the `main` branch essentially.
For future reference, next time you can [manually change the base branch of the PR](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/changing-the-base-branch-of-a-pull-request) instead of merging and making a new PR. That is particularly important to retain any comment/review history, but also WIP tracking -- like my draft review and checked files have to be redone here now

agilgur5:
> > _No description provided._
> 
> This is #3574 but to the `main` branch essentially.

For the original description, to quote @jhanarato:

> Most of the changes here are files moving from the `server/server` directory into `server/src` and `server/tests`. A number of Python files have had their imports modified, perhaps more than necessary as PyCharm has helpfully removed unused imports as part of the refactoring.
>
> Our `Makefile` retains the same targets with changes to the locations of the tests run by pytest and running the Flask manage commands via the installed project rather than stepping into the required source folder.
>
>Two `__init__.py` files have been removed as they are unneeded and reliance on the current working directory has been reduced. `conftest.py` is an example of where I have removed the weird path setting that was previously required. It should be noted that there are still some places where this still happens, such as the migrations package. We can tackle these further down the track.
>
> The `Dockerfile-flask` file has several changes, and is greatly improved in terms of caching.
>
> `docker compose watch` is now configured for `sc-flask` and hitting `w` while running `make dev` will watch the source files for changes and rebuild automatically.

Also note this PR should close #3556 (and #3564?)

> I've been able to use my PyCharm IDE integrations to run and debug a number of commands, though I've not tried to bring Flask up yet. It should certainly be doable.
>
> I've also added a `README.md` file with some examples of how to work with the Python source both natively and in the container.

server/Dockerfile-flask
```diff
@@ -8,12 +8,26 @@ RUN mkdir -p /opt/sc/sc-flask
...
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --locked --no-install-project
+
+COPY pyproject.toml uv.lock README.md ./
+COPY src/ ./src
+COPY tests/ ./tests
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --locked
```
agilgur5, line 35: **EDIT**: I got confused here by misreading the bind mounts above. See [below](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2873987239) clarifications

Why is `uv sync --locked` run twice? The previous:

```
# Setup virtual environment with uv
COPY pyproject.toml uv.lock .python-version ./
RUN uv sync --locked
```

is correct.

Also, since `src/` and `tests/` were copied over before this line, this will bust cache whenever the source code changes and have to re-install all dependencies, even if no dependencies were changed (very inefficient).
That's why it's supposed to only run after `COPY pyproject.toml uv.lock .python-version ./`, as it was before

server/Dockerfile-flask
```diff
@@ -8,12 +8,26 @@ RUN mkdir -p /opt/sc/sc-flask
...
+
+ENV PATH="/opt/sc/sc-flask/.venv/bin:$PATH"
```
agilgur5, line 41: I'm also confused why this line was added back? This was removed in #3562, which I thought was correct per my comment in: https://github.com/suttacentral/suttacentral/pull/3557#discussion_r2839171404

server/Dockerfile-flask
```diff
@@ -8,12 +8,26 @@ RUN mkdir -p /opt/sc/sc-flask
 
 WORKDIR /opt/sc/sc-flask
 
-# Setup virtual environment with uv
-COPY pyproject.toml uv.lock .python-version ./
-RUN uv sync --locked
+ENV UV_COMPILE_BYTECODE=1
+ENV UV_LINK_MODE=copy
+ENV UV_TOOL_BIN_DIR=/usr/local/bin
```
agilgur5: What's the purpose of setting these? There's no comments here

server/README.md
```diff
@@ -0,0 +1,118 @@
...
+is made available on the PATH and is automatically activated within the container. 
+
+## Native Development Outside the Container
+
+Our application can be run outside the container and this is often the best place for development.   
+
+Within the `server/` directory we can create a new virtual environment:
+
+```bash
+$ uv sync
```
agilgur5: The `$` is generally an anti-pattern when there is no output shown since it makes the whole line uncopyable -- see [`markdownlint` MD014](https://github.com/markdownlint/markdownlint/blob/v0.15.0/docs/RULES.md#md014---dollar-signs-used-before-commands-without-showing-output)

docker-compose.dev.yml
```diff
@@ -6,6 +6,10 @@ services:
       - ./client/:/opt/sc/frontend
     ports:
       - '5001:5000'
+    develop:
+      watch:
+        - action: rebuild
```
agilgur5, line 29: `rebuild` is largely unnecessary and would make the performance worse than a bind mount. As I mentioned in https://github.com/suttacentral/suttacentral/issues/3556#issuecomment-3869630475, you should use `sync` for most purposes.

You can use `rebuild` for `pyproject.toml` and `uv.lock` specifically (wherein a build command (`uv sync`) needs re-running if they change)

```suggestion
        - action: sync
```

server/tests/migrations/test_runner.py
agilgur5, line 1: I think these changes make sense, but then [this line in the top-level `README.md`'s "Style Guidelines"](https://github.com/suttacentral/suttacentral/blob/6acc2e161b8b1399d6bc12ba4e5243781b6f0685/README.md?plain=1#L185) needs to be modified a bit.

Also if you plan to separate, e.g. ETL and search as you mentioned in https://github.com/suttacentral/suttacentral/issues/3556#issuecomment-3874722360, then it would actually make more sense for them to have their own `tests/` directories (and probably Dockerfiles too)

server/README.md
```diff
@@ -0,0 +1,118 @@
+# SuttaCentral Flask Server
+
+This part of the project contains the various files needed to create the `sc-flask` docker image. Our Python code is 
+managed by `uv` and implements the backend API for the main [SuttaCentral.net](SuttaCentral.net) website. There are also
+some administrative tools and the ETL system that loads data into the ArangoDB server.
```
agilgur5: FYI, I would not recommend splitting sentences in prose (i.e. newline _inside_ of a sentence). If I have to make a change to a sentence, then I'd have re-align every line, which gets confusing and makes for complicated diffs, to say the least

I would suggest only having new lines in between sentences, not in the middle of a sentence

agilgur5 (changes requested):
I haven't gone through all the files yet (~a dozen I need to look at more closely), but I do have some significant concerns with several lines; see in-line comments below:

server/server/tools/__init__.py
agilgur5, line 1: "Whitespace-only changes." - same note as https://github.com/suttacentral/suttacentral/pull/3562#discussion_r2839135969

server/Dockerfile-flask
```diff
@@ -8,12 +8,26 @@ RUN mkdir -p /opt/sc/sc-flask
...
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --locked --no-install-project
+
+COPY pyproject.toml uv.lock README.md ./
+COPY src/ ./src
+COPY tests/ ./tests
```
agilgur5, line 33: `tests/` and dev deps should only be in the dev & test Dockerfiles, not in the prod one.
Generically speaking, if you want to run dev tools inside of Docker, you'd want a second Dockerfile, i.e. `Dockerfile-flask-dev` (or probably a single file multistage with multiple outputs would be better) that builds on top of this one, install dev deps, and adds the `tests/` dir.

(dev deps and such inside of prod containers also increase the security surface area / possible attack vectors)

server/Dockerfile-flask
jhanarato: My apologies for the lack of comments. 

Please have a look at this example from the astral repo:

https://github.com/astral-sh/uv-docker-example/blob/main/Dockerfile

server/Dockerfile-flask
jhanarato: Yeah, for sure. We're not finished here by any measure.

server/Dockerfile-flask
jhanarato: Are you sure about that? 

From the link above, with comments:

```docker
# Install the project's dependencies using the lockfile and settings
RUN --mount=type=cache,target=/root/.cache/uv \
    --mount=type=bind,source=uv.lock,target=uv.lock \
    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
    uv sync --locked --no-install-project

# Then, add the rest of the project source code and install it
# Installing separately from its dependencies allows optimal layer caching
COPY . /app
RUN --mount=type=cache,target=/root/.cache/uv \
    uv sync --locked
```

server/Dockerfile-flask
jhanarato: OK, it's getting late here and I'm a bit tired...

Are you talking about this line in the last PR?

```
ENV PATH="/opt/sc/venv/.venv/bin:$PATH"
```

That's not the same as the above line which is in the WORKDIR.

While the `.venv` is now created in the WORKDIR as you'd expect, I thought it might not be a bad idea to make it the default Python to avoid using the container version. I mean, we can now ditch the 

```
FROM python:3.11.3
```

and find another image that everyone is happy with, but I wasn't going to make such a bit change in this PR.

server/README.md
jhanarato: That seems reasonable.

server/README.md
jhanarato: Again, reasonable.

docker-compose.dev.yml
jhanarato: We can look into this. Performance is definitely worse than the mount we had before, but `uv` is blindingly fast and I would like to do most dev work outside the container, rebuilding when I need to test with the whole stack.

There are other issues involving how we run the tests. Should pytest be run against the code installed in the `.venv` or the `src/` directory?

Again note that the tests were originally run this way and I've preserved that behavior for now. Otherwise I'd be making even bigger changes, and these last two PRs are already big.

agilgur5: The `uv` example Docker repo you linked similarly uses `sync` for most things: https://github.com/astral-sh/uv-docker-example/blob/2e459bb6210e11ba0444655d5820e7204ef75571/compose.yml#L21

and only `rebuild` on `uv.lock`

> but `uv` is blindingly fast and I would like to do most dev work outside the container

That's orthogonal to using `rebuild`

> rebuilding when I need to test with the whole stack.

If you only change source code, you don't need to `rebuild`, you can just `sync` the files.

Again, `rebuild` after _every_ change in the directory is unnecessary. It slows your dev cycle to rebuild when there's no need to.

> Should pytest be run against the code installed in the `.venv` or the `src/` directory?

It doesn't matter too much in the case of an application (vs a library). I would do it against `src` unless there were a specific reason not to. As an application, it also doesn't need to have an editable install either, as I mentioned in https://github.com/suttacentral/suttacentral/issues/3564#issuecomment-3968640432

I'd strongly recommend mirroring the example's `watch:` configuration with a comment or two

server/Dockerfile-flask
agilgur5: Oh you're right, my apologies!
My eyes go a little dyslexic reading the `--mount=...` lines with no spaces 😵‍💫
The comments actually do help quite a bit with the readability and understandability

So the first `uv sync` is just deps and the second is just the project source itself, which we don't technically need an editable install of.
I am confused why they decided to use a build-time bind mount for the `uv.lock` and `pyproject.toml` instead of the simpler `COPY` of them instead 🤔
  - I'm reading through https://github.com/astral-sh/uv/issues/5792#issuecomment-2343856640, https://github.com/astral-sh/uv/pull/7166, etc to see, although it seems more specifically useful for a monorepo approach

server/Dockerfile-flask
jhanarato: I think a multistage build is the way to go. 

There are a couple of issues though. 

- We need some way to run the tests in CI. Previously we just built the `sc-flask` container and ran them in docker.
- There are some dependencies on having the source tree present. PWA size calculation in the `tools` package and the migrations scripts if I recall correctly.

jhanarato:
OK, so note that I did have some trouble with GitButler, especially regarding the `Dockerfile-flask` file. I managed to make a commit disappear. As @agilgur5 mentioned the `git reflog` command I thought I'd go hunting. I found this:

```
3c1e3d41a HEAD@{92}: commit: Add build system and build backend for "server" module.
9f5aad69a HEAD@{93}: commit (amend): Improve caching to improve image build times.
022b3cc3a HEAD@{94}: commit (amend): Improve caching to improve image build times.
5dede4484 HEAD@{95}: commit: Use two phase uv sync to cache and improve build time.
067e94308 (origin/update_build_config, update_build_config) HEAD@{96}: commit: Update Makefile to use uv
```



So when you look at the commits for this PR the first three  commits on the new branch make these changes:

```bash
jr@JR25:~/Code/suttacentral$ git diff 067e94308 9f5aad69a
diff --git a/server/Dockerfile-flask b/server/Dockerfile-flask
index 274156d16..da014ca07 100644
--- a/server/Dockerfile-flask
+++ b/server/Dockerfile-flask
@@ -8,12 +8,21 @@ RUN mkdir -p /opt/sc/sc-flask
 
 WORKDIR /opt/sc/sc-flask
 
-# Setup virtual environment with uv
-COPY pyproject.toml uv.lock .python-version ./
-RUN uv sync --locked
+ENV UV_COMPILE_BYTECODE=1
+ENV UV_LINK_MODE=copy
+ENV UV_TOOL_BIN_DIR=/usr/local/bin
 
-# Copy over remaining project files.
-COPY entrypoint.sh uwsgi.ini ./
+    --mount=type=bind,source=uv.lock,target=uv.lock \
+    --mount=type=bind,source=pyproject.toml,target=pyproject.toml \
+    uv sync --locked --no-install-project
+
+COPY pyproject.toml uv.lock ./
 COPY server/ ./server
 
+    uv sync --locked
+
+COPY entrypoint.sh uwsgi.ini ./
+
 CMD bash entrypoint.sh
\ No newline at end of file
```

Given the concerns about the `Dockerfile-flask` file I wanted to dig this up and document it. 

That said, the GitButler folk are great so I might try it out again.

server/tests/migrations/test_runner.py
jhanarato: I'm not sure whether I'll do this. If so, not right now. I also considered uv workspaces which would split each of the packages in `src/` such that they have their own `src/` and `tests/` directory. Again, maybe, maybe not.

jhanarato:
If we're all happy, I'd like to get this merged with `main` so I can get to work over the next few days. 

Thanks for the detailed review!

server/tests/migrations/test_runner.py
agilgur5: > If so, not right now

Ok, then that README line should be clarified to match the state of this PR where there's a single `tests/` directory for the Python. Otherwise it'd be out-of-sync

server/Dockerfile-flask
agilgur5: > * Previously we just built the `sc-flask` container and ran them in docker.

We can do the same thing if we still want to run them in Docker. The dev/test image would just be the prod image plus the `tests/` dir and dev dependencies on top and is what would run in CI.

server/Dockerfile-flask
agilgur5: > I am confused why they decided to use a build-time bind mount for the `uv.lock` and `pyproject.toml` instead of the simpler `COPY` of them instead 🤔

Ok, based on the above (which were to modify the guide at https://docs.astral.sh/uv/guides/integration/docker/), it looks like this is just an optimization to save some KBs in a specific layer. It's an over-optimization in our case, since we don't use that intermediate layer and the layer cache-busting is identical between the `COPY` and bind mount forms. (edited to be more specific)

I would suggest using the old `COPY` syntax over the bind mounts as it's easier to read.
Or if not that, then at least add an in-line comment here to clarify and link to the guide

server/Dockerfile-flask
agilgur5: Ok, these are explained in the full guide: https://docs.astral.sh/uv/guides/integration/docker/
- `UV_COMPILE_BYTECODE=1` is a production optimization (trade-off of increased build time for faster start-up time)
- `UV_LINK_MODE=copy` suppresses some warnings specifically due to the build-time cache used on the line `--mount=type=cache,target=/root/.cache/uv`. That cache is purely an optimization
- `UV_TOOL_BIN_DIR=/usr/local/bin` I'm not sure is necessary for our usage? Does it work without it? See also https://docs.astral.sh/uv/reference/storage/#executable-directory

I would 100% suggest some comments here, at the very least a link to the guide

agilgur5:
> I managed to make a commit disappear. As @agilgur5 mentioned the `git reflog` command I thought I'd go hunting. I found this:

Nice work figuring out how to use the `reflog` 🙂 

> ```
> 5dede4484 HEAD@{95}: commit: Use two phase uv sync to cache and improve build time.
> ```

Yes, that one is some useful context, though as I've suggested above, in-line comments in the Dockerfile would be even better! (and generally highly preferable when making very nuanced optimizations)

> If we're all happy

Could you make the changes we are agreed on to this PR? E.g. README style, comments in the Dockerfile. I also strongly believe the `action: sync` change etc should be made. I can make the changes myself in a separate PR, but iteration is normally part of code review processes 

>  I'd like to get this merged with `main`

There's also still a few files I haven't reviewed yet and do have some questions on

>  so I can get to work over the next few days.

You should be able to leave this work aside and stack on top of this if needed or make separate branches entirely. I'm guessing you might not be familiar with a multi-branch workflow? If so, I can review that with you in our pairing session too.
For example, I have around a dozen simultaneous branches (see also [my open PRs](https://github.com/suttacentral/suttacentral/pulls/agilgur5), which are a subset of those) and I switch between them as necessary and/or delete them if they get merged.

server/Dockerfile-flask
agilgur5: > I thought it might not be a bad idea to make it the default Python to avoid using the container version.

This could use an in-line comment too 😅 
But I think you've changed all the `python` usage in this PR to `uv run`, so it wouldn't be used directly.

If I'm understanding your intent correctly, if we [split off a dev Dockerfile](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2870870391), this would only be in the dev one? i.e. the only one where you'd possibly run `python` interactively, as the prod one should be immutable

server/src/sc_flask/manage.py
```diff
@@ -3,7 +3,7 @@
 import click
 from migrations.runner import run_migrations
 
-from app import app
+from sc_flask.app import app
```
agilgur5, line 6: it might make more sense to use a relative import for when files inside of `sc_flask` reference each other:
```suggestion
from .app import app
```

I see relative imports are [already used elsewhere in the codebase](https://github.com/suttacentral/suttacentral/blob/6acc2e161b8b1399d6bc12ba4e5243781b6f0685/server/server/api/views/glossary.py#L5), so it's more of a matter of consistent style.

[PEP8](https://peps.python.org/pep-0008/#imports) prefers absolute imports though, so maybe those existing lines should be changed instead 😅

server/src/sc_flask/manage.py
agilgur5, line 1: What was your rationale behind moving the files at the base of `src/` to `src/sc_flask/` instead?

If ETL and search [won't be split up](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2875594202), then it's all one Flask app if I'm understanding correctly, and so having them at the base makes more sense. There's also a clear file tree that way where the base imports from the deeper folders, as opposed to sibling folders. Sibling folders makes more sense for a monorepo of multiple packages (in which case instead of `src/`, one might use `packages/`).

Right now there are no files in `src/`, just other directories.
Right now, `tests/` are all combined, but `sc_flask/` is separated out - those two are inconsistent with each other IMO (multi-package vs single flask app approaches)

agilgur5 (commented):
> There's also still a few files I haven't reviewed yet and do have some questions on

Other than the above unresolved comments, below is my remaining question on the `sc_flask` dir.

I'll also separate out some grammar fixes for the README

server/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+There are several Python packages contained in the `src/` directory. These are installed into a virtual environment 
+via `uv` and the `Dockerfile-flask` script. Unit and integration tests can be found in `tests/`. Our virtual environment
+is made available on the PATH and is automatically activated within the container. 
+
+## Native Development Outside the Container
+
+Our application can be run outside the container and this is often the best place for development.   
+
+Within the `server/` directory we can create a new virtual environment:
```
agilgur5:
```suggestion
Within the `server/` directory, we can create a new virtual environment:
```

- missing comma

server/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+## Working with Docker
+
+The image can be rebuilt while the stack is running, or before bringing it up:
+
+```bash
+$ make run-dev-rebuild
+```
+
+Alternatively docker's `watch` mode can be activated by bringing up the stack normally
```
agilgur5:
```suggestion
Alternatively, Docker Compose's `watch` mode can be activated by bringing up the stack normally
```

- missing comma
- capital for proper noun
- it's Compose specifically (which is sometimes a separate package)

server/README.md
```diff
@@ -0,0 +1,118 @@
...
+$ make run-dev
+```
+
+and then pressing `w` to activate and deactivate watch mode. 
+
+This will cause the `sc-flask` image to  be rebuilt and restarted.
+
+## Working with UV
+
+If we need to add, update or remove packages we can do so using `uv`:
```
agilgur5:
```suggestion
If we need to add, update, or remove packages, we can do so using `uv`:
```

- missing commas

server/README.md
```diff
@@ -0,0 +1,118 @@
...
+< Anumodana! >
+  ----------
+   \   ^__^
+    \  (oo)\_______
+       (__)\       )\/\
+           ||----w |
+           ||     ||
+```
+
+If we're happy with the changes we can update the lock file:
```
agilgur5:
```suggestion
If we're happy with the changes, we can update the lock file:
```

- missing comma

server/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+This part of the project contains the various files needed to create the `sc-flask` docker image. Our Python code is 
+managed by `uv` and implements the backend API for the main [SuttaCentral.net](SuttaCentral.net) website. There are also
+some administrative tools and the ETL system that loads data into the ArangoDB server.
+
+## Python Packaging
+
+There are several Python packages contained in the `src/` directory. These are installed into a virtual environment 
+via `uv` and the `Dockerfile-flask` script. Unit and integration tests can be found in `tests/`. Our virtual environment
+is made available on the PATH and is automatically activated within the container.
```
agilgur5:
```suggestion
is made available on `PATH` and is automatically activated within the container. 
```

- match the style of `ARANGO_HOST` and other dev bits

server/README.md
agilgur5:
```suggestion
If we're happy with the changes, we can update the lock file:
```

- missing comma

server/README.md
```diff
@@ -0,0 +1,118 @@
...
+< Anumodana! >
+  ----------
+   \   ^__^
+    \  (oo)\_______
+       (__)\       )\/\
+           ||----w |
+           ||     ||
+```
+
+At this point we can commit our changes, or roll back in git, sync and rebuild.
```
agilgur5:
```suggestion
At this point we can commit our changes.
Or roll them back, re-sync, and rebuild.
```
It's a little confusing if the part after the "or" is all its own part or not, or if you meant to "sync and rebuild" and either way? It would only be necessary after a rollback, so I assume that's what you meant. In which case, this reads clearer as a separate sentence

server/README.md
```diff
@@ -0,0 +1,118 @@
...
+
+This is fast, and even faster once the dependencies are cached.
+
+We then need to make sure ArangoDB is running, usually by bringing up the whole stack:
+
+```bash
+$ make run-dev 
+```
+
+Back in `server/` we can run our application locally. Running the database migrations for instance:
```
agilgur5:
```suggestion
Back in `server/`, we can run our application locally. Running the database migrations for instance:
```
- missing comma

agilgur5 (commented):
Some purely grammar and style suggestions below

server/README.md
```diff
@@ -0,0 +1,118 @@
...
+$ uv sync
+```
+
+This is fast, and even faster once the dependencies are cached.
+
+We then need to make sure ArangoDB is running, usually by bringing up the whole stack:
+
+```bash
+$ make run-dev 
+```
```
agilgur5: This will, confusingly, duplicatively run the `sc-flask` service in a container [right now](https://github.com/suttacentral/suttacentral/blob/093a443a39cf876a783a4abc233a60384e4cc567/Makefile#L21).

I might suggest a separate command like `make run-dev-local` that only runs the necessary services if you're running the Python natively.

On the same note, showing how to run and open Flask natively would be helpful too

agilgur5 (commented):
ok and one README suggestion that isn't stylistic.

☑️  I've completed reviewing the entirety of the PR now

server/README.md
jhanarato: I've not yet run the flask app natively. When running the other commands (delete database, migrate, data load) I have used both the whole stack and arango by itself. `make run-dev-local` might be helpful, to be sure.

jhanarato:
Hey folks, 

I'm going to address the issues above before we merge.

Cheers!

server/Dockerfile-flask
jhanarato: > It's an over-optimization in our case since we don't use that intermediate layer and the caching is identical.

You're right. We do need to retain this change to ensure we cache the dependencies:

```docker
RUN --mount=type=cache,target=/root/.cache/uv \
    uv sync --locked
```

I had assumed that the two phase `uv sync` was why we got the speed up.

server/Dockerfile-flask
jhanarato: I'll add a comment for now.

server/Dockerfile-flask
jhanarato: I've included the comments provided by astral and put the docker integration link at the top of the file.

server/README.md
jhanarato: Oh right, I can soft wrap lines in PyCharm. That makes writing prose a bit easier without actually inserting newlines.

jhanarato: Removed `$` except where it is needed to distinguish commands from output.

server/README.md
jhanarato: Fixed and learned how to soft-wrap in PyCharm.

server/src/sc_flask/manage.py
jhanarato: Personally, I prefer absolute imports. However, we do not have great test coverage, apart from the work I did on legacy text imports. The potential for breaking something is pretty high.

server/Dockerfile-flask
agilgur5: > the caching is identical.

This might've been confusing - by that part, I meant the _layer_'s cache-busting, not the cache mount. I updated the wording a bit to clarify

Technically the cache mount is not _necessary_ either, but that one might be a substantial speed up. `uv` will cache by default, but specifying a cache mount will share that cache even _between rebuilds_.
Without it, in the Docker context at this line, the `uv` cache would be empty. So the cache mount should make most rebuilds only need to download new deps and not need to re-download unchanged deps.

So, for clarity, the cache mount only speeds up rebuilds. So when I run this for the first time on my end, it'll have no effect yet.


> I had assumed that the two phase `uv sync` was why we got the speed up.

It's a few different factors for `uv`. A few off the top of my head:
- As with `poetry` etc, the [dependency resolution algorithm](https://docs.astral.sh/uv/concepts/resolution/) itself is substantially faster than plain `pip` (unless there have been any recent optimizations upstream, see also `pip`'s [resolution doc](https://pip.pypa.io/en/stable/topics/dependency-resolution/))
    - Ditto with the lock file. That makes it entirely deterministic and can skip resolution entirely
    - It also is a lot better at resolving possible conflicts by default in my experience
- `poetry` is already a big speed up as such, but it's written in Python. `uv` is written in Rust and compiled to native, so as such its own code is much faster too.
   - It's part of a newer breed of dev tools (mostly in JS/TS and Python ecosystems) specifically written in compiled languages for specific ecosystems, mostly in Rust, Go, or Zig
     - In Rust in particular, zero-copy semantics are quite popular, which can significantly improve memory usage and any (de)serialization performance. Although I'm not sure how many of those techniques `uv` uses specifically, it doesn't seem to be explicitly documented 
   - Not Python, so no GIL to bottleneck when multi-threading
- `pip` caches between venvs [by default](https://pip.pypa.io/en/stable/topics/caching/) these days too, but it copies dependencies from the cache versus making hard links as `uv` does (or symlinks [as `poetry` does](https://python-poetry.org/docs/configuration#virtualenvsoptionsalways-copy)).
    - This is not really relevant in a fresh cache Docker build though
      - And the cache mount + explicit `UV_LINK_MODE=copy` (as discussed [below](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2880391380)) disable that within the Docker build (as hard links don't work across file systems)
      - But the linking will make performance improvements on a native dev computer with many projects (also when adding/removing cached deps) and the cache mount on rebuilds

server/README.md
jhanarato: `uv run --env-file=env/.local.dev.env python -m sc_flask.app` does indeed bring up the server.

server/Dockerfile-flask
```diff
@@ -1,19 +1,48 @@
...
 
-# Copy over remaining project files.
+# Copy the python project files
+COPY pyproject.toml uv.lock README.md ./
+COPY src/ ./src
+COPY tests/ ./tests
+
+# Create a .venv with dependencies and the packages in src/
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --locked
```
agilgur5: I think I confused you here as there's a few different "caches" -- I clarified the wording in https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2881950048

`src/` and `tests/` should still go _after_ installing dependencies, so that they don't bust the dependencies' _layer_ cache. Since the deps change less frequently than the source code, separating them out is common.
The main thing I was scrutinizing was the `COPY` vs. bind mounts, which are equivalent in our usage -- so the _ordering_ was actually fine. (But I did [_initially_](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2870796262) mistakenly think the ordering wasn't fine when I misread the bind mounts -- that was my bad. Sorry for the confusion!)

And if you want an editable install, you'll need to do run `uv sync` a second time too. So this is what I meant:
```suggestion
# Copy dependency manifests and install them into a virtual environment
COPY pyproject.toml uv.lock README.md ./
RUN --mount=type=cache,target=/root/.cache/uv \
    uv sync --locked --no-install-project
    
# Copy source code and editable install packages into the .venv
COPY src/ ./src
COPY tests/ ./tests
RUN --mount=type=cache,target=/root/.cache/uv \
    uv sync --locked
```

Again, sorry for the confusion! Let me know if you're still confused or if it makes sense now 😅

server/Dockerfile-flask
```diff
@@ -1,19 +1,48 @@
...
+
+# Make our uv managed .venv the default Python to avoid confusion
+# with the system environment provided by the docker image.
+# If you run python in the docker container without uv run it
```
agilgur5:
```suggestion
# If you run python in the docker container without `uv run` it
```

Backticks to differentiate the "run python" wording from "`uv run`" _command_.
I got confused on my first read of this with the two "run"s until I re-read it 😵‍💫 hopefully this change makes it more explicit

server/Dockerfile-flask
```diff
@@ -1,19 +1,48 @@
+# Using the python image is redundant now that we have a setup with uv.
+# Perhaps we should update to something else such as the astral images:
```
agilgur5, line 2: For production, I always recommend distroless images, as the security surface area is so much smaller when there is no shell or package manager in the container.

But as a result, they are hard to dev in, so in a [next step with multi-stage](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2870870391), we can have two different base images for dev vs prod. (ideally test and prod are the same base image though) 

I can also show you how to use [debug containers](https://docs.docker.com/reference/cli/docker/debug/), which would be better for dev/prod parity but have their own learning curve, so no rush there

server/Dockerfile-flask
```diff
@@ -1,19 +1,48 @@
...
 
 WORKDIR /opt/sc/sc-flask
 
-# Setup virtual environment with uv
-COPY pyproject.toml uv.lock .python-version ./
-RUN uv sync --locked
+# Enable bytecode compilation
+ENV UV_COMPILE_BYTECODE=1
+
+# Copy from the cache instead of linking since it's a mounted volume
```
agilgur5:
```suggestion
# Copy from the cache instead of hard linking due to the cache mount
```

- _hard_ linking specifically. symlinks can still be usable over different filesystems, but not hard links.
- more explicitly state "cache mount" so it's clearer that you're referencing `--mount=type=cache`

server/Dockerfile-flask
```diff
@@ -1,19 +1,48 @@
...
 
 RUN mkdir -p /opt/sc/sockets
 RUN mkdir -p /opt/sc/sc-flask
 
 WORKDIR /opt/sc/sc-flask
 
-# Setup virtual environment with uv
-COPY pyproject.toml uv.lock .python-version ./
-RUN uv sync --locked
+# Enable bytecode compilation
```
agilgur5:
```suggestion
# Enable bytecode compilation for better start-up performance at the cost of longer build time
```

more explicitly state the trade-off

server/README.md
```diff
@@ -0,0 +1,116 @@
+# SuttaCentral Flask Server
+
+This part of the project contains the various files needed to create the `sc-flask` docker image. Our Python code is managed by `uv` and implements the backend API for the main [SuttaCentral.net](SuttaCentral.net) website. There are also some administrative tools and the ETL system that loads data into the ArangoDB server.
```
agilgur5:
```suggestion
This part of the project contains the various files needed to create the `sc-flask` docker image.
Our Python code is managed by `uv` and implements the backend API for the main [SuttaCentral.net](SuttaCentral.net) website.
There are also some administrative tools and the ETL system that loads data into the ArangoDB server.
```

1 sentence per line here [as well](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2870846595)

server/README.md
```diff
@@ -0,0 +1,116 @@
...
+
+## Python Packaging
+
+There are several Python packages contained in the `src/` directory. These are installed into a virtual environment via `uv` and the `Dockerfile-flask` script. Unit and integration tests can be found in `tests/`. Our virtual environment is made available on `PATH` and is automatically activated within the container.
```
agilgur5:
```suggestion
There are several Python packages contained in the `src/` directory.
These are installed into a virtual environment via `uv` and the `Dockerfile-flask` script.
Unit and integration tests can be found in `tests/`.
Our virtual environment is made available on `PATH` and is automatically activated within the container.
```

1 sentence per line here [too](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2885487095)

agilgur5 (commented):
I think I confused you on the `Dockerfile` a bit (my bad!), see my suggestions below for what I meant and I can clarify anything still confusing

server/src/sc_flask/manage.py
agilgur5: > Personally, I prefer

With most style things, my opinion is to pick one and stick with it 😅 Preferably chosen by an auto-formatter to avoid bikeshedding

> However, we do not have great test coverage, apart from the work I did on legacy text imports. The potential for breaking something is pretty high.

Ah ok. So may change the existing relative imports to absolute imports later.
Fortunately, testing imports just requires a smoke test of importing the file. I might do that when I get `ruff` set-up

server/src/sc_flask/manage.py
jhanarato: Our ETL is definitely distinct from the flask server. 

`data_loader` uses configuration settings from Flask, but I'm planning to remove this dependency:

https://github.com/suttacentral/suttacentral/issues/3576

`data_loader` could be run in a separate container. I certainly considered this. I'm not sure of the performance implications though.

Once I've addressed the `data_loader` issues I created last year I will be refactoring both the front end and back end of search. I do feel that search could be independently deployed. 

Now *why* did move the files? I want all the application code installed in a `.venv`. I just did a little experiment and yes, if you have a single module at the top level it is still installed in the `.venv`. What would you suggest? I've also moved `wait_for_arango.py` to `scripts/`.

One thing I have considered are uv workspaces:

https://docs.astral.sh/uv/concepts/projects/workspaces/

As for the `tests/` hierarchy, some points:

- Every top level package that had unit tests has a directory. 
- It might be best practice to add `__init__.py` files to `tests/`. I wasn't sure and didn't.
- The idea is to mimic the structure of `src/` directory, but that's a lot of refactoring and can wait, I think.

server/src/sc_flask/manage.py
jhanarato: > So may change the existing relative imports to absolute imports later.

Yes, we should do that.

> Fortunately, testing imports just requires a smoke test of importing the file.

I was thinking that. We can do that and restructure the tests together.

> I might do that when I get ruff set-up

Great!

server/src/sc_flask/manage.py
jhanarato: Oh, one other thing. We do need a package with the same name as the project in the `pyproject.toml` file. With an underscore instead of a dash, of course. That happens to be `sc_flask`. Not an excuse to dump everything in it.

server/tests/migrations/test_runner.py
jhanarato: I see. 

Yes, ETL should be split off and may or may not be run in a separate container. I've only scanned the python code for search and can't make a decision right now. This is something that I'm deeply interested in and I'm reading this book among other things:

https://www.cambridge.org/highereducation/books/introduction-to-information-retrieval/669D108D20F556C5C30957D63B5AB65C#overview

I do think that the API for search is one thing - it could be Flask, FastAPI or whatever. Then there's the indexing and retrieval side of things (the interesting stuff). I think we can take a ports & adapters approach to this. Have an interface that the API can use and dependency-inject the retrieval mechanism. Before I can tackle this and develop an architecture I want to test and refactor the current code.

It's going to take a while.

server/Dockerfile-flask
jhanarato: I agree. We can do this in the near future.

As mentioned in our zoom chat, container debugging is not available with the community version of PyCharm. It's available in Pro but I've heard the experience isn't great. But hey, we can debug outside the container now!

server/Dockerfile-flask
jhanarato: I'm fine with separating the tests for CI. We need to get tests working again in CI anyway. Do you want this done as part of this PR or make it something for the near future?

docker-compose.dev.yml
jhanarato: We might need a `.dockerignore` file for `sync`. Things like `__pycache__` and `.pytest_cache`, plus I've got a `.idea` directory in there too.

The other thing - won't `sync` just copy the files into `src/` and not run `uv sync`? That would make changes to `src/` unavailable in the `.venv`.

As I said at the end of the zoom call, I'm still a bit hazy about what actually needs the `src/` directory to be populated.

server/Dockerfile-flask
jhanarato: > Technically the cache mount is not necessary either, but that one might be a substantial speed up. uv will cache by default, but specifying a cache mount will share that cache even between rebuilds.

That's right. Since I'm doing a rebuild every time I want to test a change, whether via the docker compose watch configuration or manually that's handy. 

If we can make sync work for docker compose watch then it's less of an issue.

server/Dockerfile-flask
jhanarato: If I understand correctly, this will make the sync work in `docker compose watch`?

docker-compose.dev.yml
agilgur5: > We might need a `.dockerignore` file for `sync`. Things like `__pycache__` and `.pytest_cache`, plus I've got a `.idea` directory in there too.

 We can skip some of those by `sync`ing the specific files and directories we need, similar to the Dockerfile `COPY`s. `__pycache__` and `*.pyc` might still be necessary to ignore though since they're inside subdirectories.

Specifically, `sync` for `src/`, `test/`, and `pyproject.toml`, then `rebuild` for `uv.lock` when dependencies change.

> won't `sync` just copy the files into `src/` and not run `uv sync`?

Correct.

> That would make changes to `src/` unavailable in the `.venv`.

Ah this is probably one of the key misunderstandings. The changes would be available in the `.venv` because [editable installs are symlinked](https://docs.astral.sh/uv/concepts/projects/dependencies/#editable-dependencies). So the `.venv` has none of the contents of SC source code in it, just the deps and a symlink to the source.
Hopefully that clears things up a bit!

Again, we also could skip the editable install into the `.venv` entirely and just import from `src` directly.

server/Dockerfile-flask
agilgur5: Near future IMO, since all the tests and devDeps being in the image is how it was already, so that can be its own separate change. It'll also be a pretty nuanced change as well that'll be good to separately review

server/Dockerfile-flask
agilgur5: > If we can make sync work for docker compose watch then it's less of an issue.

Yep then it's primary use will be when deps change -- which may still be a big speed up then too

✅  Also this thread can be resolved -- follow-up in https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2885438407

server/Dockerfile-flask
agilgur5: Not quite. This isn't necessary for the `sync`, which bypasses the Dockerfile entirely. See my recent comment in the Compose file thread, I think that should clear up a misunderstanding: https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2893938337

The suggestion is purely a layer cache optimization so that if you rebuild and no dependencies have changed, Docker won't have to re-install them again. (i.e. in the above suggestion, if you only change source code, only line 32 would be layer cache-busted, while the lines before it would use the layer cache)
It is _conceptually_ similar to the Compose `sync` as such, but otherwise orthogonal. Could say it's a similar optimization in two places though, one during build and one during `watch`

server/Dockerfile-flask
agilgur5: I think you missed hitting apply on this suggestion

agilgur5:
I think you missed hitting apply [on one of my comment suggestions](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2885443338)

**EDIT**: ✅ completed

server/src/sc_flask/manage.py
agilgur5: > > I might do that when I get ruff set-up

Btw, if you click the 3 dots icon next to a comment and hit "Quote reply" while having something highlighted, GH will automatically copy the formatting into a comment too.

Screenshot for reference:
Screenshot 2026-03-06 at 2 29 29 PM

server/src/sc_flask/manage.py
agilgur5: > We do need a package with the same name as the project in the `pyproject.toml` file.

Do you have a docs link for that?
This vaguely reminds me of a mistake I made in https://github.com/agilgur5/django-serializable-model/issues/2 / https://github.com/agilgur5/django-serializable-model/pull/3, but that functioned a bit differently; the name was only needed if you want to use it as a namespace to import from. As in, it'd be necessary if we wanted to use the syntax for `from sc_flask.app`, but if the files were at the top level we could do `from app` instead (how it was before the PR)

> I just did a little experiment and yes, if you have a single module at the top level it is still installed in the `.venv`. What would you suggest?

I suppose it kinda depends on the above naming and future plans on splitting 😅
As in, if you were to split things, what would the _final_ directory structure look like then?

> I've also moved `wait_for_arango.py` to `scripts/`.

Yea that seemed like a good change to me 👍 

> * It might be best practice to add `__init__.py` files to `tests/`. I wasn't sure and didn't.

`__init__.py` is only necessary if you're going to have an importable package, and most tests have nothing to export (and the test runner normally does its own thing instead of importing). So they're usually not needed -- and if it isn't needed or used, I wouldn't recommend having it (it would be extraneous)

> One thing I have considered are uv workspaces:

Yea the multi-package directory structure you have in this PR would go along with that. It isn't strictly necessary to use workspaces though (it's more a meta-dev workflow optimization) and it has a bit of its own learning curve too.

jhanarato: I'd like to keep the `sc_flask` package for now. In my mind it should hold the Flask related code. 

`app.py` and `config.py` are clearly Flask related. 

`manage.py` depends heavily on Flask, though mainly to get configuration data.

 `wsgi.py` is an entry point for running the Flask app.

I'd say `manage.py` will change as we refactor the configuration loading, but I don't know how exactly.

> I suppose it kinda depends on the above naming and future plans on splitting 😅
> As in, if you were to split things, what would the final directory structure look like then?

I would like to separate concerns. I've not analyzed the entire codebase, but I'd be thinking of:

- The Flask API
- Data store and retrieval
- Search indexing and retrieval
- Nothing that is called common or utils. 

I'd use a ports and adapters architecture to make Flask and ArangoDB. There is a reasonable chance that we'll need to replace the latter. I've hung out on their Slack server for a while, and it doesn't paint a picture of a healthy community or company. 

I'd like to look into Pydantic for our data flow. Having used Serde in Rust, I can see how it would be useful. We convert a lot of JSON to dictionaries and back again. JSON from git into Arango into the Flask API.

server/src/sc_flask/manage.py
jhanarato: > > > I might do that when I get ruff set-up
> 
> Btw, if you click the 3 dots icon next to a comment and hit "Quote reply" while having something highlighted, GH will automatically copy the formatting into a comment too.

Ahh, nice!

README.md
jhanarato: Update to reflect new directory structure.

jhanarato: Change to new `uv` and `.venv` way of doing things.

docker-compose.dev.yml
```diff
@@ -6,6 +6,10 @@ services:
...
+          path: .
```
jhanarato: You know what? While I was working on this, it was acting rather strangely. The path should be `.server/` not `.`

With the current configuration, we rebuild every time a file changes in the entire project, not just the flask files. This also includes the docker compose file itself. Oops.

docker-compose.dev.yml
jhanarato: > > We might need a `.dockerignore` file for `sync`. Things like `__pycache__` and `.pytest_cache`, plus I've got a `.idea` directory in there too.
> 
> We can skip some of those by `sync`ing the specific files and directories we need, similar to the Dockerfile `COPY`s. `__pycache__` and `*.pyc` might still be necessary to ignore though since they're inside subdirectories.
> 
> Specifically, `sync` for `src/`, `test/`, and `pyproject.toml`, then `rebuild` for `uv.lock` when dependencies change.

OK, this is all working quite nicely. There is no `.dockerignore` file needed as we sync the above as suggested.

> > won't `sync` just copy the files into `src/` and not run `uv sync`?
> 
> Correct.
> 
> > That would make changes to `src/` unavailable in the `.venv`.
> 
> Ah this is probably one of the key misunderstandings. The changes would be available in the `.venv` because [editable installs are symlinked](https://docs.astral.sh/uv/concepts/projects/dependencies/#editable-dependencies). So the `.venv` has none of the contents of SC source code in it, just the deps and a symlink to the source. Hopefully that clears things up a bit!
> 
> Again, we also could skip the editable install into the `.venv` entirely and just import from `src` directly.

I'm sort of getting this now. So the `.venv` is required for the external dependencies, but we don't need to install our own source?

agilgur5: > So the `.venv` is required for the external dependencies, but we don't need to install our own source?

Correct. Since you can import the source directly.
The primary use case for [editable installs](https://setuptools.pypa.io/en/latest/userguide/development_mode.html) is to test a library package you intend to publish. It can also be useful in monorepos with shared libraries

✅ Also the changes to the Compose file look good to me now (as of https://github.com/suttacentral/suttacentral/pull/3575/changes/eaa794dbab6ed73219e391133b9ba650a2272433), so I think this thread can be resolved

server/src/sc_flask/manage.py
jhanarato: > > We do need a package with the same name as the project in the `pyproject.toml` file.
> 
> Do you have a docs link for that? 

I don't. 

What happens is this: if I change the name of the project to `sc-server` for instance, and then run `uv sync`, I get this error:

```bash
jr@JR25:~/Code/suttacentral/server$ uv sync
Resolved 95 packages in 815ms
  × Failed to build `sc-server @ file:///home/jr/Code/suttacentral/server`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `uv_build.build_editable` failed (exit status: 1)

      [stderr]
      Error: Expected a Python module at: src/sc_server/__init__.py

      hint: This usually indicates a problem with the package or the build environment.
```

jhanarato: Ahh, I think I see the problem. This relates to the uv build backend:

https://docs.astral.sh/uv/concepts/build-backend/#using-the-uv-build-backend

We can add this to make it specific:

```toml
[tool.uv.build-backend]
module-name = ["foo", "bar"]
```

Funny thing is, before I refactored to create the `src/` directory, the build backend would be happy enough if foo existed but not bar. Anyway, it's a uv thing, not a python thing.

So what do you think? Move `app.py`, `config.py` and `wsgi.py` back to the top level and remove `sc_flask`?

agilgur5: > are clearly Flask related

Just to be explicit, the contents of the directory are all good to me, the question is just if `sc_flask/` should exist at all versus files at the top-level of `src/`

> * The Flask API

In this hypothetical structure, would `api/` then be moved under `sc_flask/`?
I also see that `api/` has nothing much at it's top-level either so could be merged together with `sc_flask/`. That makes more sense once there is a split though

> * Nothing that is called common or utils.

Is there actually no shared code between all those? 

> This relates to the uv build backend:

> Anyway, it's a uv thing, not a python thing.

I'm not sure if you're familiar with [PEP 517 build backends](https://peps.python.org/pep-0517/#build-backend-interface), but that split up the front-end CLI or package manager (e.g. `uv`) from the build back-end (e.g. `uv-build`). So you can use the `uv` package manager with other build backends, not only `uv-build`.
Build back-ends have separate configurations, but a lot is similar between different ones too. If a `module-name` isn't provided in the config, it may be inferred; `uv-build` specifically may just require an explicit config (which is probably a good thing IMO). 
Also, as you may have noticed, a build backend isn't necessary if we don't have an editable install -- if we import the source code directly, there's no editable install to "build".

> So what do you think? Move `app.py`, `config.py` and `wsgi.py` back to the top level and remove `sc_flask`?

IMO, I would leave them at the top-level until there is a split, then move them into `api/`. But you've got more experience with this codebase and how it is currently organized than me, so I also don't want to give too opinionated a suggestion either, especially when it is more stylistic in nature 😅

server/server/tools/__init__.py
agilgur5: ✅ a `.gitattributes` file was verbatim added in https://github.com/suttacentral/suttacentral/pull/3575/commits/196c80725db812a87723b0374b0868dee4b04d20, although I'm actually not sure if it fixes this specifically since it's later in the commit history 😅

server/tests/migrations/test_runner.py
agilgur5: > that README line should be clarified to match the state of this PR where there's a single `tests/` directory for the Python. Otherwise it'd be out-of-sync

✅ Follow-up in https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2899007111

> This is something that I'm deeply interested in and I'm reading this book among other things:

Yes, I recall you mentioning this in [a forum post](https://discourse.suttacentral.net/t/new-bilara-update/42516/4?u=agilgur5). Information Retrieval was a class in my college and I have some experience with it and indexing, although not that much. I might detail more in that thread, but my primary suggestion may be to check out embedded DBs as well -- since SC's data is mostly static, it could potentially improve performance quite a bit if it were embedded (think SQLite vs MySQL).

> I do think that the API for search is one thing - it could be Flask, FastAPI or whatever. Then there's the indexing and retrieval side of things (the interesting stuff). I think we can take a ports & adapters approach to this.

Agree. Particularly as there are _already_ two search engines with Arango + Algolia (but Algolia is not self-hosted, so a little different)

> It's going to take a while.

For sure, especially with the legacy code bits and such 😅

agilgur5 (approved):
Still discussing [the `sc_flask` directory](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2880596566) and the README has [one TODO leftover](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2899007111), but we've gone through all the major bits and otherwise this all looks good to me!

Thanks for all the iteration and getting these great changes in @jhanarato !

server/src/sc_flask/manage.py
jhanarato: If I do the following:

- change the project name to `sc-server`
- add `module_name.py` to `src/`
- add the module names `["sc_flask", "module_available"]`

I get the following error:

```bash
r@JR25:~/Code/suttacentral/server$ uv run module_available
  × Failed to build `sc-server @ file:///home/jr/Code/suttacentral/server`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `uv_build.build_editable` failed (exit status: 1)

      [stderr]
      Error: Expected a Python module at: src/module_available/__init__.py

      hint: This usually indicates a problem with the package or the build environment.
```

That's what led me to create `sc_flask`

jhanarato: I suggest we keep everything in packages and make a note in `README.md` as to why it is the way it is.

jhanarato:
I just noticed a bug when running `make load-data`. When calculating the PWA sizes, it still tries to write to `/opt/sc/sc-flask/server/tools/`. Changing this to use `src` instead of `server` in the path configuration of `data_loader.run()` does solve the problem for ETL, but I need to investigate where this output is used and ensure it is consistent. 

That said, does the PWA version of the site even work?

jhanarato:
> I just noticed a bug when running `make load-data`. When calculating the PWA sizes, it still tries to write to `/opt/sc/sc-flask/server/tools/`. Changing this to use `src` instead of `server` in the path configuration of `data_loader.run()` does solve the problem for ETL, but I need to investigate where this output is used and ensure it is consistent. 

OK, I see. The data generated is written to the `pwa_sizes` collection. I'll figure out a reasonable place for the json to be written & read. `src/` is not a good location.

jhanarato:
> > I just noticed a bug when running `make load-data`. When calculating the PWA sizes, it still tries to write to `/opt/sc/sc-flask/server/tools/`. Changing this to use `src` instead of `server` in the path configuration of `data_loader.run()` does solve the problem for ETL, but I need to investigate where this output is used and ensure it is consistent.
> 
> OK, I see. The data generated is written to the `pwa_sizes` collection. I'll figure out a reasonable place for the json to be written & read. `src/` is not a good location.

I've opened issue #3579 to resolve this and will leave `pwa_sizes.json` in the `src/` directory for now.

server/src/sc_flask/manage.py
agilgur5: Yea that's fine with me; at this point it might be more effort than it's worth to figure it out.

Just to note down for posterity, I suspect it might be possible with one or more of:
- adding an `__init__.py` into `src/`
- setting `namespace = true` in the `uv-build` configuration
- removing the editable install entirely
- using a different build back-end

agilgur5:
> That said, does the PWA version of the site even work?

Kinda sorta maybe on some browsers. It works for some but not others; see also #2600 

> > > but I need to investigate where this output is used and ensure it is consistent.
> > 
> > 
> > OK, I see. The data generated is written to the `pwa_sizes` collection. I'll figure out a reasonable place for the json to be written & read. `src/` is not a good location.
> 
> I've opened issue #3579 to resolve this and will leave `pwa_sizes.json` in the `src/` directory for now.

Ok, I was also wondering where this was used and if it was necessary 😅

server/Dockerfile-flask
agilgur5: ✅ Follow-up in #3581

jhanarato:
Oops, found another bug!

```
$ docker logs sc-flask 
Waiting for arango to start
Bytecode compiled 3934 files in 359ms
error: Failed to spawn: `scripts/wait_for_arango.py`
  Caused by: No such file or directory (os error 2)
 ```

This is kinda bad, we don't wait for arangodb to start up.

Simple fix though, we just need to add `scripts/` to `Dockerfile-flask`

**Comment by jhanarato** [comment]
I've pushed the fix now. 

It's a bit strange though. The `echo` command in `entrypoint.sh` no longer shows up in the logs. I've moved it to a `print()` in the python code now, which works.

Ahh, on inspection, I think `echo` stopped working when we started using the dockerfile to run the `entrypoint.sh` script rather than docker compose.

It would be nice to have a health check for ArangoDB rather than this script.

File: server/src/sc_flask/manage.py
Comment (agilgur5) [reply server/src/sc_flask/manage.py:1]: ✅  Comment added in https://github.com/suttacentral/suttacentral/pull/3575/changes/9a9905a282a1a322c9b168affe02774a8e8bb96c and follow-up issue https://github.com/suttacentral/suttacentral/issues/3583 created

File: server/README.md [hunk 25]
```diff
@@ -0,0 +1,124 @@
...
+           ||----w |
+           ||     ||
+```
+
+At this point we can commit our changes.
+Or roll them back, re-sync, and rebuild.
+
+## Note on `sc_flask` package
+
+Due to some difficulties with the `uv` build backend, it is difficult to have modules in `src/` that aren't in a package. The modules in `sc_flask` should be at the top level of `src/`. The problem might be solved by removing the editable installation. See issue #3583.
```
agilgur5, line 124: I'd put each sentence on a separate line for consistency. But nbd, as this whole section is likely to be changed simultaneously

server/Dockerfile-flask
```diff
@@ -1,19 +1,51 @@
...
+# If you run python in the docker container without `uv run` it
+# will still use the .venv
+ENV PATH="/opt/sc/sc-flask/.venv/bin:$PATH"
+
+# Copy over the remaining files required to run the application.
+COPY scripts/ ./scripts
 COPY entrypoint.sh uwsgi.ini ./
-COPY server/ ./server
 
-CMD bash entrypoint.sh
```
agilgur5, line 19: You might know this, but `CMD bash entrypoint.sh` and `CMD ["bash", "entrypoint.sh"]` are actually not equivalent since Docker has ["shell form" and "exec form"](https://stackoverflow.com/a/47940538/3431180). In this case, you are executing a shell explicitly anyway, with `bash` and with the `#!/usr/bin/env bash` shebang. 

IMO better not to use a shell if not necessary (or multiple subshells), so I prefer the change you made.

But the reason I mention this is because it _might_ be why the `echo` ended up being swallowed or buffered  as you observed in https://github.com/suttacentral/suttacentral/pull/3575#issuecomment-4062709260

agilgur5 (approved):
LGTM! I just have two tiny comments below.

I approved [earlier](https://github.com/suttacentral/suttacentral/pull/3575#pullrequestreview-3914355886) as well as I think at that point it passed the "umabiguous improvement" standard we [discussed in dev chat](https://discourse.suttacentral.net/chat/c/-/12/6696), so, from my perspective, you're good to merge then or later. Though of course, a final check by me or someone else on the remaining changes is good to have, but I didn't want to block after that point (doubly as I knew I might be out for a bit for my injections 😅 )

> It's a bit strange though. The `echo` command in `entrypoint.sh` no longer shows up in the logs. I've moved it to a `print()` in the python code now, which works.

🤔 
 
> Ahh, on inspection, I think `echo` stopped working when we started using the dockerfile to run the `entrypoint.sh` script rather than docker compose.

Docker compose just _overrides_ the existing `CMD`, so that _shouldn't_ be the case. It might be the "shell form" vs "exec form" being used -- see follow-up below in https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2969343622.
It might also be something else though 🤔, as it's in a shell anyway, but perhaps there's some buffering or swallowing going on depending on the subshells (I'd have to test that a bit more)

server/Dockerfile-flask
agilgur5: I suggested this change upstream in https://github.com/astral-sh/uv-docker-example/pull/83 as well

server/Dockerfile-flask
agilgur5: I suggested this change upstream in https://github.com/astral-sh/uv-docker-example/pull/83 as well

jhanarato:
> I approved [earlier](https://github.com/suttacentral/suttacentral/pull/3575#pullrequestreview-3914355886) as well as I think at that point it passed the "umabiguous improvement" standard we [discussed in dev chat](https://discourse.suttacentral.net/chat/c/-/12/6696), so, from my perspective, you're good to merge then or later. Though of course, a final check by me or someone else on the remaining changes is good to have, but I didn't want to block after that point (doubly as I knew I might be out for a bit for my injections 😅 )

Haha! Sorry about that. I just wanted you to review my final commits to see if I'd done something silly.

OK! Let's merge, @ihongda !

agilgur5:
Yea that works too. And I did leave a few comments above too. But they're not too important and can be done as follow-ups if necessary. "Approve with comments" to me means "could use small improvements, but nbd", versus plain "Approve" is "all good" (but different people interpret it differently -- we could put that into a `CONTRIBUTING.md` too)

server/Dockerfile-flask
jhanarato: Good to know. I'll leave it as is.

ihongda:
Thanks @jhanarato @agilgur5 👍🏼 
I will merge these codes.

server/src/sc_flask/manage.py
agilgur5: > Ah ok. So may change the existing relative imports to absolute imports later.

Ok so we can enforce this with `ruff`'s `flake8-tidy-imports` implementation, specifically by [configuring the rule `ban-relative-imports = "all"`](https://docs.astral.sh/ruff/settings/#lint_flake8-tidy-imports_ban-relative-imports)

> Fortunately, testing imports just requires a smoke test of importing the file. I might do that when I get `ruff` set-up

This one I'm not sure `ruff` can do. Pylint has an [imports checker](https://pylint.pycqa.org/en/stable/user_guide/checkers/features.html#imports-checker) that can do this with the rule `import-error (E0401)`, but `ruff` [doesn't seem to implement that rule](https://docs.astral.sh/ruff/rules/#error-ple)? Perhaps because it is better implemented with a type-checker, i.e. [`ty`](https://github.com/astral-sh/ty)?

`ruff` is now set-up in #3656

server/Dockerfile-flask
agilgur5: ✅ This is followed up on in #3581

PR #3577, opened 2026-03-06
fix: handle potential null reference for getSentenceText in SCTextLegacy

ihongda:


client/elements/text/sc-text-legacy.js
```diff
@@ -665,7 +666,7 @@ export class SCTextLegacy extends SCTextCommon {
 
   _setSCBottomSheet(scBottomSheet, word, chineseLookup, currentTarget) {
     scBottomSheet.currentTarget = currentTarget;
-    const keyword = scBottomSheet.getSentenceText() || word.dataset.latin_text || word.textContent;
+    const keyword = scBottomSheet?.getSentenceText() || word.dataset.latin_text || word.textContent;
```
agilgur5, line 669: So with the `import` you added above, this should be unnecessary, no? It's also not the only place `scBottomSheet` is used

I'm hoping there's a linter check to ensure that if, for instance, `<sc-bottom-sheet>` is used, it is also imported

agilgur5:
> _No description provided._

For reference, this resolves a bug where Chinese lookup would not activate in unsegmented root texts, as reported [on the forum](https://discourse.suttacentral.net/t/chinese-lookup-not-working/42766?u=agilgur5)

client/elements/text/sc-text-legacy.js
ihongda: Importing is necessary because this component needs to be used in both the segmented text component and the legacy text component.

I plan to include necessary testing to avoid this problem.

agilgur5: > Importing is necessary

Sorry if it wasn't clear, my comment is on the `?` optional chaining addition. The `?` should no longer be necessary now that you've `import`ed the component, no? As in, it should always exist now

> I plan to include necessary testing to avoid this problem.

That too could check for that 😅
I mentioned a linter check as that would be a quicker way to detect the lack of import via static analysis (i.e. without needing to run a test at all)

ihongda: > Sorry if it wasn't clear, my comment is on the ? optional chaining addition. The ? should no longer be necessary now that you've imported the component, no? As in, it should always exist now

Yes, you are correct. If the module has already been imported and no null values ​​are expected, the ? is indeed unnecessary.

PR #3578, opened 2026-03-09
clean(deps): remove unused `sphinx` deps

agilgur5:
## Summary

Remove unused `sphinx` dep and its unused transitive deps in `pyproject.toml` and `uv.lock`

This was something I noticed when I first started contributing (c.f. footnotes of https://github.com/suttacentral/bilara-data/pull/4739#issuecomment-3694488270) as well as in code review of https://github.com/suttacentral/suttacentral/pull/3557#discussion_r2839209854

## Details

- remove `sphinx`, its `contrib` plugins, and its transitive dependencies
  - `sphinx` is [not used](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral%20sphinx&type=code) in the codebase
    - there is a single old issue that was written up about it for dev docs (#809), but it seems to have never been implemented
      - IMO, I'd suggest using plain markdown inside relevant directories for dev docs (see also [prior dev chat](https://discourse.suttacentral.net/chat/c/-/12/5193)), unless some specific features are desired
  - the transitive dependencies removed are not used in the codebase directly nor indirectly by any other dependency, as can be seen in [the `uv.lock`](https://github.com/suttacentral/suttacentral/blob/0c60a09c9bbc1c2716512e23e339fa2d772857da/server/uv.lock)
    - removed `alabaster`, `babel`, `docutils`, `imagesize`, `pygments`, and `snowballstemmer`
      - here are [individual](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+alabaster&type=code) [code](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+babel+language%3APython+&type=code) [search](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+docutils&type=code) [links](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+imagesize&type=code) [of](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+pygments&type=code) [these](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+snowballstemmer&type=code) being unused
  - all of these (and more) will also pop up as unused / unimported deps if you run a dependency checker like [FawltyDeps](https://github.com/tweag/FawltyDeps) (e.g. `uvx fawltydeps`)
  
 ## Verification
 
 - Code search (see above)
 - Ran `uvx falwtydeps` (see above)


jhanarato:
I can do this as part of #3581. I'll have to review the dependencies that make it through to production anyway.

agilgur5:
While the concerns are related, the changes can be done completely independently. Smaller PRs are usually better (easier to review, merge, revert, etc).

I also have another few PRs in this series already cooked up, and each of them has a similarly detailed analysis of the dependencies

jhanarato:
> While the concerns are related, the changes can be done completely independently. Smaller PRs are usually better (easier to review, merge, revert, etc).
> 
> I also have another few PRs in this series already cooked up, and each of them has a similarly detailed analysis of the dependencies

Sounds good to me.

jhanarato:
> I can do this as part of #3581. I'll have to review the dependencies that make it through to production anyway.

Hmm... looks like I thought this was an issue, not a PR. Let's take a look at what you've done.

jhanarato:
OK, plan is I'll go through the history and double check each package to see that it can be safely removed.

agilgur5:
You can do that as an additional triple check if you want to, but it isn't strictly necessary since my checks should be sufficient and you can validate those as-is unless you want to add another layer of verification.

There's essentially two types of checks:
1. Static analysis (I did this in this PR)
    1. Automated as with `fawltydeps` above, but which can require configuration to get all possible usage, so is not _necessarily_ foolproof (i.e. it can catch everything _in theory_, depending on how you use it)
    1. More manual with code searches above
1. Dynamic analysis
    1. Run the code, CI, tests, and other dev workflows and make sure nothing breaks 
    
 Historical analysis is more on the static side, and I did do it where it was easy to determine the origin (if it wasn't, I didn't spend too much time tracing), as in the one Sphinx issue above. A part of it can be dynamic, as certain developer workflows might not be documented well or some scripts or commands can depend on it in a more transitive way that can be hard for static analysis to detect. That part is something I don't have full clarity on (for both my lack of experience in this codebase _and_ the lack of some dev docs 😅) and your assistance could be helpful on.
 
 But verify however you want to, just wanted to add some context!
 Also note that the deps I've started with are the ones that are _easier_ to check. Like I didn't remove `pngnq` (yet) as in https://github.com/suttacentral/suttacentral/pull/3562#discussion_r2839157388 as I wasn't sure if it might be a sub-dependency of something (it probably isn't and probably can be removed, but I went with the easier/safer ones that I could more reliably check first)

jhanarato:
> You can do that as an additional triple check if you want to, but it isn't strictly necessary since my checks should be sufficient and you can validate those as-is unless you want to add another layer of verification.

It's probably fine, and we can just go ahead. I wanted to offer more than a "looks good to me".

> There's essentially two types of checks:
> 
>     1. Static analysis (I did this in this PR)
>        
>        1. Automated as with `fawltydeps` above, but which can require configuration to get all possible usage, so is not _necessarily_ foolproof (i.e. it can catch everything _in theory_, depending on how you use it)
>        2. More manual with code searches above
> 
>     2. Dynamic analysis
>        
>        1. Run the code, CI, tests, and other dev workflows and make sure nothing breaks

Our Python unit test coverage is pretty limited. Generally I do the following:

- Run `make delete-database`
- Run `make migrate`
- Run `make load-data`
- Run `make index-arangosearch`
- Run `make run-dev`
- Check I can view a sutta or two
- Perform a search or two

Hopefully that would flush out anything as coarse as a missing python dependency, but you never know. The PWA stuff was a bit of a mystery.

> 
> Historical analysis is more on the static side, and I did do it where it was easy to determine the origin (if it wasn't, I didn't spend too much time tracing), as in the one Sphinx issue above. A part of it can be dynamic, as certain developer workflows might not be documented well or some scripts or commands can depend on it in a more transitive way that can be hard for static analysis to detect. That part is something I don't have full clarity on (for both my lack of experience in this codebase _and_ the lack of some dev docs 😅) and your assistance could be helpful on.

I actually enjoy going back in time and seeing how we got to where we are now. I call it "software archaeology".

> But verify however you want to, just wanted to add some context! Also note that the deps I've started with are the ones that are _easier_ to check. Like I didn't remove `pngnq` (yet) as in [#3562 (comment)](https://github.com/suttacentral/suttacentral/pull/3562#discussion_r2839157388) as I wasn't sure if it might be a sub-dependency of something (it probably isn't and probably can be removed, but I went with the easier/safer ones that I could more reliably check first)

I can see that Blake added `pngnq` and nothing added around the same time seems related.

agilgur5:
> I call it "software archaeology".

Yea you mentioned this on the call and while I was writing that comment I was like "oh wait J.R. likes this kind of stuff" 😆
I think we also touched on the difficulty in this codebase that the some of the old decisions weren't necessarily documented either, so we still have to make some guesses around the intents etc 🙃 

> I can see that Blake added `pngnq` and nothing added around the same time seems related.

Like this one being more circumstantial than conclusive evidence. It's very likely correct, but we probably have no way to concretely tell why it was added back then. Even if we were to ask Blake and get a response, memory of small things typically doesn't last long either 😅 That's one of the primary reasons I document decisions heavily - for posterity!

<hr /> 

Btw, I would suggest not _full_ quoting since it can make things harder to read (can imagine all the repetition if everyone did that), and only quoting smaller parts when necessary and eliding when relevant with "[...]" (Discourse has the nice feature of expanding and linking back to quotes, but more manual on GH).
You also might've missed the part I mentioned in [my comment on how to quote](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2897492783) that if you highlight text in the GH comment and _then_ click the quote button, it will only copy the highlighted part. That works similarly to Discourse, though Discourse also has a nice tooltip when highlighting too.
For example, [this prior comment](https://github.com/suttacentral/suttacentral/pull/3578#issuecomment-4051627013) used a full quote immediately after the quoted text, so is unnecessarily repetitive. SC Discourse even has a bot that edits comments and removes the full quote when that specific scenario happens (very nice!)

jhanarato:
> * there is a single old issue that was written up about it for dev docs ([Documentation: Sphinx #809](https://github.com/suttacentral/suttacentral/issues/809)), but it seems to have never been implemented

That seems to be the case. That was submitted in 2018 by one of the Polish developers. I note that JIRA was used at one point in the distant past.

The dependency was introduced by @gradam in commit d1e46ca which otherwise has nothing to do with `sphinx`

It seems that the secondary dependencies are shown in `pyproject.toml` but not in `requirements.txt`. I assume pip decides on their versions unless specified.

Looks good to me!

jhanarato (approved):
Looks good to me.

jhanarato:
Given that this touches `pyproject.toml`, we should merge with `main` now to avoid conflicts with my work.

Thanks @ihongda !

Issue #3581, opened 2026-03-12
Use multi-stage build for `sc-flask` image


jhanarato:
While reviewing PR #3575, improvements were suggested to the build configuration. By using a [multi-stage build](https://docs.docker.com/build/building/multi-stage/) we can use a slim image in production with development dependencies removed.

This will involve updating the `docker compose` scripts and creating separate dockerfiles for development and production. As part of this work we can also address the CI test environment.

See the references below for `uv` and multistage builds:

https://github.com/astral-sh/uv-docker-example

https://docs.astral.sh/uv/guides/integration/docker/

agilgur5:
As I mentioned in https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2885467018 and as the `uv` guide roughly describes, I'd recommend using a [distroless](https://github.com/GoogleContainerTools/distroless) image for production. 

Can build `dev` and `test` on top of `prod`, or can use a different base image and copy over `.venv`, keeping in mind that having close dev/prod parity is ideal

> This will involve updating the `docker compose` scripts and creating separate dockerfiles for development and production

Can use a single multi-stage `Dockerfile` with Compose using the [`build.target`](https://docs.docker.com/reference/compose-file/build/#target) configuration

jhanarato:
> As I mentioned in [#3575 (comment)](https://github.com/suttacentral/suttacentral/pull/3575#discussion_r2885467018) and as the `uv` guide roughly describes, I'd recommend using a [distroless](https://github.com/GoogleContainerTools/distroless) image for production.

The most recent distroless `uv` image would be:

`ghcr.io/astral-sh/uv:latest` or `ghcr.io/astral-sh/uv:0.10.10`

If we go down the distroless path, we'll need to remove this from the Dockerfile:

```docker
RUN apt-get update && apt-get install -y pngnq && apt-get install -y python3-enchant
```

@agilgur5  did mention that enchant isn't used. Looks like it was introduced in 19f4ff74a8ba4555c77122dfc984aa3c6b7f7845 and removed in 19f4ff74a8ba4555c77122dfc984aa3c6b7f7845

We can also remove `pyenchant` from the `pyproject.toml` dependencies.

`pngnq` is a command line utility:

https://github.com/stuart/pngnq

There are no references to the CLI command `pngnq` so I assume this is also safe to delete.

So yes, we can remove the RUN line.

[...]

PR #3585, opened 2026-03-15
Use multi-stage build

jhanarato:
I've never done this before, but it seemed to work and I'd love some feedback.

There are now three stages in the build process, starting with `prod`, then building `test` on `prod` and `dev` on `test`.

The `dev` target is just a placeholder at this point and duplicates `test`.

I'm not using the distroless `uv` as a base just yet.

**Note on branching**

I've created the new branch from `update_build_config`. The process, as I understand it, is to merge PR #3575, then delete the `update_build_config` branch and then the base for this PR will become `main`.

I might have that completely wrong. Let me know, @agilgur5 !

### DIFF
--- a/server/Dockerfile-flask
+++ b/server/Dockerfile-flask
@@ -6,9 +6,6 @@ FROM python:3.11.3
 # Install uv
 COPY --from=ghcr.io/astral-sh/uv:0.9.30 /uv /uvx /bin/
 
-
-RUN apt-get update && apt-get install -y pngnq && apt-get install -y python3-enchant
-
 RUN mkdir -p /opt/sc/sockets
 RUN mkdir -p /opt/sc/sc-flask
 
--- a/server/Dockerfile-flask
+++ b/server/Dockerfile-flask
@@ -1,14 +1,16 @@
-# Using the python image is redundant now that we have a setup with uv.
-# Perhaps we should update to something else such as the astral images:
+# TODO: Use distroless uv image
 # https://docs.astral.sh/uv/guides/integration/docker/
-FROM python:3.11.3
+# Create the production image
+FROM python:3.11.3 AS prod
 
 # Install uv
 COPY --from=ghcr.io/astral-sh/uv:0.9.30 /uv /uvx /bin/
 
+# Create required directories
 RUN mkdir -p /opt/sc/sockets
 RUN mkdir -p /opt/sc/sc-flask
 
+# Application code will be installed here.
 WORKDIR /opt/sc/sc-flask
 
 # Enable bytecode compilation for better start-up performance at the cost of longer build time
@@ -17,7 +19,7 @@ ENV UV_COMPILE_BYTECODE=1
 # Copy from the cache instead of hard linking due to the cache mount
 ENV UV_LINK_MODE=copy
 
-# Omit development dependencies
+# Omit development dependencies for production
 ENV UV_NO_DEV=1
 
 # Copy dependency manifests and install them into a virtual environment
@@ -27,7 +29,6 @@ RUN --mount=type=cache,target=/root/.cache/uv \
     
 # Copy source code and editable install packages into the .venv
 COPY src/ ./src
-COPY tests/ ./tests
 RUN --mount=type=cache,target=/root/.cache/uv \
     uv sync --locked
 
@@ -46,3 +47,19 @@ COPY entrypoint.sh uwsgi.ini ./
 ENV FLASK_APP="sc_flask.wsgi:app"
 
 CMD ["bash", "entrypoint.sh"]
+
+# Create a test environment based on production.
+FROM prod AS test
+
+# Install development dependencies in .venv
+ENV UV_NO_DEV=0
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --locked --no-install-project
+
+# Make our unit and integration tests available
+COPY tests/ ./tests
+
+# Create a dev environment based on test.
+FROM test as dev
+
+# TODO: Install anything extra that dev may require.
--- a/server/Dockerfile-flask
+++ b/server/Dockerfile-flask
@@ -60,6 +60,6 @@ RUN --mount=type=cache,target=/root/.cache/uv \
 COPY tests/ ./tests
 
 # Create a dev environment based on test.
-FROM test as dev
+FROM test AS dev
 
 # TODO: Install anything extra that dev may require.
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -1,5 +1,7 @@
 services:
   sc-flask:
+    build:
+      target: dev
     env_file:
       - server/env/.dev.env
     volumes:
--- a/docker-compose.prod.yml
+++ b/docker-compose.prod.yml
@@ -1,5 +1,7 @@
 services:
   sc-flask:
+    build:
+      target: prod
     restart: always
     env_file:
       - server/env/.prod.env
--- a/docker-compose.test.yml
+++ b/docker-compose.test.yml
@@ -1,5 +1,7 @@
 services:
   sc-flask:
+    build:
+      target: test
     volumes:
       - ./client/:/opt/sc/frontend
     env_file:


xr843 (commented):
Nice work on the multi-stage build, @jhanarato! The `prod → test → dev` layering is a clean pattern — smaller prod images and proper separation of concerns. A few observations:

---

**1. Inconsistent `AS` casing**

```dockerfile
FROM python:3.11.3 AS prod    # uppercase
FROM prod AS test              # uppercase
FROM test as dev               # lowercase ← inconsistent
```

Minor, but Dockerfile best practices use uppercase `AS` throughout. Suggest: `FROM test AS dev`

**2. Missing newline at end of Dockerfile**

The file currently ends without a trailing newline (the diff shows `\ No newline at end of file`). POSIX convention and most linters expect a final newline.

**3. Removal of `pyenchant` and `pngnq` — intentional?**

The PR removes both `apt-get install -y pngnq && apt-get install -y python3-enchant` and the `pyenchant` pip dependency. This seems like a separate concern from the multi-stage restructuring. If these packages are no longer needed, great — but it might be worth confirming they're unused, or splitting this into a separate commit for clarity. If `pyenchant` is still imported somewhere in the codebase, this would break at runtime.

**4. `scripts/` COPY ordering is good**

Moving `COPY scripts/ ./scripts` after the source code and before `entrypoint.sh` makes the layer caching more logical — dependency install layers are cached separately from source code changes. 👍

**5. Moving the "Waiting for arango" print**

Moving the print from `entrypoint.sh` to `wait_for_arango.py` makes sense — keeps the log message closer to the actual wait logic.

---

Overall this is a solid first pass at multi-stage builds. The main thing I'd want confirmed is point #3 (whether `pyenchant`/`pngnq` removal is intentional and safe).

agilgur5:
> The `prod → test → dev` layering is a clean pattern —

@xr843 a lot of your prose and code seem to be AI generated, with no disclosure either (IMO lack of disclosure of AI written goes against right speech and feels impolite given the inauthenticity). Among the many signals that can be found in your comment include: unicode arrows, excessive em dashes, excessive bolding, excessive separators, and excessive sycophancy.

SuttaCentral's [goal](https://discourse.suttacentral.net/t/ai-1-let-s-make-suttacentral-100-ai-free-forever/33374?u=agilgur5) is to be 100% human written, so if you want to contribute here, your comments and code should not be AI generated. Please read the linked forum post for more details.

xr843:
@agilgur5 You're right, and I apologize. I did use AI to help write that review comment and should have disclosed that upfront. I understand SuttaCentral values human-written contributions, and I respect that policy. I'll keep this in mind for any future contributions. Sorry about that, and thank you for pointing it out.

jhanarato:
Hi @xr843, 

Glad to see you're interested in suttacentral. As @agilgur5 pointed out, the policy here is not to use AI generated code and content. If you do want to get involved, try cloning and building the project yourself, then get used to reading the code base. I'm more than happy to answer any questions you might have.

jhanarato:
While I'm happy with this as a first stab at the problem and was hoping for some feedback, there's a lot more work involved so I'll withdraw this for now.

The work I've done on the health check has born fruit and I'll merge it into this branch.

I've also realised that we probably don't want uv installed in the production image and need to give the matter some thought.

server/Dockerfile-flask
```diff
@@ -41,10 +39,27 @@ RUN --mount=type=cache,target=/root/.cache/uv \
...
+# Install development dependencies in .venv
+ENV UV_NO_DEV=0
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --locked --no-install-project
+
+# Make our unit and integration tests available
+COPY tests/ ./tests
+
+# Create a dev environment based on test.
+FROM test as dev
```
agilgur5:
```suggestion
FROM test AS dev
```

nit: consistent casing

server/Dockerfile-flask
```diff
@@ -1,17 +1,16 @@
...
 # https://docs.astral.sh/uv/guides/integration/docker/
-FROM python:3.11.3
+# Create the production image
+FROM python:3.11.3 AS prod
 
 # Install uv
 COPY --from=ghcr.io/astral-sh/uv:0.9.30 /uv /uvx /bin/
 
-
-RUN apt-get update && apt-get install -y pngnq && apt-get install -y python3-enchant
```
agilgur5, line 10: I'd split this into a separate PR, or two, as the concerns are orthogonal. I have a WIP branch for `pyenchant` already too, so you can cover removing `pngnq` and the rest of this line 🙂

agilgur5 (commented):
> While I'm happy with this as a first stab at the problem and was hoping for some feedback, there's a lot more work involved so I'll withdraw this for now.

Sorry, I was out for a bit due to my injection schedule (c.f. [dev chat](https://discourse.suttacentral.net/chat/c/-/12/7080)). I actually do think this is a good first step!

> The work I've done on the health check has born fruit and I'll merge it into this branch.

I actually quite prefer your smaller PR here! Rather than combining concerns with the health check (see the bottom of https://github.com/suttacentral/suttacentral/issues/3581#issuecomment-4102654956). It could be even smaller (see below).

> I've also realised that we probably don't want uv installed in the production image and need to give the matter some thought.

Ideally not for the end prod image, but one step at a time! That's one of the last optimizations I might make and is not as significant as this change or moving to a distroless image. See https://github.com/suttacentral/suttacentral/issues/3581#issuecomment-4102654956

### Specific comments

> 1. **Inconsistent `AS` casing**

> 2. **Missing newline at end of Dockerfile**

I do agree with the LLM on these two mostly stylistic points

> 3. **Removal of `pyenchant` and `pngnq` — intentional?**

Yes, this was intentional. @jhanarato didn't mention the corresponding issue in the description, #3581, which does describe this piece in https://github.com/suttacentral/suttacentral/issues/3581#issuecomment-4059576378, which harkens back to https://github.com/suttacentral/suttacentral/pull/3562#discussion_r2839157388.

> or splitting this into a separate commit for clarity

I do agree with this separation that the LLM suggested. I have a WIP branch which removes `pyenchant` specifically and goes through how it's unused. `pngnq` is a separate story afaict.

That removal is also unrelated to the multi-stage piece too, and technically not necessary for the eventual distroless switch either, as I explained at the top of https://github.com/suttacentral/suttacentral/issues/3581#issuecomment-4102654956.
So they're separate concerns and as such ideally separate PRs.

xr843:
Hi @jhanarato @agilgur5, thank you for the feedback. You're right — I should have disclosed that I used AI assistance in drafting that comment. I apologize for the oversight and will respect SuttaCentral's policy going forward.

I'm genuinely interested in contributing to SuttaCentral. I'll start by cloning the project and reading through the codebase myself. @jhanarato thanks for the welcome — I may reach out with questions as I get familiar with the code.

agilgur5:
> You're right — I should have disclosed that I used AI assistance in drafting that comment. I apologize for the oversight and will respect SuttaCentral's policy going forward.

Given the em dashes, excessive verbosity, excessive sycophancy, and repetition from [the previous comment](https://github.com/suttacentral/suttacentral/pull/3585#issuecomment-4086303534), this seems like it is _still_ an LLM talking and not an original, human written comment...

xr843:
How about we do a video test to see which LLM it is?。。。

agilgur5:
> How about we do a video test to see which LLM it is?。。。

Well this looks human written, but it's notable that the style in this comment is vastly different from all your previous comments. Please only make original, human written comments here. And elsewhere, I'd suggest always disclosing LLM use (similar to the way I cite my sources when applicable).

The usage of 。characters for ellipsis (instead of ...) raises an eyebrow as LLMs are known for using different characters in punctuation, but if you're on a different keyboard, that could also explain it.
If you're using an LLM to _translate_, a translation app would be fine, and you can put the original language in the comment too. No need to use an LLM specifically for that purpose.

server/Dockerfile-flask
jhanarato: > I'd split this into a separate PR, or two, as the concerns are orthogonal. I have a WIP branch for `pyenchant` already too, so you can cover removing `pngnq` and the rest of this line 🙂

Ahh yes, I was conscious of the conflict when I removed it from `pyproject.toml`.

agilgur5: ✅ Removed `pyenchant` separately in #3588.

I'll leave removing `pngnq` to you!

jhanarato:
Oh dear, now the commits are combined here too. Dang!

agilgur5:
Fixed that! But the `AS` casing suggestion got removed when you force pushed [an hour ago](https://github.com/suttacentral/suttacentral/pull/3585#event-23797902211).

Can cherry-pick 42214e5117823603c2a52d00fd8c1ad912f84bcb back on top if you want, or re-apply the suggestion below:

server/Dockerfile-flask
```diff
@@ -48,4 +46,20 @@ COPY entrypoint.sh uwsgi.ini ./
...
+# Install development dependencies in .venv
+ENV UV_NO_DEV=0
+RUN --mount=type=cache,target=/root/.cache/uv \
+    uv sync --locked --no-install-project
+
+# Make our unit and integration tests available
+COPY tests/ ./tests
+
+# Create a dev environment based on test.
+FROM test as dev
```
agilgur5:
```suggestion
FROM test AS dev
```

re-apply https://github.com/suttacentral/suttacentral/pull/3585#discussion_r2969403319

jhanarato: Cherry picked!

agilgur5: I think you forgot to push the change?

jhanarato: Wrong branch! I cherry picked it onto the health check branch.

jhanarato: Interesting that the commits have a different SHA hash

agilgur5: The hash is derived from the contents _plus_ the message, time, the commit it's based on, etc. In this case, we had force pushed the base of the branch etc, so that'll change the hash of everything on top

agilgur5 (approved):
LGTM! Great first step again 🙂

server/Dockerfile-flask
jhanarato: That's fine. When I was checking the branchs for the cherry picked commit I was assuming it would have the same hash. It doesn't, and that's fine.

It also had a terrible commit message. Best not to just accept github's message.

jhanarato:
> LGTM! Great first step again 🙂

A tick of approval? Then you've given your [chanda](https://suttacentral.net/define/chanda?lang=en).

Issue #3586, opened 2026-03-15
Replace `wait_for_arango.py` script with health check


jhanarato:
Instead of running a python script to see if arango is up, we can create a health check in docker compose.

The server will accept a request at this endpoint:

`/_db/suttacentral/_admin/server/availability`

It doesn't require a login, so pretty simple I'd say.

jhanarato:
The docker how-to can be found here:

https://docs.docker.com/compose/how-tos/startup-order/

jhanarato:
I've implemented this now and it looks good. No pull request as of yet.

PR #3590, opened 2026-03-22
Issue 3586 - Control `sc-flask` startup with `sc-arangodb` health check

jhanarato:
I started work on the health check with a test container. This can probably be deleted now as you still need to observe the order of log messages to verify that the test passes. I've left it in to "show my working" but it ends up being the bulk of the PR. 

With the new configuration you can see that `sc-flask` and `sc-arangodb` no longer have interleaved log messages. 

Implementation is as follows:

## ArangoDB

- I've created a new script , `healthcheck.sh` to implement the health check. As far as I know there's not a simple way to do this in one line to put straight into the docker compose configuration.
- `sc-arangodb` now has a dockerfile which copies the health check script into the image.

## Flask Server

- The `wait_for_arango.py` script has been removed.
- The call to `wait_for_arango.py` in the `entrypoint.sh` script has been removed. 

## Docker Compose

- `sc-flask` now starts up after the health check for `sc-arangodb` indicates "healthy"
- `sc-arangodb` now has a build step to include the health check and uses the script to check itself.


jhanarato:
Hmm... now the multi-stage commits are in this PR. 

What did I do wrong?

server/entrypoint.sh
```diff
@@ -1,5 +1,4 @@
 #!/usr/bin/env bash
-uv run scripts/wait_for_arango.py
```
agilgur5, line 2: So this PR doesn't actually need to be stacked on anything - can be based off `main`, just that this specific line would merge conflict with #3575 (`update_build_config`).
And well, it would use the old directory structure which might be confusing, but most changes should merge/rebase on top without issue.

Hopefully we can get HongDa to merge #3575 soon to make things a little easier!
You could also base this (`health_check`) off it (`update_build_config`) instead of `multi_stage_build` as you're not touching the same lines as `multi_stage_build` -- right now, `health_check` and `multi_stage_build` shouldn't merge conflict at all.

arangodb/healthcheck.sh
```diff
@@ -0,0 +1,16 @@
+# Health check script to be used in conjuction with docker compose.
```
agilgur5:
```suggestion
#!/usr/bin/sh
# Health check script to be used in conjunction with docker compose.
```

- add a shebang for explicitness as to the shell
  - c.f. `shellcheck` [`SC2148`](https://www.shellcheck.net/wiki/SC2148)
- fix a typo, missing "n" in "conjuction" -> "conjunction"

arangodb/healthcheck.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# Whilst in the container, run the `arangosh` command and obtain the
+# name of the database
+DBNAME=$(arangosh --server.database $ARANGO_BASE_DB_NAME \
+  --server.username $ARANGO_USER \
+  --server.password $ARANGO_ROOT_PASSWORD \
+  --javascript.execute-string "print(db._name())");
```
agilgur5:
```suggestion
DBNAME=$(arangosh --server.database "$ARANGO_BASE_DB_NAME" \
  --server.username "$ARANGO_USER" \
  --server.password "$ARANGO_ROOT_PASSWORD" \
  --javascript.execute-string "print(db._name())");
```

- per [`SC2148`](https://www.shellcheck.net/wiki/SC2086), double quote variable names to prevent globbing and splitting

agilgur5: I don't know much about ArangoDB, but I also wonder if there's a shorthand to use all those env vars by default

arangodb/healthcheck.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# The exit code is 0 if the database name returned is
+# the same as specified in the environment variable
+if [ $DBNAME = $ARANGO_BASE_DB_NAME ]; then
```
agilgur5:
```suggestion
if [ "$DBNAME" = "$ARANGO_BASE_DB_NAME" ]; then
```
- same as above, per [`SC2148`](https://www.shellcheck.net/wiki/SC2086), double quote variable names to prevent globbing and splitting

agilgur5: Bash also has more specific syntax for this kind of comparison; does the Arango image have `bash` in it, or just `sh`?

(Ideally it'd have neither and be distroless, that being said)

server/arango-tester/tests/test_arango_healthcheck.py
agilgur5, line 1: I wonder if the Python is necessary at all - could just check for a passing healthcheck.

For example, with [a one-liner Docker command](https://stackoverflow.com/a/52620336/3431180).
Or could even just `echo "started"; exit 0` immediately in this service -- it will only output if the healthcheck ran since you have the `condition`

arangodb/Dockerfile-arangodb
agilgur5, line 1: The `arangodb/` dir could be moved into `server` too, since it is only server-side. But it's not the Flask app so 🤷  could go either way

server/arango-tester/Dockerfile-arango-tester
```diff
@@ -0,0 +1,10 @@
+FROM ghcr.io/astral-sh/uv:debian-slim
+
+COPY pyproject.toml README.md uv.lock /
+
+RUN --mount=type=cache,target=/root/.cache/uv \
+    ["uv", "sync", "--locked", "--no-install-project"]
```
agilgur5: ah did you have to use exec form because it's distroless? if so, good experience with a small little test project!

server/arango-tester/pyproject.toml
```diff
@@ -0,0 +1,13 @@
+[project]
+name = "arango-tester"
+version = "0.1.0"
+description = "Unit test for sc-arangodb container health check"
+readme = "README.md"
+requires-python = ">=3.11.3"
+dependencies = [
+    "pytest>=9.0.2",
+    "python-arango>=8.3.0",
```
agilgur5: this is also where `uv` workspaces could be useful; to share the dependencies between this project and the Flask one, and make sure they have the same versions

server/arango-tester/README.md
```diff
@@ -0,0 +1,29 @@
+# Test ArangoDB is healthy
+
+This is a simple test for adding a health check to the `sc-arangodb` docker compose configuration.
+
+Run from the project base directory with:
+
+```
+docker compose -f docker-compose.yml -f server/arango-tester/docker-compose.yml up sc-arangodb sc-arango-tester --build
```
agilgur5:
```suggestion
docker compose -f server/arango-tester/docker-compose.yml up sc-arangodb sc-arango-tester --build
```

So you can omit the first `docker-compose.yml` file if you use the [`include`](https://docs.docker.com/compose/how-tos/multiple-compose-files/include/) configuration. (I mentioned `extends` in https://github.com/suttacentral/suttacentral/issues/3561#issuecomment-3904739827, which is for specific `services`, whereas `include` is for the whole Compose file).

server/arango-tester/README.md
```diff
@@ -0,0 +1,29 @@
...
+```
+
+Running a second time without the `--build` flag will ensure the build time doesn't interfere with the test.
```
agilgur5: `docker compose up` will build by default if not already done, so you could omit the flag entirely.
If it's not yet built, running it twice without the flag will be the same as running it once with the flag and once without.

agilgur5 (commented):
Some small comments below, but this mostly looks good.

> Issue 3586

Also you'll want to specify the issue id, i.e. #3586, in your issue description so that it links the PR to it. Or in this specific case, use the [phrase](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/using-keywords-in-issues-and-pull-requests#linking-a-pull-request-to-an-issue) "Closes #3586" to auto-close when the PR is merged (as I mentioned in https://github.com/suttacentral/suttacentral/issues/3561#issuecomment-3942709455)

jhanarato:
I've had some feedback on the Arango Community Slack:

> Hi [@Jhanarato](https://arangodb-community.slack.com/team/U73GXH8TU), 
there are way better APIs to do a health check. What is exactly what you would like to achieve exactly? There are a ton of APIs that are dedicated to this purpose. 
Is this a single instance or a cluster that you want to check? 
For one there are universal metrics API that give you more than just 0 or 1. _admin/metrics or less verbose ones like _/admin/support-info.

> you can use arangosh to call the apis above : arango.GET('...') and if you use the -deb image you'll find curl too.
also, you can always take an image and expand it to your needs.

Obviously as we're going Distroless, we probably don't want to add curl to ArangoDB. 

However, we could consider removing access to ArangoDB from outside the docker network.

agilgur5:
> > there are way better APIs to do a health check.

Oh duh, right, this is a third-party service, so it should have health checks built-in.
I've been thinking about health checks for the Flask app (which can include its own DB check), which we'd have to write ourselves, but a well integrated third party app should have them built-in since Docker, k8s, etc use them.

I did find these old threads:
- https://github.com/arangodb/arangodb/issues/1506
- https://github.com/arangodb/arangodb-docker/issues/39

those suggest using `/_api/version`. They do mention that auth might be a hassle to configure.

 Idk if possible, but if we wanted to, we could also turn off auth for reads in general, since this DB doesn't have any data that isn't already public (so reads have close to 0 value to any attacker who has achieved some partial network access. mainly a DoS risk, but that can be done in other ways)

The current k8s documentation also uses that endpoint for health checks: https://arangodb.github.io/kube-arangodb/docs/design/health.html

Interestingly that isn't what they suggested in Slack though?

> we could consider removing access to ArangoDB from outside the docker network.

Yea good point. Only NGINX should be accessible outside the docker network, everything else should be internal only.
The ingress could be further restricted too: ArangoDB only accessible via Flask container and Flask only via NGINX (i.e. the ArangoDB container shouldn't be able to talk to other services and NGINX shouldn't talk to ArangoDB directly)

We might be able to use the docker network options to also prevent egress, i.e. ArangoDB should have no outbound connections, Flask only to a few APIs, and NGINX only to the Flask API via the UWSGI socket (Egress limitations help prevent attackers with C2 servers, e.g. data exfiltration, worms, supply chain attacks, etc that rely on communicating back to an attacker's central server. And reduce ability to move laterally in general).

Although Docker seems quite limited in what it can do here natively compared to k8s [`NetworkPolicy`](https://kubernetes.io/docs/concepts/services-networking/network-policies/) (which I am very used to; or rather the more advanced [`CiliumNetworkPolicy`](https://docs.cilium.io/en/latest/security/policy/))

Tangentially related, we should also run Docker in [rootless mode](https://docs.docker.com/engine/security/rootless/), or use `podman` which was designed to be used without root.

We could probably have two whole separate, dedicated issues on hardening the networking and the Docker configuration 😅

arangodb/healthcheck.sh
jhanarato: > Bash also has more specific syntax for this kind of comparison; does the Arango image have `bash` in it, or just `sh`?

Just `sh` I'm afraid.

server/arango-tester/tests/test_arango_healthcheck.py
jhanarato: > I wonder if the Python is necessary at all - could just check for a passing healthcheck.

It's not necessary. As I said [above](https://github.com/suttacentral/suttacentral/pull/3590#issue-4115954861), the whole test is now unnecessary. And given that we still need to inspect the log output it's not a great test.

Remove it from the PR entirely?

server/arango-tester/Dockerfile-arango-tester
jhanarato: > ah did you have to use exec form because it's distroless? if so, good experience with a small little test project!

It's not distroless, no:

```docker
FROM ghcr.io/astral-sh/uv:debian-slim
```

If I recall correctly, I *tried* to make it distroless but was not successful. The issue had to do with running `uv` without `sh` installed. The `RUN` command was probably my attempt to fix the problem.

server/arango-tester/README.md
jhanarato: > `docker compose up` will build by default if not already done, so you could omit the flag entirely.

Interesting. I'd expect this behavior, but my recent work made me think otherwise. See these `Makefile` commands:

```make
run-dev:
	@$(COMPOSEDEV) up $(SERVICES)
	
run-dev-rebuild:
	@$(COMPOSEDEV) up --build --remove-orphans -d $(SERVICES)
```

I've been running the latter to get my changes incorporated in the image.

jhanarato:
I tried this in an interactive session with `arangosh`:

```
127.0.0.1:8529@_system> arango.GET("/_admin/server/availability");
{ 
  "error" : false, 
  "code" : 200, 
  "mode" : "default" 
}
```

You don't need to login to access the endpoint:

http://localhost:8529/_admin/server/availability

But `arangosh` does ask for your password:

```
/ # arangosh --javascript.execute-string "print(\"arango.GET("/_admin/server/availability")\")"
Please specify a password: 
```

server/Dockerfile-flask
jhanarato: This will fail if the, now empty, `scripts/` directory is removed.

jhanarato:
Changed to draft while I sort out the branching situation.

arangodb/Dockerfile-arangodb
agilgur5: Oh, I see that `nginx/` is out front in the [current directory layout](https://github.com/suttacentral/suttacentral/blob/ed222f15709428cbff4f4feba519e9cd24245718/docker-compose.yml#L34) too, rather than inside `server/`. In that case, `arangodb/` makes sense to be parallel to it and Flask in a "services" structure

server/arango-tester/README.md
agilgur5: So `--build` _forces_ a rebuild. If it is not specified, `up` will use any existing image that matches the Compose. It won't check if any files changed though; it's a simple cache of if an image exists or not.

So `--build` is still useful when you definitely want a rebuild and it's not your first time running `up`.

The `watch:` config that we added in #3575 ([Compose file link](https://github.com/suttacentral/suttacentral/pull/3575/changes#diff-9542f82d64bbeebd91f6236324bfe199e9657e2cb1fd9779d5d6dcdcf9cd4de1)), has a more intelligent rebuild where we have specified when that is necessary during a `--watch`.
But that won't run outside of `--watch`, so will still need a `--build` if you want a fully up-to-date image for say, prod.

server/arango-tester/Dockerfile-arango-tester
agilgur5: > It's not distroless, no:

Oh woops. See https://github.com/suttacentral/suttacentral/issues/3581#issuecomment-4188357313 et al for a discussion on distroless Python images

server/arango-tester/tests/test_arango_healthcheck.py
agilgur5: Oh woops, my bad. I recall reading that, but must've forgot once I dug into the code, sorry about that!

> I've left it in to "show my working" but it ends up being the bulk of the PR.

If you want to still "show your work", you could leave that commit in this PR and then just add another commit removing it. Alternatively, could start a new PR without that in the history and close and leave this PR for posterity. I'm fine with either option

agilgur5:
> But `arangosh` does ask for your password:

Curious what happens if you just enter blank or if there is otherwise a way of skipping this for an unathenticated endpoint?
I'm guessing it can't be skipped as that would require a pretty intelligent parser for the CLI to determine ahead of the JS execution that no password is needed (and that would be a prime target for vulnerabilities too, so may not be worth the hassle to maintain such a feature 😅 ), so I wouldn't blame them for not having such a feature

server/entrypoint.sh
jhanarato: #3575 is merged. I may need to do another cherry pick for this. I'll wait for #3602 to be merged first.

agilgur5:
Replaced by #3609. Probably could have force pushed that onto this branch and changed the base to `main` to get something equivalent, but nbd

Issue #3580, opened 2026-03-12
Separate ETL from API container


jhanarato:
It might make sense to move the functionality of the `make load-data` command into a separate container. By separating the concerns of serving API requests and importing data from the git repository, understanding the code will be easier and reduced dependencies will improve security.

I need to confirm that this won't incur to much of a performance hit.

jhanarato:
Now that I think about it, we can probably hive off the `migrate` package and commands. They really should just depend on ArangoDB.

jhanarato:
I've noticed that we have this line in the `sc-flask` `entrypoint.sh` script:

```
touch /tmp/.done.info
```

The `wait_for_flask.sh` script checks this file exists. This could be a health check on the `sc-flask` container.

migrate does not need to wait for flask. In fact it doesn't, it runs before flask starts in the `entrypoint.sh` script.

migrate can run in its own container and start up after `sc-arangodb` and `sc-flask` can run afterwards using `service_completed_successfully` in docker compose.

The front end and nginx containers can wait for `sc-flask` to become healthy. It's quite possible they're serving connections before it comes up.

agilgur5:
In the deployment automation thread I started in dev chat, I [made some suggestions](https://discourse.suttacentral.net/chat/c/-/12/t/172/7109) after realizing that `make load-data` etc all run on the server directly vs in CI/CD:

> Those can be automated a few different ways:
> 
> 1. CI could trigger automation on the server (CD) to run the scripts (and log them). This makes more sense in a cluster set-up, that being said. For example, if one had GitHub Actions (CI) hit a webhook to trigger ArgoCD (CD) on a k8s cluster (which could then trigger a script or Argo Workflows on the cluster, etc). I believe SC already has a [variation of this](https://github.com/suttacentral/suttacentral/blob/ed222f15709428cbff4f4feba519e9cd24245718/deployment_webhook.py#L19)
> 1. In a "blue-green deployment", you'd always be running a 'v2' like you have, and the v2 would run those on start-up. It would only replace the main one if it succeeds; if there are errors, it should stay as v2. (in k8s, a Deployment is by default "rolling" and [init containers](https://kubernetes.io/docs/concepts/workloads/pods/init-containers/) failing will prevent the "rollout" from proceeding)
[...]

PR #3600, opened 2026-03-27
Remove shell dependencies

jhanarato:
Another step towards a distroless production image:

- No more RUN commands that depend on the shell.
- The `entrypoint.sh` script has been rewritten in Python.

Note that we use `subprocess.run` in the new python script. These can be removed by:

- Running `sc_flask.migrate` directly.
- Using the `pyuwsgi` package. 

Changing the CLI invocation to some other form will work in the first case.

The latter works fine, but I decided against modifying `pyproject.toml` and `uv.lock` to avoid the inevitable merge conflict.

### DIFF
--- a/server/Dockerfile-flask
+++ b/server/Dockerfile-flask
@@ -26,12 +26,12 @@ ENV UV_NO_DEV=1
 # Copy dependency manifests and install them into a virtual environment
 COPY pyproject.toml uv.lock README.md ./
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv sync --locked --no-install-project
+    ["/bin/uv", "sync", "--locked", "--no-install-project"]
     
 # Copy source code and editable install packages into the .venv
 COPY src/ ./src
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv sync --locked
+    ["/bin/uv", "sync", "--locked"]
 
 # Make our uv managed .venv the default Python to avoid confusion
 # with the system environment provided by the docker image.
@@ -55,7 +55,7 @@ FROM prod AS test
 # Install development dependencies in .venv
 ENV UV_NO_DEV=0
 RUN --mount=type=cache,target=/root/.cache/uv \
-    uv sync --locked --no-install-project
+    ["/bin/uv", "sync", "--locked", "--no-install-project"]
 
 # Make our unit and integration tests available
 COPY tests/ ./tests
--- a/server/Dockerfile-flask
+++ b/server/Dockerfile-flask
@@ -6,9 +6,10 @@ FROM python:3.11.3 AS prod
 # Install uv
 COPY --from=ghcr.io/astral-sh/uv:0.9.30 /uv /uvx /bin/
 
-# Create required directories
-RUN mkdir -p /opt/sc/sockets
-RUN mkdir -p /opt/sc/sc-flask
+# Create directory to write socket files to.
+# Using `WORKDIR` rather than `RUN mkdir` avoids
+# the need to run a shell command.
+WORKDIR /opt/sc/sockets
 
 # Application code will be installed here.
 WORKDIR /opt/sc/sc-flask
--- a/server/Dockerfile-flask
+++ b/server/Dockerfile-flask
@@ -23,6 +23,9 @@ ENV UV_LINK_MODE=copy
 # Omit development dependencies for production
 ENV UV_NO_DEV=1
 
+# Ensure python stdout/stderr are written to terminal logs
+ENV PYTHONUNBUFFERED=1
+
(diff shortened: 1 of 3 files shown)


server/Dockerfile-flask
```diff
@@ -6,9 +6,10 @@ FROM python:3.11.3 AS prod
 # Install uv
 COPY --from=ghcr.io/astral-sh/uv:0.9.30 /uv /uvx /bin/
 
-# Create required directories
-RUN mkdir -p /opt/sc/sockets
-RUN mkdir -p /opt/sc/sc-flask
+# Create directory to write socket files to.
+# Using `WORKDIR` rather than `RUN mkdir` avoids
```
agilgur5, line 10: `mkdir` is a binary, can check `ls /bin` to see if it exists
So it might be possible to rewrite in exec mode:
```diff
- RUN mkdir -p /opt/sc/sockets
- RUN mkdir -p /opt/sc/sc-flask
+ RUN ["/bin/mkdir", "-p", "/opt/sc/sockets", "/opt/sc/sc-flask"]
```

agilgur5: Oh, right, nevermind, Google's Distroless images don't have anything in `/bin`. Can't even use `ls`, let alone `mkdir` 😅 That's a good thing with regard to attack surface, albeit a bit confusing with regard to GNU compat (no GNU coreutils [package](https://github.com/GoogleContainerTools/distroless/blob/accc36a5c1af3ce5bf7d7b1d75f293a4798c4c03/base/config.bzl#L7) essentially, taking the GNU out of GNU/Linux).
DHI has a `-dev` image variant [with coreutils](https://github.com/docker-hardened-images/catalog/blob/294a4397f5f1b9c387faf457939b670427493a37/image/uv/debian-12/0-dev.yaml#L31) and other packages

But this comment should probably be re-written still as it isn't just about not having a shell, but also no `mkdir` binary at all

server/Dockerfile-flask
```diff
@@ -22,15 +23,18 @@ ENV UV_LINK_MODE=copy
 # Omit development dependencies for production
 ENV UV_NO_DEV=1
 
+# Ensure python stdout/stderr are written to terminal logs
+ENV PYTHONUNBUFFERED=1
```
agilgur5, line 27: I understand why you might want to set this so that logs are written immediately, but disabling buffering is also a performance degradation (as it will increase FS I/O. it can be significant if there is a lot of logging).
Perhaps leave this only in the test and dev images?

I'm also not sure how it's related to the purpose of this PR

server/scripts/entrypoint.py
```diff
@@ -0,0 +1,30 @@
+import subprocess
+from pathlib import Path
+
+def run_migrate():
+    print("Running migrate")
+    # There is surely a way to invoke this directly in python,
```
agilgur5, line 6: Yes, you could `import` it directly.

We may end up removing this from the entrypoint as part of #3580 anyway

server/scripts/entrypoint.py
```diff
@@ -0,0 +1,30 @@
...
+    # but for now, just imitate the previous entrypoint.sh script.
+    subprocess.run(["uv", "run", "python", "-m", "sc_flask.manage", "migrate"])
```
agilgur5, line 8: We probably want to have `check=True` so that this function will raise an error if the subprocess errors. And also want to make sure the logs aren't swallowed. Invoking it directly in Python would avoid the need for those considerations

agilgur5 (commented):
Nice work refactoring out the shell usage.

I left a few comments below on some nuances of `mkdir`, `subprocess`, and `PYTHONUNBUFFERED`

server/Dockerfile-flask
jhanarato: IIRC, I added this so that we see the output of `entrypoint.py` in the logs.

You're right that it should only be in the dev image.

server/scripts/entrypoint.py
jhanarato: > We may end up removing this from the entrypoint as part of https://github.com/suttacentral/suttacentral/issues/3580 anyway

I suspect #3580 is a fair way off, depending on how I divide my time.

We not only have to import it, but also need to make the flask app available to it. I forget how to do this, but it's not hard.

server/scripts/entrypoint.py
jhanarato: > Invoking it directly in Python would avoid the need for those considerations

That's the way then... I'll take a look.

jhanarato:
I'm going to have to turn this into a new PR. Feel free to comment here, or there when I get to it.

agilgur5:
Yea I figure we can discuss a bit here, then move to a new PR

PR #3608, opened 2026-04-05
Removed unused image files code.

jhanarato:
I'm back looking at the ETL code. :tada: 

This is dead code. @ihongda commented out the call to `load_image_links()` in 8c4228ed7f8d6a0669e16ed5a2a5b921f079a9bb.

I removed it and ran my tests without mishap.

```cli
make load-data
make test-load-data
```


jhanarato:
It occurs to me to check if there's a migration needed, such as to remove an unused collection.

jhanarato:
I gather this refers to the images in question:

https://github.com/suttacentral/suttacentral/blob/ed222f15709428cbff4f4feba519e9cd24245718/server/src/data_loader/images_files.py#L12-L15

And we write to the `images` collection.

Oh, and incidentally, we're hitting the legacy server:

https://github.com/suttacentral/suttacentral/blob/ed222f15709428cbff4f4feba519e9cd24245718/server/src/data_loader/images_files.py#L43-L45

Which isn't there anymore!

jhanarato:
This migration creates the `images` collection:

https://github.com/suttacentral/suttacentral/blob/main/server/src/migrations/migrations/add_images_collection_011.py

But there are no more references to it. I can try my hand at a migration to remove it.

PR #3609, opened 2026-04-07
Control sc-flask startup with sc-arangodb health check

jhanarato:
This is essentially the same as PR #3590 and addresses issue #3586.

I've recreated the `issue_3586_health_check` branch with cherry picking. I also squashed the two commits relating to double quoting variable names.

The `arango-tester` has been removed in a final commit.


jhanarato:
I've confirmed that `sc-flask` appears in the logs only after `sc-arangdb` is ready.

arangodb/healthcheck.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/sh
+# Health check script to be used in conjunction with docker compose.
+
+# Whilst in the container, run the `arangosh` command and obtain the
+# name of the database
+DBNAME=$(arangosh --server.database "$ARANGO_BASE_DB_NAME" \
+  --server.username "$ARANGO_USER" \
+  --server.password "$ARANGO_ROOT_PASSWORD" \
+  --javascript.execute-string "print(db._name())");
```
jhanarato, line 9: We discussed changing this to an API call within `arangosh`. However, we might need to parse the JSON returned. Not sure how best to do that.

jhanarato: Oh, of course. I can create a JS script to do this and use `--javascript.execute`

jhanarato: And a shebang is an option:

https://docs.arango.ai/arangodb/3.11/components/tools/arangodb-shell/details/#using-arangosh-via-unix-shebang-mechanisms

jhanarato: I'll experiment

arangodb/healthcheck.sh
```diff
@@ -0,0 +1,17 @@
+#!/usr/bin/sh
+# Health check script to be used in conjunction with docker compose.
+
+# Whilst in the container, run the `arangosh` command and obtain the
+# name of the database
+DBNAME=$(arangosh --server.database "$ARANGO_BASE_DB_NAME" \
+  --server.username "$ARANGO_USER" \
+  --server.password "$ARANGO_ROOT_PASSWORD" \
```
jhanarato, line 8: As for these environment variables, there's an explanation here:

https://docs.arango.ai/arangodb/3.11/components/tools/arangodb-shell/examples/

> By default, arangosh tries to connect to an ArangoDB server running on server localhost on port 8529. It uses the username root and an empty password by default. Additionally, it connects to the default database (_system).

arangodb/healthcheck.sh
agilgur5: > However, we might need to parse the JSON returned. Not sure how best to do that.

> Oh, of course. I can create a JS script to do this and use `--javascript.execute`

You mean the `availability` endpoint from https://github.com/suttacentral/suttacentral/pull/3590#issuecomment-4131625033?

Yes, in JS you should be able to do:
```javascript
JSON.parse(arango.GET("/_admin/server/availability")).code === 200
```

Although I'm not sure if `arango.GET` is synchronous or if the `JSON` built-in is available

agilgur5 (commented):
CI is [failing](https://github.com/suttacentral/suttacentral/actions/runs/24061059802/job/70177234639?pr=3609) due to the healthcheck failing:

```
Container sc-arangodb  Error
dependency failed to start: container sc-arangodb is unhealthy
```

Might need a timeout in CI and make sure the logs check runs even if `make prepare-tests` fails

jhanarato:
> CI is [failing](https://github.com/suttacentral/suttacentral/actions/runs/24061059802/job/70177234639?pr=3609) due to the healthcheck failing:

Oops. I noticed it was being cancelled, but didn't make the connection.

We'll take a look.

jhanarato:
I'm going to wait until #3617 is merged, then take another look.

jhanarato:
I'm now trying to remember what I was doing in April...

So what is happening here. Our `Makefile` looks like this:

https://github.com/suttacentral/suttacentral/blob/fb21916faa1650119e2d302c9cb41e5ad5ea1d14/Makefile#L54-L58

I think we need to adapt `wait_for_flask.sh` to wait for the health check.

jhanarato:
> CI is [failing](https://github.com/suttacentral/suttacentral/actions/runs/24061059802/job/70177234639?pr=3609) due to the healthcheck failing:
> 
> ```
> Container sc-arangodb  Error
> dependency failed to start: container sc-arangodb is unhealthy
> ```
> 
> Might need a timeout in CI and make sure the logs check runs even if `make prepare-tests` fails

I don't see the message above. Just that it times out.

OK, maybe the logs get deleted. Here we go!

https://github.com/suttacentral/suttacentral/actions/runs/27323136344/job/80718428253?pr=3609

jhanarato:
Interesting... 88ebc17963504a165c184778c3223f225effc192

ihongda:
> Interesting... [88ebc17](https://github.com/suttacentral/suttacentral/commit/88ebc17963504a165c184778c3223f225effc192)

Hi @jhanarato 
At the time, this line was likely commented out because front-end tests were not being run in ci.yml.

PR #3607, opened 2026-04-05
Fix test configuration mistakes introduced with build system updates

jhanarato:
This should have been changed when we updated the build config.

PR #3596, opened 2026-03-25
refactor(deps): split Python devDeps into their own group

agilgur5:
## Summary

Use `dependency-groups.dev` in `pyproject.toml` for devDeps and remove them from `dependencies`

~This is currently a draft PR as they may be more to move or to synchronize after all my clean-up PRs (#3591, #3588, #3587, #3578)~
- ✅ Those have all been merged and this has been updated and is ready to review now.

## Details

- `pytest*` deps are only used during dev and not needed during prod
  - c.f. https://docs.astral.sh/uv/concepts/projects/dependencies/#development-dependencies
  - before `uv` / `pyproject.toml` (#3557), in the previous `requirements.txt`, one convention was to split out a `requirements-dev.txt`, but that could result in conflicts and other issues

## Future Work

- [ ] I was thinking that `faker` might be a dev dep too, but it appears to be used in prod code inside of models' [`generate` methods](https://github.com/suttacentral/suttacentral/blob/0c60a09c9bbc1c2716512e23e339fa2d772857da/server/server/common/models.py#L93).
  - Those could be moved into extended test models instead, but they actually aren't used there either and are only used in a [`_generate_models` helper function](https://github.com/suttacentral/suttacentral/blob/0c60a09c9bbc1c2716512e23e339fa2d772857da/server/server/common/utils.py#L63), whose uses seem to be dead code? So possibly all that code and it can be removed entirely?

PR #3610, opened 2026-04-07
feat: add support for 'pdhp' UID in range queries and tests

ihongda:


server/Dockerfile-flask
```diff
@@ -19,16 +20,17 @@ ENV UV_COMPILE_BYTECODE=1
 # Copy from the cache instead of hard linking due to the cache mount
 ENV UV_LINK_MODE=copy
 
-# Omit development dependencies for production
+# Omit development dependencies
 ENV UV_NO_DEV=1
 
 # Copy dependency manifests and install them into a virtual environment
-COPY pyproject.toml uv.lock README.md ./
+COPY pyproject.toml uv.lock README.md conftest.py ./
```
agilgur5, line 27: Per the comment directly above this line, this line is intended for dependency manifests (and anything they depend on, like the README). [`conftest.py`](https://github.com/suttacentral/suttacentral/blob/808e26411a97862fbfbb840bbda63dcbbb9a5d74/server/tests/conftest.py#L3) appears to be a test file, and so does not match

Test files also shouldn't be in the `prod` image, but rather in the `test` image below. The test dependencies (like `pytest`) won't even exist in the prod image after #3596, as they shouldn't be used in prod

server/conftest.py
agilgur5, line 1: Why was [`conftest.py`](https://github.com/suttacentral/suttacentral/blob/808e26411a97862fbfbb840bbda63dcbbb9a5d74/server/tests/conftest.py#L3) moved from `server/tests/` to `server/`? This file does indeed appear to be a test file

agilgur5 (commented):
There seems to be some erroneous changes here to the Dockerfile - it seems like the multi-stage build (#3585 ) was misunderstood. cc @jhanarato

server/Dockerfile-flask
jhanarato: I've addressed this in PR #3607. Thanks for the heads up!

server/conftest.py
jhanarato: As above, resolved in #3607.

Issue #3606, opened 2026-04-04
Add Reference button for Chinese Taisho numbers


ayya-vimala:
#### References in menu

In the "views" menu at the top of each sutta page, there is an option to select the display of a number of references:

<img width="784" height="357" alt="Image" src="https://github.com/user-attachments/assets/4ef7d8b4-d473-4213-bd35-50ea3314496b" />

When an option is clicked these reference numbers show up in the sutta text like this:

<img width="441" height="367" alt="Image" src="https://github.com/user-attachments/assets/f8e75593-c691-4c06-aaac-e735838ef80d" />

And when "main" is clicked you get the SC segment numbers:

<img width="516" height="330" alt="Image" src="https://github.com/user-attachments/assets/fe7ff32c-4a59-4b2f-b173-828f46a71523" />

However, this only works well for Pali texts. For the Chinese texts, it shows the SC segment numbers upon selecting "main" but there is no option to show the Taisho numbers, which is usually how Chinese texts are referred to in articles and how the parallels are shown.

#### Feature request

It would be nice if the Chinese texts also show the Taisho reference numbers and if there is a button in the Reference menu to toggle this.
In Bilara-Data repo these reference numbers exist:

<img width="399" height="363" alt="Image" src="https://github.com/user-attachments/assets/58b3cdd6-3f11-43c0-8c71-4f4cdb60e063" />

agilgur5:
For reference, there is a [related forum post](https://discourse.suttacentral.net/t/references-for-aligned-non-pali-texts/43213?u=agilgur5)

> In Bilara-Data repo these reference numbers exist:

For reference, linking [to the location](https://github.com/suttacentral/bilara-data/blob/da5110bb5537266502931ceb7b6ac18c468abbf6/reference/lzh/sct/sutta/ma/ma1_reference.json#L2) in the screenshot

ayya-vimala:
This is in PR so can be deleted after it has shown up on staging and is okay.

ihongda:
Ven. @ayya-vimala I have made the changes and pushed to `Staging`:

https://staging.suttacentral.net/ma1/en/patton?lang=en&layout=plain&reference=taisho&notes=asterisk&highlight=false&script=latin

<img width="978" height="658" alt="Image" src="https://github.com/user-attachments/assets/30815276-0037-4938-a16d-58290a179bc6" />

ayya-vimala:
Looks perfect @ihongda ! Thanks so much!

PR #3616, opened 2026-04-13
feat: add reference button for Chinese Taisho numbers.

ihongda:
Related issue: https://github.com/suttacentral/suttacentral/issues/3606


jhanarato (approved):
I'm not too familiar with the client, but the server code looks good to me. 

The tests and source files are all in the right place now, so merges should be easier in future.

Happy hacking!

Issue #3613, opened 2026-04-08
Reintegrate tests into CI


jhanarato:
To make everyone's life a happier one, we should re-enable testing in our GitHub CI configuration.

This should be as simple as:

- Fixing any failing tests. 
- Decide what to do with flaky tests. (e.g. `test_changetracker::test_function_source`)
- Decide what to do with slow tests. (e.g. there are a couple of long running tests in `tests/api`).

thesunshade:
If I had a vote, I'd vote for this twice.

jhanarato:
Another consideration:

Do we need the CI to be populated before the tests run?

Ideally we'd have a smaller test dataset for integration tests, with most tests not requiring ArangoDB at all. But first things first!

jhanarato:
CI does not currently populate the database and index for search:

https://github.com/suttacentral/suttacentral/blob/main/.github/workflows/ci.yml

We run `make prepare-tests`, but that gives me 44 failed tests.

The tests introduced in 46224759d2a0008b301c7ed739f4cb5a6d7fc87f and 48ee2060f5a891c7cd5b05e7b0d32c930c669636 seem to require a populated database.

For now we could mark these tests as slow and omit them from CI.

jhanarato:
If I could get a bit of advice, @agilgur5 

I'd like to set things up so that if I push to a specified branch it runs the tests. 

Can I do the following?

- Create a branch `reintegrate_tests_with_ci`
- Run the tests only when I push to this branch?

I'd imagine I'd edit these lines:

https://github.com/suttacentral/suttacentral/blob/debaed23416eb2c2bd2a1f9456f46e98575898b1/.github/workflows/ci.yml#L3-L7

but only on my branch. 

Not sure how it all works.

jhanarato:
I've created a file in `.github/workflows` called `new-ci.yml`. It looks like this:

```yml
name: SuttaCentral CI

on:
  push:
    branches: [ reintegrate_tests_with_ci ]

jobs:
  SuttaCentral-CI:
    runs-on: ubuntu-latest

    steps:
      - name: Checkout repo
        uses: actions/checkout@v3

      - name: Install Docker
        uses: docker/setup-docker-action@v4

      - name: Build
        run: |
          make prepare-tests
          # docker wait sc-frontend

      # split debug logging into separate steps so that it is in its own sections and easier to read and parse
      - name: Debug log - compose logs
        run: |
          docker compose logs
      - name: Debug log - docker ps
[...]

Issue #3485, opened 2025-07-12
Remove dependency of data loader on Flask


jhanarato:
We should be able to run and test the data load process independently of Flask. Without that dependency we no longer need to run it in docker. This has the following benefits:

- Any unit tests that don't rely on ArangoDB can be run in an IDE or CI/CD with no docker containers running.
- Any integration tests that do rely on ArangoDB only need that container running.
- Unit tests and integration tests can be run directly via `pytest` on the development machine or via IDE integration. (I use PyCharm Community Edition which does not support debugging docker containers)

I see that the main dependency is on the configuration data that is used to configure the `arangodb.py` module. Refactoring that will be part of the task.

jhanarato:
Note these methods of `common.arangodb.ArangoDB`:

```py
class ArangoDB:
    ...
    @staticmethod
    def _get_db_from_g_or_none() -> Union[Database, None]:
        """
        Returns db object if present in g object otherwise returns None
        """
        return getattr(g, '_database', None)

    @staticmethod
    def _get_client_from_g_or_none() -> Union[ArangoClient, None]:
        """
        Returns db object if present in g object otherwise returns None
        """
        return getattr(g, '_database_client', None)
    ...
```

We stash these objects from the ArangoDB API into Flask's `g` global object which may or may not improve performance, but definitely increases coupling.

agilgur5:
> We stash these objects from the ArangoDB API into Flask's `g` global object which may or may not improve performance, but definitely increases coupling.

If we just wanted to keep a global around in-memory, we could probably just export a singleton instead of using `g`.

That's usually how I implement globals at least.
Since the DB is read-only, we also don't have to worry about thread-safety and such either.

jhanarato:
> > We stash these objects from the ArangoDB API into Flask's `g` global object which may or may not improve performance, but definitely increases coupling.
> 
> If we just wanted to keep a global around in-memory, we could probably just export a singleton instead of using `g`.

Given the number of times we create a `StandardDatabase` object in the ETL, a singleton would be overkill. I think it just hangs on to an HTTP connection which is why you might want to use it for Flask.

> That's usually how I implement globals at least. Since the DB is read-only, we also don't have to worry about thread-safety and such either.

Even if we introduced parallelism, we just need to make sure the data dependencies are preserved. e.g. Don't update a collection that hasn't been created yet. A dependency graph might work.

jhanarato:
[...]

PR #3617, opened 2026-04-14
Reintegrate tests with CI

jhanarato:
A morning well spent!

All tests run in CI, with the exception of the following:

- API tests
- Some `data_loader` tests I wrote that require ArangoDB.

The first category of tests are excluded in `make test-ci`, whilst the second category are skipped if the `CI` environment variable is set.

I think that running `make migrate` in CI would allow those in the second category to run.

This work relates to issue #3613

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: SuttaCentral CI
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, reintegrate_tests_with_ci ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, reintegrate_tests_with_ci ]
 
 jobs:
   SuttaCentral-CI:
@@ -30,5 +30,5 @@ jobs:
         run: |
           docker ps
 
-      # - name: Test
-      #   run: make test
+      - name: Test
+        run: make test-ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -31,4 +31,4 @@ jobs:
           docker ps
 
       - name: Test
-        run: make test-ci
+        run: make test-server-ci
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: SuttaCentral CI
 
 on:
   push:
-    branches: [ main, reintegrate_tests_with_ci ]
+    branches: [ main ]
   pull_request:
-    branches: [ main, reintegrate_tests_with_ci ]
+    branches: [ main ]
 
 jobs:
   SuttaCentral-CI:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -34,5 +34,5 @@ jobs:
         run: |
           make lint
 
-      # - name: Test
-      #   run: make test
+      - name: Test
+        run: make test-server-ci
--- a/Makefile
+++ b/Makefile
@@ -49,13 +49,19 @@ reload-nginx:
 reload-uwsgi:
 	@docker exec sc-flask uv run uwsgi --reload /tmp/uwsgi.pid
 
+
 # Tests.
+prepare-test-load-data:
+	@docker exec -t sc-flask uv run scripts/prepare_data_loader_tests.py
+
 # Starts containers so that we are ready to run tests in them.
 prepare-tests:
 	@make create-network
 	-@docker compose -f docker-compose.yml -f docker-compose.test.yml up -d
 	@echo "waiting for all services to fully start"
 	@bash wait_for_flask.sh
+	@make prepare-test-load-data
+
 # Run tests
 test:
 	@make test-server
--- a/Makefile
+++ b/Makefile
@@ -69,6 +69,9 @@ test-load-data:
 test-api:
 	docker compose run --entrypoint "python /opt/sc/api-tester/run-tests.py" sc-api-tester
 
+test-ci:
+	@docker exec -t sc-flask uv run pytest -s --ignore tests/api/ tests/
+
 migrate:
 	@docker exec -t sc-flask bash -c "uv run python -m sc_flask.manage migrate"
 
--- a/Makefile
+++ b/Makefile
@@ -70,14 +70,14 @@ test:
 test-server:
 	@docker exec -t sc-flask uv run pytest -s --ignore tests/data_loader/ tests/
 
+test-server-ci:
+	@docker exec -t sc-flask uv run pytest -s --ignore tests/api/ tests/
+
 test-load-data:
 	@docker exec -t sc-flask uv run pytest -v tests/data_loader/
 
 test-api:
 	docker compose run --entrypoint "python /opt/sc/api-tester/run-tests.py" sc-api-tester
-
-test-ci:
-	@docker exec -t sc-flask uv run pytest -s --ignore tests/api/ tests/
 
 migrate:
 	@docker exec -t sc-flask bash -c "uv run python -m sc_flask.manage migrate"
--- a/Makefile
+++ b/Makefile
@@ -51,8 +51,6 @@ reload-uwsgi:
 
 
 # Tests.
-prepare-test-load-data:
-	@docker exec -t sc-flask uv run scripts/prepare_data_loader_tests.py
 
 # Starts containers so that we are ready to run tests in them.
 prepare-tests:
@@ -61,6 +59,9 @@ prepare-tests:
 	@echo "waiting for all services to fully start"
 	@bash wait_for_flask.sh
 	@make prepare-test-load-data
+
+prepare-test-load-data:
+	@docker exec -t sc-flask uv run scripts/prepare_data_loader_tests.py
 
 # Run tests
 test:
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ lint-server: lint-server-deps
 .PHONY: lint-server-deps
 lint-server-deps:
 	@docker exec -t sc-flask uv lock --check
-	@docker exec -t sc-flask uv run fawltydeps
+	@docker exec -t sc-flask uv run fawltydeps --exclude scripts/
 
 migrate:
 	@docker exec -t sc-flask bash -c "uv run python -m sc_flask.manage migrate"
--- a/Makefile
+++ b/Makefile
@@ -49,19 +49,29 @@ reload-nginx:
 reload-uwsgi:
 	@docker exec sc-flask uv run uwsgi --reload /tmp/uwsgi.pid
 
+
 # Tests.
+
 # Starts containers so that we are ready to run tests in them.
 prepare-tests:
 	@make create-network
 	-@docker compose -f docker-compose.yml -f docker-compose.test.yml up -d
 	@echo "waiting for all services to fully start"
 	@bash wait_for_flask.sh
+	@make prepare-test-load-data
+
+prepare-test-load-data:
+	@docker exec -t sc-flask uv run scripts/prepare_data_loader_tests.py
+
 # Run tests
 test:
 	@make test-server
 
 test-server:
 	@docker exec -t sc-flask uv run pytest -s --ignore tests/data_loader/ tests/
+
+test-server-ci:
+	@docker exec -t sc-flask uv run pytest -s --ignore tests/api/ tests/
 
 test-load-data:
 	@docker exec -t sc-flask uv run pytest -v tests/data_loader/
--- a/server/scripts/prepare_data_loader_tests.py
+++ b/server/scripts/prepare_data_loader_tests.py
@@ -1,0 +1,16 @@
+from common.arangodb import get_system_db
+from common.utils import current_app
+from migrations.runner import run_migrations
+
+test_db_name = 'suttacentral_data_load_tests'
+
+app = current_app()
+app.config['ARANGO_DB'] = test_db_name
+
+with app.app_context():
+    system_db = get_system_db()
+
+    if system_db.has_database(test_db_name):
+        system_db.delete_database(test_db_name)
+
+    run_migrations()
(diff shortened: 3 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -2,9 +2,9 @@ name: SuttaCentral CI
 
 on:
   push:
-    branches: [ main ]
+    branches: [ main, reintegrate_tests_with_ci ]
   pull_request:
-    branches: [ main ]
+    branches: [ main, reintegrate_tests_with_ci ]
```
agilgur5: These two additions above should be removed now as they're no longer necessary. As I [mentioned in the issue](https://github.com/suttacentral/suttacentral/issues/3613#issuecomment-4236840509), it's only temporarily needed to add a branch under `push` to test prior to making a PR. Once you have a PR to `main`, the PR event will trigger

agilgur5:
> * API tests

Is this because all of those are long running? Or could this be further segmented later and this is a first step? Wasn't totally sure

> I think that running `make migrate` in CI would allow those in the second category to run.

If necessary, could also populate with test data for a quicker run. Once there is an pre-populated ArangoDB image (part of discussion in #3580), that could also make it quick to run in CI

jhanarato:
> Is this because all of those are long running? Or could this be further segmented later and this is a first step? Wasn't totally sure

The API tests require a populated database. 

> If necessary, could also populate with test data for a quicker run. Once there is an pre-populated ArangoDB image (part of discussion in https://github.com/suttacentral/suttacentral/issues/3580), that could also make it quick to run in CI

I use [fakes](https://github.com/suttacentral/suttacentral/blob/5290acad94a6ab0d070ef1f7afe6ae72b7677a5a/server/tests/data_loader/test_unsegmented.py#L18-L47) and dependency injection for my tests. I'm quite happy with the way it worked out. If you look at the `skipif` decorators, I've only had to add them at the class level in pytest. The tests in those classes specifically test database access. They're integration tests really.

But yes, a minimal data set - perhaps a git repository - would help

jhanarato:
Is it worth adding migrate to `prepare-tests` in this PR, or shall we get this merged first?

jhanarato:
> Is it worth adding migrate to `prepare-tests` in this PR, or shall we get this merged first?

No, let's do that in a separate PR. I think this one is good to go.

server/tests/data_loader/test_changetracker.py
```diff
@@ -38,6 +42,7 @@ def bar():
 
     assert function_source(bar) == '    def bar():\n        pass\n'
 
+@pytest.mark.skipif(os.getenv("CI") != "true", reason="Requires collection to be present.")
```
jhanarato: This is odd... I'm skipping these tests locally. Making this a draft for now.

jhanarato:
So my data loader tests require the `suttacentral_data_load_tests` database to exist and that it is migrated.

I must have removed the docker volume recently, and haven't run my tests since.

This PR is now a work in progress.

jhanarato:
OK, figured it out.

I have a fixture:

https://github.com/suttacentral/suttacentral/blob/5290acad94a6ab0d070ef1f7afe6ae72b7677a5a/server/tests/data_loader/test_arangoload.py#L12-L17

and test:

https://github.com/suttacentral/suttacentral/blob/5290acad94a6ab0d070ef1f7afe6ae72b7677a5a/server/tests/data_loader/test_arangoload.py#L35-L43

Together these create and populate a database called `suttacentral_data_load_tests`.

Incidentally, that's a good example of how to run migrations programmatically.

.github/workflows/ci.yml
jhanarato: I get that now. The `pull_request` branch is what's being merged into.

jhanarato:
Oh nice, my changes worked first time!

Makefile
```diff
@@ -49,13 +49,19 @@ reload-nginx:
 reload-uwsgi:
 	@docker exec sc-flask uv run uwsgi --reload /tmp/uwsgi.pid
 
+
 # Tests.
+prepare-test-load-data:
```
agilgur5: in the current organization of the `Makefile`, this actually should go after `prepare-tests`, same way that `test-server` is after `test`

Makefile
```diff
@@ -69,6 +75,9 @@ test-load-data:
 test-api:
 	docker compose run --entrypoint "python /opt/sc/api-tester/run-tests.py" sc-api-tester
 
+test-ci:
```
agilgur5: I might suggest calling this `test-server-ci` and put it below `test-server` as it's a variant of that for CI.

Also to disambiguate with eventual `test-client`. E.g. `test`, `test-ci`, `test-server`, `test-client`, `test-server-ci`

server/tests/data_loader/test_changetracker.py
```diff
@@ -1,3 +1,7 @@
+import os
+
+import pytest
```
agilgur5: these imports are no longer necessary since you removed the `skipif`s

server/tests/data_loader/test_unsegmented.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import os
```
agilgur5: these imports are no longer necessary since you removed the `skipif`s

agilgur5 (commented):
I haven't looked through the new script yet, but left a few comments on the rest

server/scripts/prepare_data_loader_tests.py
```diff
@@ -0,0 +1,16 @@
...
+
+test_db_name = 'suttacentral_data_load_tests'
+
+app = current_app()
+app.config['ARANGO_DB'] = test_db_name
+
+with app.app_context():
+    system_db = get_system_db()
+
+    if system_db.has_database(test_db_name):
```
jhanarato, line 13: Perhaps we should change this function:

https://github.com/suttacentral/suttacentral/blob/99d739a80e08317336f98b0c81fec8f40a1fc559/server/src/common/arangodb.py#L153-L154

If you run `make delete-database` when the database does not exist it will crash.

agilgur5: Versus returning `null` or something? Try/catch on error is a Python idiom ([EAFP](https://docs.python.org/3/glossary.html#term-EAFP) vs LBYL), so erroring out makes sense to me

Could have a helper func `delete_db_if_exists` (or `ensure_delete_db` or something) that incorporates the check or a try/catch if it's common usage (I'm not sure how often that's necessary in SC's usage). As an example, Django has helpers like [`get_or_create`](https://docs.djangoproject.com/en/6.0/ref/models/querysets/#get-or-create), [`get_object_or_404`](https://docs.djangoproject.com/en/6.0/topics/http/shortcuts/#get-object-or-404), etc as shorthands for common DB access patterns.

jhanarato: Let's leave it as is for now. I'm going to take a close look at the `common.arangodb` module in the near future.

jhanarato:
We have a problem here after I combined the lint & testing changes.

https://github.com/suttacentral/suttacentral/actions/runs/26427145262/job/77792928803?pr=3617

@agilgur5 - can you take a look?

agilgur5:
> https://github.com/suttacentral/suttacentral/actions/runs/26427145262/job/77792928803?pr=3617

Copying the error here for quicker context (and because action logs eventually get pruned [after 90 days](https://docs.github.com/en/organizations/managing-organization-settings/configuring-the-retention-period-for-github-actions-artifacts-and-logs-in-your-organization)):

```
These imports appear to be undeclared dependencies:
- 'common'
- 'migrations'
```

`fawltydeps` recognized the `common` and `migrations` packages elsewhere, so this seems like its `venv` module auto-detection is failing in the `scripts/` directory or something 🤔 
I'm probably gonna need to pull down your branch to debug that. A few hypotheses:
- ~does it need an `__init__.py` in `scripts/`?~
  - _EDIT_: that didn't change anything
- Is it because `scripts/` is not inside of `src/` (but neither is `tests/`)?
  - _EDIT_: kinda - I managed to fix this by setting `base_dir = "src/"`, which isn't mentioned in the [Configuration doc](https://tweag.github.io/FawltyDeps/configuration/), but is in the [Usage](https://tweag.github.io/FawltyDeps/usage/#correctly-identifying-1st-vs-3rd-party-imports) and [FAQ](https://tweag.github.io/FawltyDeps/FAQ/#when-do-i-need-to-use-base-dir) docs. It seems like a more recent option: https://github.com/tweag/FawltyDeps/pull/489. Will open a docs PR to `fawltydeps` to fill in the Configuration page.
    - _EDIT2_: ✅ See https://github.com/tweag/FawltyDeps/pull/511
    - _EDIT2_: ✅ Follow-up in #3657

.github/workflows/ci.yml
```diff
@@ -30,9 +30,8 @@ jobs:
         run: |
           docker ps
 
+      - name: Test
+        run: make test-server-ci
```
agilgur5: I did intend for `lint` to run first. Static analysis normally is run first as it is faster, has lower hanging fruit to fix, could show root cause of test failures, etc (or, in other words, is lower in the test pyramid)

(We could also parallelize these eventually, but that's a bit of a separate story. Partially can be covered with `poe` in #3625 (parallelize across multiple cores/threads), partially can be done as separate jobs in GHA (parallelize across multiple runners))

jhanarato: So this only shows up on this branch because of the merge I did?

I can try reordering the tasks if you like.

jhanarato:
OK, I'm up to speed here now.

Checked out this branch and ran

```cli
make run-dev-rebuild
make prepare-tests
make lint
```

And got the same error.

jhanarato:
With details:

```
suttacentral on  reintegrate_tests_with_ci [$] via 🐍 v3.11.3 (sc-flask) took 3s 
❯ docker exec -t sc-flask uv run fawltydeps --detailed
Bytecode compiled 2650 files in 93ms
ERROR:fawltydeps.extract_imports:Could not parse code from sc-data/html_text/jpn/pli/sutta/mn/split.py: invalid syntax. Perhaps you forgot a comma? (split.py, line 21)

These imports appear to be undeclared dependencies:
- 'common'
    imported at:
      scripts/prepare_data_loader_tests.py:1
      scripts/prepare_data_loader_tests.py:2
- 'migrations'
    imported at:
      scripts/prepare_data_loader_tests.py:3
  ```

**Comment by jhanarato** [comment]
Simply excluding the `scripts/` directory works. 

I've opened #3657 to tackle this. 

Please take a final look and I'll merge.

**Comment by agilgur5** [comment]
> Simply excluding the `scripts/` directory works.
> 
> I've opened #3657 to tackle this.

Ah yea, right, that's a quick workaround. Should prefer to configure in the `pyproject.toml` with comments, rather than [with CLI flags](https://github.com/suttacentral/suttacentral/pull/3617/changes#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R93), but nbd as I've found a proper fix so it'll only be there temporarily.

Thanks for splitting out the issue!

> Please take a final look and I'll merge.

👍

File: server/scripts/prepare_data_loader_tests.py [hunk 9]
```diff
@@ -0,0 +1,16 @@
+from common.arangodb import get_system_db
+from common.utils import current_app
+from migrations.runner import run_migrations
+
+test_db_name = 'suttacentral_data_load_tests'
+
+app = current_app()
+app.config['ARANGO_DB'] = test_db_name
+
+with app.app_context():
```
agilgur5, line 10: ah `app` is the Flask dependency that you described in https://github.com/suttacentral/suttacentral/issues/3485, right?
Or, this uses `get_system_db()` instead of `get_db` , but that calls `get_client()` anyway so is pretty similar 🤔  

Would definitely be good to not need to rely on Flask / `app` or have it coupled that way to the DB!
There seems to be a decent bit of [boilerplate](https://github.com/suttacentral/suttacentral/blob/2ab6f956dff31bffe105b36add8d221f5b7327e7/server/tests/data_loader/test_unsegmented.py#L98) copied around because of that, so removing it should shrink that too.

server/scripts/prepare_data_loader_tests.py
```diff
@@ -0,0 +1,16 @@
+from common.arangodb import get_system_db
+from common.utils import current_app
+from migrations.runner import run_migrations
+
+test_db_name = 'suttacentral_data_load_tests'
```
agilgur5, line 5: I see that the constant `suttacentral_data_load_tests` is already used in [a](https://github.com/suttacentral/suttacentral/blob/2ab6f956dff31bffe105b36add8d221f5b7327e7/server/tests/data_loader/test_arangoload.py#L16) [few](https://github.com/suttacentral/suttacentral/blob/2ab6f956dff31bffe105b36add8d221f5b7327e7/server/tests/data_loader/test_changetracker.py#L44) [places](https://github.com/suttacentral/suttacentral/blob/2ab6f956dff31bffe105b36add8d221f5b7327e7/server/tests/data_loader/test_unsegmented.py#L100). It should probably be refactored into a `constants.py` or similar file at some point if the naming is intentionally identical

agilgur5 (approved):
LGTM - let's get these up and running again!

I left some non-blocking comments/questions below, feel free to merge and address those after

server/scripts/prepare_data_loader_tests.py
jhanarato: I have a minimal fixture that breaks the link:

```py
@pytest.fixture()
def database() -> arango.database.StandardDatabase:
    app = current_app()
    with app.app_context():
        return get_db()  # type: ignore
```

agilgur5: What do you mean by "breaks the link"?
I might be misinterpreting; since `current_app()` and `get_db()` are still used there, it's still dependent on the Flask app, no?

Issue #3486, opened 2025-07-12
Remove function hash checking


jhanarato:
The data loader code contains a class called `ChangeTracker`. This was created early in development to do two things:

- Track changes to files, such as unsegmented HTML files that need to be parsed. (Good! there are about 70,000 of them)
- Track changes to source code. 

The latter code can be removed now that we're not deep in early development.

jhanarato:
There are a bunch of functions involved in tracking source code changes:

https://github.com/suttacentral/suttacentral/blob/d72181ae1c1b7c3cf31e47b9837f671556d64457/server/src/data_loader/change_tracker.py#L71-L133

There is only one use of one of these functions outside the `change_tracker` module:

https://github.com/suttacentral/suttacentral/blob/d72181ae1c1b7c3cf31e47b9837f671556d64457/server/src/data_loader/navigation.py#L25-L41

The `function_hashes` collection is created here:

https://github.com/suttacentral/suttacentral/blob/d72181ae1c1b7c3cf31e47b9837f671556d64457/server/src/migrations/migrations/extend_change_tracker_016.py#L5-L11

and only used within the `change_tracker` module.

jhanarato:
Running the `data_loader` in the debugger (yes, we can use the debugger now!) showed that we do get to line 98 in `navigation.py`. I modified it to assert when the change tracker check returns false:

```py
def add_navigation_docs_and_edges(change_tracker, db, structure_dir, sc_bilara_data_dir):
    tree_dir = structure_dir / 'tree'

    names_files = sorted((sc_bilara_data_dir / 'root' / 'misc' / 'site' / 'name').glob('**/*.json'))
    tree_files = sorted(tree_dir.glob('**/*.json'))
    super_tree_file = tree_dir / 'super-tree.json'
    tree_files.remove(super_tree_file)

    root_languages = languages.process_languages(structure_dir / 'super_root_lang.json', True)
    super_extra_info = process_extra_info_file(structure_dir / 'super_extra_info.json')
    text_extra_info = process_extra_info_file(structure_dir / 'text_extra_info.json')

    if change_tracker.is_any_function_changed(
            [_perform_update_queries, _process_super_tree_file, _process_tree_files, _process_names_files]
    ):
        nav_details_docs = _process_names_files(names_files, root_languages, super_extra_info, text_extra_info)

        nav_details_edges = _process_tree_files(tree_files) + _process_super_tree_file(super_tree_file)

        db['super_nav_details'].truncate()
[...]

PR #3619, opened 2026-04-19
Remove change tracker function hash checks

jhanarato:
The ChangeTracker used to store function hashes and compare them.

The `is_file_new_or_changed()` and `is_any_file_new_or_changed()` methods would check the hashes of the source code, though the method names might not suggest this. There was a flag, `check_calling_functions` that was `True` by default.

Testing showed that the `data_loader.navigation` module used `ChangeTracker.is_any_function_changed()` which could never be true. The `super_nav_details` and `super_nav_details_edges` collections would be populated anyway so I've removed that check.

I've added a migration to remove the `function_hashes` and removed any code relating to the collection, associated methods and their tests.

This completes issue #3486

### DIFF
--- a/server/src/data_loader/change_tracker.py
+++ b/server/src/data_loader/change_tracker.py
@@ -1,9 +1,5 @@
-import gc
-import hashlib
-import inspect
 from collections.abc import Iterable, Iterator
 from pathlib import Path
-from typing import Callable
 
 from arango.database import Database
 
@@ -13,19 +9,7 @@ class ChangeTracker:
         self.base_dir = base_dir
         self.db = db
 
-        self.old_function_hashes = {
-            doc['_key']: doc['hash']
-            for doc in db.aql.execute(
-                '''
-            FOR doc IN function_hashes
-                RETURN doc
-        '''
-            )
-        }
-        self.new_function_hashes = {}
-
         # Extract the mtimes from arangodb
-
         self.old_mtimes = {
             entry['path']: entry['mtime']
             for entry in db.aql.execute(
@@ -37,7 +21,6 @@ class ChangeTracker:
         }
 
         # Get the mtimes from the file system
-
         self.new_mtimes = {}
         for path in base_dir.glob('**/*'):
             if path.is_dir():
@@ -53,33 +36,16 @@ class ChangeTracker:
         print(f'{len(self.changed_or_new)} files to be processed')
         print(f'{len(self.deleted)} files to be deleted')
 
-    def is_file_new_or_changed(self, path: Path, check_calling_function: bool = True) -> bool:
-        if check_calling_function and self.is_function_changed(who_is_calling()):
-            return True
+    def is_file_new_or_changed(self, path: Path) -> bool:
         return str(path.relative_to(self.base_dir)) in self.changed_or_new
 
-    def is_any_file_new_or_changed(self, files: list[Path], check_calling_function: bool = True) -> bool:
-        if check_calling_function and self.is_function_changed(who_is_calling()):
-            return True
-        return any(self.is_file_new_or_changed(file, False) for file in files)
+    def is_any_file_new_or_changed(self, files: list[Path]) -> bool:
+        return any(self.is_file_new_or_changed(file) for file in files)
 
     def changed_files(self, paths: Iterable[Path]) -> Iterator[Path]:
         for path in paths:
-            if self.is_file_new_or_changed(path, check_calling_function=False):
+            if self.is_file_new_or_changed(path):
                 yield path
-
-    def is_any_function_changed(self, functions: Iterable[Callable]) -> bool:
-        return any(self.is_function_changed(function) for function in functions)
-
-    def is_function_changed(self, function: Callable | None) -> bool:
-        key = f'{function.__module__}.{function.__qualname__}'
-        return self.is_thing_changed(key, function)
-
-    def is_thing_changed(self, key: str, thing: Callable) -> bool:
-        function_hash = hashlib.md5(function_source(thing).encode()).hexdigest()
-
-        self.new_function_hashes[key] = function_hash
-        return self.old_function_hashes.get(key) != function_hash
 
     def update_mtimes(self) -> None:
         # Update mtimes in arangodb
@@ -98,36 +64,3 @@ class ChangeTracker:
             ],
             on_duplicate="replace",
         )
-        self.db['function_hashes'].truncate()
-        docs = [{'_key': k, 'hash': v} for k, v in self.new_function_hashes.items()]
-        self.db['function_hashes'].import_bulk_logged(docs, wipe=True)
-
-
-def function_source(function: Callable) -> str:
-    return ''.join(inspect.getsourcelines(function)[0])
-
-
-def who_is_calling(depth: int = 2) -> Callable | None:
-    """Return the calling function at given stack depth
-    
-    adapted from https://stackoverflow.com/a/4506081/4092906
-    """
-    frame_info = inspect.stack()[depth]
-    frame = frame_info.frame
-    code = frame.f_code
-    globs = frame.f_globals
-    functype = type(lambda: 0)
-    funcs = []
-    for func in gc.get_referrers(code):
-        if type(func) is functype and getattr(func, "__code__", None) is code and getattr(func, "__globals__", None) is globs:
-            funcs.append(func)
-            if len(funcs) > 1:
-                return None
-    return funcs[0] if funcs else None
-
-
-def whoami() -> Callable | None:
-    """
-    Returns the function that calls it
-    """
-    return who_is_calling(depth=2)
--- a/server/src/data_loader/navigation.py
+++ b/server/src/data_loader/navigation.py
@@ -10,7 +10,7 @@ from data_loader.extra_info import process_extra_info_file
 from data_loader.util import json_load
 
 
-def add_navigation_docs_and_edges(change_tracker, db, structure_dir, sc_bilara_data_dir):
+def add_navigation_docs_and_edges(db, structure_dir, sc_bilara_data_dir):
     tree_dir = structure_dir / 'tree'
 
     names_files = sorted((sc_bilara_data_dir / 'root' / 'misc' / 'site' / 'name').glob('**/*.json'))
@@ -22,23 +22,20 @@ def add_navigation_docs_and_edges(change_tracker, db, structure_dir, sc_bilara_d
     super_extra_info = process_extra_info_file(structure_dir / 'super_extra_info.json')
     text_extra_info = process_extra_info_file(structure_dir / 'text_extra_info.json')
 
-    if change_tracker.is_any_function_changed(
-            [_perform_update_queries, _process_super_tree_file, _process_tree_files, _process_names_files]
-    ):
-        nav_details_docs = _process_names_files(names_files, root_languages, super_extra_info, text_extra_info)
+    nav_details_docs = _process_names_files(names_files, root_languages, super_extra_info, text_extra_info)
 
-        nav_details_edges = _process_tree_files(tree_files) + _process_super_tree_file(super_tree_file)
+    nav_details_edges = _process_tree_files(tree_files) + _process_super_tree_file(super_tree_file)
 
-        db['super_nav_details'].truncate()
-        db['super_nav_details_edges'].truncate()
-        db['super_nav_details'].import_bulk(nav_details_docs)
-        db['super_nav_details_edges'].import_bulk(
-            nav_details_edges,
-            from_prefix='super_nav_details',
-            to_prefix='super_nav_details',
-        )
+    db['super_nav_details'].truncate()
+    db['super_nav_details_edges'].truncate()
+    db['super_nav_details'].import_bulk(nav_details_docs)
+    db['super_nav_details_edges'].import_bulk(
+        nav_details_edges,
+        from_prefix='super_nav_details',
+        to_prefix='super_nav_details',
+    )
 
-        _perform_update_queries(db)
+    _perform_update_queries(db)
 
 
 def _perform_update_queries(db):
--- a/server/src/migrations/migrations/remove_function_hashes_collection_058.py
+++ b/server/src/migrations/migrations/remove_function_hashes_collection_058.py
@@ -1,0 +1,11 @@
+from common.arangodb import get_db
+from migrations.base import Migration
+
+# Why is every migration called SecondMigration? Copy & paste?
+class SecondMigration(Migration):
+    migration_id = 'remove_function_hashes_collection_058'
+    tasks = ['remove_collections']
+
+    def remove_collections(self):
+        db = get_db()
+        db.delete_collection('function_hashes')
--- a/server/tests/data_loader/test_changetracker.py
+++ b/server/tests/data_loader/test_changetracker.py
@@ -1,137 +1,116 @@
+import pytest
+from arango.database import Database
+
 import common.utils
 from common import arangodb
-from data_loader.change_tracker import whoami, who_is_calling, function_source, ChangeTracker
+from data_loader.change_tracker import ChangeTracker
+
+@pytest.fixture
+def mtimes_db():
+    app_ = common.utils.current_app()
+    app_.config['ARANGO_DB'] = 'suttacentral_data_load_tests'
+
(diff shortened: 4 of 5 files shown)


agilgur5 (commented):
> This completes issue #3486

Reminder to use closing keywords as per https://github.com/suttacentral/suttacentral/issues/3561#issuecomment-3942709455

> ## Remove change tracker

The title seems too broad -- this only removes the function tracking part

> There was a flag, `check_calling_functions` that was `True` by default.

Interesting, if I'm understanding that code correctly, it would invalidate the cache (say there are new changes) if the functions themselves changed - i.e. if the change tracking algorithm itself changed.
That's a surprisingly complicated meta-tracker  👀 

> Testing showed that the `data_loader.navigation` module used `ChangeTracker.is_any_function_changed()` which could never be true.

It might take me a bit to interpret https://github.com/suttacentral/suttacentral/issues/3486#issuecomment-4264641245 et al to understand why that was the case.
I'm also curious if it should be fixed to be used properly - if it could be useful (although removing code is always a lot cleaner! 😅). I guess if it were always `false` for so long, that would suggest it wasn't very useful. I also imagine there might be better ways of doing that or working around that, but I'd have to think about it 🤔 

I only took a quick look through the changes so far, but they make sense to me so far from the original premise

server/src/data_loader/change_tracker.py
```diff
@@ -53,34 +36,17 @@ def __init__(self, base_dir: Path, db: Database):
...
                 yield path
 
-    def is_any_function_changed(self, functions: Iterable[Callable]) -> bool:
-        return any(self.is_function_changed(function) for function in functions)
-
-    def is_function_changed(self, function: Callable | None) -> bool:
-        key = f'{function.__module__}.{function.__qualname__}'
-        return self.is_thing_changed(key, function)
-
-    def is_thing_changed(self, key: str, thing: Callable) -> bool:
```
agilgur5, line 78: > `def is_thing_changed(`

well that's certainly a name 😅

server/src/migrations/migrations/remove_function_hashes_collection_058.py
```diff
@@ -0,0 +1,11 @@
+from common.arangodb import get_db
+from migrations.base import Migration
+
+# Why is every migration called SecondMigration? Copy & paste?
```
agilgur5, line 4: > `#` Why is every migration called SecondMigration? Copy & paste?

Yes. Neither [the code](https://github.com/suttacentral/suttacentral/blob/f1a4c0e1045693907b917e3725ad6e397b294a34/server/src/migrations/runner.py#L50) nor [the docs](https://github.com/suttacentral/suttacentral/blob/f1a4c0e1045693907b917e3725ad6e397b294a34/README.md?plain=1#L134) (or its example) require it to be called that.

It probably could just be called `Migration` (like [in Django](https://docs.djangoproject.com/en/6.0/topics/migrations/#migration-files)) if the base class were named differently.

Should probably remove this temp comment:

```suggestion
```

agilgur5: Separately, migration file names usually start with the number for clear ordering (e.g. `0001_migration_name.py`), rather than end with it 😅

agilgur5 (commented):
> Testing showed that the `data_loader.navigation` module used `ChangeTracker.is_any_function_changed()` which could never be true

I think you have a typo, it should be "could never be `false`" per https://github.com/suttacentral/suttacentral/issues/3486#issuecomment-4264641245, https://github.com/suttacentral/suttacentral/issues/3486#issuecomment-4264660710, and the commit message of https://github.com/suttacentral/suttacentral/commit/49f98e3838a54952e380fec3fb9a176b545a86d3

As in, it always caused a re-import same as #3463
So it would be equivalent to run the code in all cases and remove the checks

<hr />

Separate note, it's (mostly) easier to read this PR when you check the ["Hide whitespace" option](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/reviewing-changes-in-pull-requests/reviewing-proposed-changes-in-a-pull-request#starting-a-review) to ignore indentation-only changes (it doesn't quite know how to easily display the refactor in the test file, that being said)

server/tests/data_loader/test_changetracker.py
```diff
@@ -1,137 +1,116 @@
...
-
-def test_nested_call():
-    def nested_returning_itself():
-        return whoami()
-
-    def nested_returning_caller():
-        return who_is_calling()
+from data_loader.change_tracker import ChangeTracker
+
+@pytest.fixture
```
agilgur5, line 8: nice refactor with this common fixture!

I see you squashed [some commits](https://github.com/suttacentral/suttacentral/issues/3486#event-24644676190) - I think this diff would be easier to interpret if the refactor were split into a separate commit. That way the removal and the refactor are cleanly split and each commit's diff is easy to read in isolation

jhanarato: That's the thing. Originally I had about 20 commits. I thought you preferred a squash?

The problem with squashing is that it's not obvious that many careful steps were taken.

jhanarato: The fixture does reduce a lot of duplication. 👍

jhanarato:
> Interesting, if I'm understanding that code correctly, it would invalidate the cache (say there are new changes) if the functions themselves changed - i.e. if the change tracking algorithm itself changed. That's a surprisingly complicated meta-tracker 👀

It may have been useful early on in development when the data model was in flux. I don't think it serves any useful purpose now.  

> > Testing showed that the `data_loader.navigation` module used `ChangeTracker.is_any_function_changed()` which could never be true.
> 
> It might take me a bit to interpret [#3486 (comment)](https://github.com/suttacentral/suttacentral/issues/3486#issuecomment-4264641245) et al to understand why that was the case. I'm also curious if it should be fixed to be used properly - if it could be useful (although removing code is always a lot cleaner! 😅). I guess if it were always `false` for so long, that would suggest it wasn't very useful. I also imagine there might be better ways of doing that or working around that, but I'd have to think about it 🤔

To be honest, I can't be sure that it never worked. I didn't write a lot of tests to figure out what was happening. This PR simply removes that functionality and the need to test it. What we end up with is something that has reasonable test coverage and works. 

That said, I'm still curious as to why it didn't work before. I'll take some time to investigate further.

jhanarato:
I just created a new branch off main and copied in the new tests. I get two failures:

```
FAILED tests/data_loader/test_changetracker.py::TestIsFileNewOrChanged::test_false_when_not_new_or_changed - AssertionError: assert not True
FAILED tests/data_loader/test_changetracker.py::TestIsAnyFileNewOrChanged::test_false_when_none_new_or_changed - AssertionError: assert not True
```

jhanarato:
I think this test failed intermittently for me:

https://github.com/suttacentral/suttacentral/blob/f1a4c0e1045693907b917e3725ad6e397b294a34/server/tests/data_loader/test_changetracker.py#L35-L39

server/tests/data_loader/test_changetracker.py
agilgur5: Usually I do, specifically since small PRs _usually_ obviate the need for separate commits.
But if there's a particularly separate change (that could be a separate PR but doesn't necessitate one) or if something that is easier to read when separated, then for some exceptions like those, the separation makes sense. For example, if I introduce a linter+formatter and it makes automated changes, I'd split the automated pieces as a separate commit (or stack a PR).

In this case, the 1/20 I was looking for is https://github.com/suttacentral/suttacentral/commit/465f2da6725f22bca7a6b428fe4f2d849d6de382 I believe

> The problem with squashing is that it's not obvious that many careful steps were taken.

I agree on that. Usually when I see lots of commits, they are small iterations for which PR versions (such as in [Graphite](https://graphite.com/docs/pull-request-versions), [GitLab](https://docs.gitlab.com/user/project/merge_requests/versions/), Gerrit, etc) may be a better fit. They don't exist on GH though and are only intended for iteration after the initial draft, not before.

Your commits in this specific case are mostly tiny steps though, so actually a bit different than what I usually see. Tiny atomic changes per commit make sense too. You may have found the smallest possible unit for some of these! 

Perhaps the perspective is different - from a reviewer/reader PoV, for a _small_ PR, I generally want the highlights, rather than very, very tiny incremental steps as you have. 20 is a lot for a _small_ PR. But in this case, 1 might be too little. Middle path?
Highlights, like in the PR description, or clear separations, like in the commits, might be where you made a very specific, intentional design decision or intentionally simplified the diff, etc

Also I totally didn't mean to push my commit style onto you! If you prefer 20, you can do 20. More small ones is probably better than less large ones, although certainly there's a balance when reading through a whole history of commits.

jhanarato:
You can follow along with my investigation here:

https://github.com/suttacentral/suttacentral/tree/change_tracker_investigation

If I set `check_calling_function=False` the tests pass.

jhanarato:
Note that we update the function hashes here:

https://github.com/suttacentral/suttacentral/blob/f1a4c0e1045693907b917e3725ad6e397b294a34/server/src/data_loader/change_tracker.py#L101-L103

jhanarato:
OK, in certain circumstances `is_function_changed()` does return False:

```py
def changeable_function():
    i = 1

class TestIsFunctionChanged:
    def test_function_changed_when_hashes_collection_empty(self, mtimes_db, tmp_path):
        mtimes_db.collection('function_hashes').truncate()
        tracker = ChangeTracker(base_dir=tmp_path, db=mtimes_db)
        assert tracker.is_function_changed(changeable_function)

    def test_function_unchanged(self, mtimes_db, tmp_path):
        mtimes_db.collection('function_hashes').truncate()
        tracker = ChangeTracker(base_dir=tmp_path, db=mtimes_db)
        tracker.is_function_changed(changeable_function)
        tracker.update_mtimes()
        tracker = ChangeTracker(base_dir=tmp_path, db=mtimes_db)
        assert not tracker.is_function_changed(changeable_function)
```

jhanarato:
> I think this test failed intermittently for me:

Yep, that's the one: https://github.com/suttacentral/suttacentral/issues/3613#issue-4224148984

server/src/migrations/migrations/remove_function_hashes_collection_058.py
jhanarato: Python module names must start with a letter or an underscore. There is a way to make it work though.

jhanarato:
Hmm... you know what,`is_function_changed()` does return False when you run the ETL with a populated database.

See my latest test: 1258acb0013b06092296e9747bead895b02e98dc

jhanarato:
Question is, did this fix anything?

https://github.com/suttacentral/suttacentral/pull/3465

jhanarato:
> i.e. if the change tracking algorithm itself changed.

Plan is to switch to `libgit2` and check the differences between commits rather than file modification timestamps. That's unrelated to the function hashing. 

As for keeping data in sync with algorithms: is that not the job of migrations?

jhanarato:
And this passes!

```py
    def test_is_file_new_or_changed_when_file_is_new_and_we_run_twice(self, mtimes_db, tmp_path):
        mtimes_db.collection('function_hashes').truncate()

        new_file = tmp_path / 'new_file.txt'
        new_file.touch()

        # First run of ETL
        tracker = ChangeTracker(base_dir=tmp_path, db=mtimes_db)
        assert tracker.is_file_new_or_changed(path=new_file, check_calling_function=True)
        tracker.update_mtimes()

        # Second run of ETL
        tracker = ChangeTracker(base_dir=tmp_path, db=mtimes_db)
        assert not tracker.is_file_new_or_changed(path=new_file, check_calling_function=True)
```

jhanarato:
Well it works. Is it worth keeping?

jhanarato:
I checked out the commit prior to my refactoring. This is what was actually called:

```py
def load_html_texts(change_tracker: ChangeTracker, data_dir: Path, db: Database, html_dir: Path):
    print('Loading HTML texts')

    force = change_tracker.is_any_function_changed(
        [textdata.TextInfoModel, textdata.ArangoTextInfoModel]
    )

    if force:
        print('This might take a while')
        db['html_text'].truncate()

    with textdata.ArangoTextInfoModel(db=db) as tim:
        for lang_dir in tqdm(html_dir.glob('*')):
            if not lang_dir.is_dir:
                continue
            tim.process_lang_dir(
                lang_dir=lang_dir,
                data_dir=data_dir,
                files_to_process=change_tracker.changed_or_new,
                force=force,
            )
```

The question is whether `force` was ever true, truncating the collection and requiring an update of every HTML file.

jhanarato:
Ahh, this fails:

```py
class TestIsAnyFunctionChanged:
    def test_all_functions_unchanged(self, mtimes_db, tmp_path):
        mtimes_db.collection('function_hashes').truncate()

        # First run of ETL
        tracker = ChangeTracker(base_dir=tmp_path, db=mtimes_db)
        tracker.is_any_function_changed([changeable_function, another_changeable_function])
        tracker.update_mtimes()

        # Second run of ETL
        tracker = ChangeTracker(base_dir=tmp_path, db=mtimes_db)
        assert not tracker.is_any_function_changed([changeable_function, another_changeable_function])
```

What happens is that `any()` returns after the first element that is `True`. We add the first function and then exit. The second time we run the first element is `False` because we've stored it, then we check the second element and get `True` because we exited before adding it. Thus the function will return `True` whenever the list of functions is greater than 1.

jhanarato:
Blake explains his reasoning for creating these function checks here ce1caa9fa5c21b7086e1ecdf373ef199807ac57b

> This is partly for ease of iteration but mainly so there's no need to inform others to do a full rebuild.

jhanarato:
> Thus the function will return `True` whenever the list of functions is greater than 1

Wrong again! 

What I think happens is that when `is_any_function_changed()` is called within the ETL, we'll store the hash of one more function. So given a list of `n` functions, the `n + 1` run of the ETL will cause `is_any_function_changed()` to return False.

So the answer to this:

> Question is, did this fix anything?
> 
> #3465

is, yes, sort of. But the bug would not have been as bad as I thought.

jhanarato:
Here's the fix:

```py
    def is_any_function_changed(self, functions: Iterable[Callable]) -> bool:
        changed = [self.is_function_changed(function) for function in functions]
        return any(changed)
 ```

File: server/src/data_loader/navigation.py [hunk 4]
```diff
@@ -22,23 +22,20 @@ def add_navigation_docs_and_edges(change_tracker, db, structure_dir, sc_bilara_d
     super_extra_info = process_extra_info_file(structure_dir / 'super_extra_info.json')
     text_extra_info = process_extra_info_file(structure_dir / 'text_extra_info.json')
 
-    if change_tracker.is_any_function_changed(
```
jhanarato, line 25: Testing shows that this check might be False after the 5th ETL if the functions don't change. 

I don't see how that would be helpful here.

jhanarato: What happens if the files change after the 4th run of the ETL?

jhanarato:
Further testing shows that passing a class to `is_function_changed()` does seem to work, so @ayya-vimala's code might have worked if not for `any()` short-circuiting.

Using an imported class also seems to work.

ayya-vimala:
Did I create that code? If so, it is at least 8-9 years ago so it might be hopelessly outdated!

jhanarato:
> Did I create that code? If so, it is at least 8-9 years ago so it might be hopelessly outdated!

You did! Not just outdated, but the code you called was broken. I removed it almost a year ago.

All good!

server/src/data_loader/navigation.py
jhanarato: It seems the file change check works once `ChangeTracker` has decided all the functions have not changed.

jhanarato:
I'd like to merge this. Any objections, @agilgur5 ?

server/tests/data_loader/test_changetracker.py
```diff
@@ -1,137 +1,116 @@
+import pytest
+from arango.database import Database
```
agilgur5, line 2: `Database` is an unused import here. I've removed it in #3735

agilgur5 (commented):
I hadn't quite gotten around to reviewing the tests in this PR before HongDa merged it 😅 
There were some lint errors here automatically found in CI by `ruff` after it was merged (#3656).

I'll review Jhanarato's analysis later

PR #3623, opened 2026-05-03
Add tests for `import_bulk_load()` and fix defects

jhanarato:
This is the start of work for issue #3618. 

I've written unit tests for the function and discovered a few issues along the way:

- It checks for `overwrite` parameter in `args` rather than `kwargs`.
- The `kwargs` variable is logged, probably for debugging and then forgotten.
- An attempt was made to use `insert_many()` and then abandoned. Some code remained that was unreachable.

These have been fixed now.

Issue #3618, opened 2026-04-16
Remove `import_bulk_logged` hack  in `common/arangodb` module


jhanarato:
Way back in commit ee639c7e8632a110d49a86b99c2dbea6a4290cc6 Blake added a comment describing issues with the way the `python-arango` package worked:

```
# python-arango import_bulk function is pretty dangerous to use
# in current version unless you set "halt_on_error=True" it silently
# ignores errors. But if you set "halt_on_error=True" you get an
# exception but lose the response body containing the error details
# so you have no idea what went wrong!
# It's pretty much a no-win.
# I expect this will be improved in the library in time but for now
# I'm patching in a more sensible version which always logs errors
# and can be told to raise an exception.
```

In commit 67bf63404d07ed4a51ca4602f5b4f68ddb9a0569 the comment was removed along with `import_bulk_safe()`

In commit a93e77b203558d190e303a822fd3c51256c65ffb `import_bulk_logged()` was added, but without an accompanying comment.

A few more changes were made along these lines. 

Now we have 

https://github.com/suttacentral/suttacentral/blob/99d739a80e08317336f98b0c81fec8f40a1fc559/server/src/common/arangodb.py#L14-L69

I had a conversation on the arango community slack a while back about this. Longer than 90 days, so I don't have access to the conversation now. However there are a couple of things to note:

- Newer versions of the python API may have solved the problem.
- There are two ways to add multiple documents in the REST API: `/_api/batch` and `/_api/import`. 

I want to compare the `python-arango` package versions, see if I can replicate these problems and see if they can are fixed in a newer version. Perhaps changing from batch to import or vice versa will help.

jhanarato:
When blake made commit ee639c7e8632a110d49a86b99c2dbea6a4290cc6, the `python-arango` package was at version `3.12.1`. We're now at `7.5.3`. The latest version on Pypi is 8.3.2.

Let's see if we can reproduce the problems mentioned in the comment:

- `import_bulk` silently ignores errors unless `halt_on_error=True`
- If you set `halt_on_error=True` you get an exception without the error details.

jhanarato:
I've done a deep dive into the issues with `import_bulk`:

https://github.com/jhanarato/arango_bulk_test

jhanarato:
Note:

`import_bulk()` uses the `/_api/import` endpoint while `insert_many()` passes an array of objects to the `/_api/document` endpoint.

jhanarato:
[...]

PR #3626, opened 2026-05-08
Introduce `Collection` wrapper class

jhanarato:
Part of Issue #3618 

- Create a wrapper class for collections
- Update first call site of `import_bulk_load()` to use the new class
- Added a nifty `database()` function that avoids clients of `Collection` knowing about our dependence on Flask.

The latter will likely end up in `common.arangodb` but I'll wait until #3623 is merged.

### DIFF
--- a/server/src/common/collections.py
+++ b/server/src/common/collections.py
@@ -36,3 +36,6 @@ class Collection:
 
     def documents(self) -> Iterator[Any]:
         yield from (doc for doc in self._collection.all())
+
+    def clear(self) -> None:
+        self._collection.truncate()
--- a/server/src/common/collections.py
+++ b/server/src/common/collections.py
@@ -1,0 +1,35 @@
+from collections.abc import Sequence
+from typing import Any
+
+from arango.collection import StandardCollection
+from arango.database import StandardDatabase
+
+from common.arangodb import get_db
+from common.utils import current_app
+
+
+def database() -> StandardDatabase:
+    app = current_app()
+    with app.app_context():
+        return get_db()  # type: ignore
+
+
+class Collection:
+    """ SuttaCentral wrapper class for ArangoDB's collection API.
+    Exposes required functionality whilst hiding the rest.
+    """
+    def __init__(self, name: str):
+        self._collection: StandardCollection = database()[name]
+
+    @property
+    def name(self) -> str:
+        return self._collection.name
+
+    def recreate(self, documents: Sequence[dict[str, Any]]) -> None:
+        self._collection.import_bulk_logged(documents, wipe=True)
+
+    def __len__(self) -> int:
+        return len(self._collection)
+
+    def keys(self):
+        return self._collection.keys()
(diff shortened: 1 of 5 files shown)


jhanarato:
I keep adding stuff to `Collection` as I work through the call sites. This can be a draft until I'm done.

agilgur5:
Copying some discussion from the [dev chat](https://discourse.suttacentral.net/chat/c/-/12/8834) over here for context. My main question was:
> I was going to ask in the first PR what the goal of the `Collection` abstraction was, since it seems like it mainly abstracts the `db` interface, but then `import_bulk_logged` is still used inside it - so it didn't change that bit from the [original issue](https://github.com/suttacentral/suttacentral/issues/3618).

And Jhanarato's response:
> My original plan was to abstract away the different calls to `import_bulk_logged()` within `Collection` and then merge the `import_bulk_logged()` code into it. 
>
> As there were quite a few calls to `import_bulk_logged()` and all the call sites were slightly different, I wanted to write tests for each call site.
>
> That then would have been a massive PR, so I decided on one PR per call site.
>
> I used PyCharm to find all the calls to `import_bulk_logged()`. That would have been quite a few calls, but doable.
>
> About halfway through I realised that our devs eventually stopped using it and just used the `python-arango` `import_bulk()`. 
>
> If I were to refactor every call to `import_bulk()` and use `Collection`, that would be many more PRs. 
> 
> At this point I questioned the benefit of the `Collection` abstraction. It would have allowed dependency injection and the use of a `FakeCollection`. We could avoid hitting the database in our unit tests. But working with the real database is not going to slow down the tests very much, and there's a lot of AQL that skips the `import_bulk()` call.
>
> Plus the tests were becoming boilerplate [...]

I'll link the closed PRs here for back-link reference, since they were closed without a direct explanation in their comments:
https://github.com/suttacentral/suttacentral/pull/3653, https://github.com/suttacentral/suttacentral/pull/3652, https://github.com/suttacentral/suttacentral/pull/3651, https://github.com/suttacentral/suttacentral/pull/3650, https://github.com/suttacentral/suttacentral/pull/3649, https://github.com/suttacentral/suttacentral/pull/3648, https://github.com/suttacentral/suttacentral/pull/3647, https://github.com/suttacentral/suttacentral/pull/3646, https://github.com/suttacentral/suttacentral/pull/3645, https://github.com/suttacentral/suttacentral/pull/3644, https://github.com/suttacentral/suttacentral/pull/3642, https://github.com/suttacentral/suttacentral/pull/3641, https://github.com/suttacentral/suttacentral/pull/3640, https://github.com/suttacentral/suttacentral/pull/3636, https://github.com/suttacentral/suttacentral/pull/3634, https://github.com/suttacentral/suttacentral/pull/3633, https://github.com/suttacentral/suttacentral/pull/3630, https://github.com/suttacentral/suttacentral/pull/3629, https://github.com/suttacentral/suttacentral/pull/3627

jhanarato:
Thanks for the tidy up!

PR #3627, opened 2026-05-08
Use Collection in load_currencies()

jhanarato:
Replaced use of `import_bulk_logged()` and tidied up.

Issue #3618

### DIFF
--- a/server/tests/data_loader/test_currencies.py
+++ b/server/tests/data_loader/test_currencies.py
@@ -3,7 +3,7 @@ from pathlib import Path
 
 import pytest
 
-from common.collections import database, Collection
+from common.collections import Collection
 from data_loader.currencies import load_currencies
 
 AUD = {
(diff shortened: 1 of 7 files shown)


server/tests/data_loader/test_currencies.py
```diff
@@ -0,0 +1,53 @@
...
+
+    return tmp_path
+
+
+class TestLoadCurrencies:
+    def test_adds_three_currencies(self, currency_data):
+        db = database()
+        load_currencies(currency_data)
+        amex = {doc['american_express'] for doc in  db.collection('currencies').all()}
+        assert amex == {True, False, False}
```
jhanarato: Oops. Bad test there - the set of booleans is going to become `{True, False}` I really need to grab the documents and inspect them. A `Collection.documents()` iterator is the way to go.

PR #3632, opened 2026-05-13
Support for testing the `data_loader.arangoload` module

jhanarato:
As I proceed through #3618 in a series of PRs, I need to tackle several call sites for `import_bulk_logged()` in the `arangoload` module. These will be tackled in a series of PRs as before, but there's some commonality to the tests that I've tried to capture here. The PRs for `import_bulk_logged()` call sites in `arangoload`  will be stacked on this one.

I've introduce a ports-and-adapters approach here, which will likely extend to our source data repository and arango collections at least. For now I've just created a protocol class that allows us to fake the `ChangeTracker`.

As I was modifying the imports I also tided up a bit, using absolute imports and otherwise tidying via the PyCharm "Optimise Imports" feature.


jhanarato:
Again, while I'm here I'll fix up some of the PEP-8 issues.

jhanarato:
I've cherry picked 332ccc05fc472c1e42c1271cdfaa420e7dc984e8 into PR #3663. That was the best bit, so I'll close this now.

PR #3635, opened 2026-05-16
Fix: only use parallels.json when generating relationship edges.

jhanarato:
Our ETL started crashing when https://github.com/suttacentral/sc-data/blob/main/relationship/new_parallels.json was added. This fixes the problem. 

We can merge this, or remove the new file.

### DIFF
--- a/server/src/data_loader/arangoload.py
+++ b/server/src/data_loader/arangoload.py
@@ -104,7 +104,7 @@ def get_uid_matcher(db):
 def generate_relationship_edges(
         change_tracker, relationship_dir, additional_info_dir, db
 ):
-    relationship_files = list(relationship_dir.glob('*.json'))
+    relationship_files = [relationship_dir / Path('parallels.json')]
 
     if not change_tracker.is_any_file_new_or_changed(relationship_files):
         return


jhanarato:
My testing was simply:

```
make delete-database
make migrate
make load-data
```

Relevant output from the last command:

```
   30: Generating and loading relationships
Generating Parallels
  0%|                                                                                                                                                                                                                                                                                                                                                      | 0/8124 [00:00<?, ?it/s]ERROR:root:Relationship from uid could not be matched: t765.2 (dropped)
ERROR:root:Relationship from uid could not be matched: t765.1 (dropped)
 66%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████▌                                                                                                               | 5373/8124 [00:00<00:00, 7562.62it/s]ERROR:root:Relationship from uid could not be matched: sa1330##t0367b26 (dropped)
ERROR:root:Relationship from uid could not be matched: t765.83 (dropped)
100%|████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 8124/8124 [00:00<00:00, 11735.44it/s]

**Comment by agilgur5** [comment]
Related to https://github.com/suttacentral/sc-data/issues/368#issuecomment-4458737504

File: server/src/data_loader/arangoload.py [hunk 1]
```diff
@@ -104,7 +104,8 @@ def get_uid_matcher(db):
 def generate_relationship_edges(
         change_tracker, relationship_dir, additional_info_dir, db
 ):
-    relationship_files = list(relationship_dir.glob('*.json'))
+    # A quick fix to avoid matching other files in the directory.
```
agilgur5: Huh, why was this even a glob in the first place? It looks like it was just the one file and now it is just two [in the directory](https://github.com/suttacentral/sc-data/tree/28432a8aba321be8f46d5100579150cc1e518a19/relationship). I would guess to potentially split up into multiple files (over one large one) or to have relationships other than parallels. I didn't check the history too deep to see if there used to be more files in that directory 

I don't think this comment is necessary though since it is temporal rather than explanatory (it does not define what the "fix" is - you'd still need to find the context in the history). I would also think that once the structure etc is finalized, we could just overwrite this code with code to use the new file; no need to have both code paths if it is equivalent data.

```suggestion
```

agilgur5 (approved):
I have a small suggestion regarding the comment, but LGTM otherwise

server/src/data_loader/arangoload.py
jhanarato: I agree. It's a message for the commit, not the code.

agilgur5: Yes exactly, that's what I was trying to explain 😅

agilgur5: > I didn't check the history too deep to see if there used to be more files in that directory

Went through the history of `sc-data` and it's been a single file [since the directory has existed ~9 years ago](https://github.com/suttacentral/sc-data/tree/b4458540be372c50105b31df71091f7360311601/relationship), which is near the beginning of that entire repo's history.

So I can't see why it needed to be a glob pattern in the first place.

In the ETL history in this repo, it looks like it was a single file load too, but then was changed to a glob ~9 years ago in https://github.com/suttacentral/suttacentral/commit/40f7d943d09959421abc83f02c96f80c8e5c8768. Maybe whatever change was intended for multiple files there was never fully implemented 🤷

ayya-vimala:
My apologies ... I was still used to files only being used when called. I realize things have changed a bit on SC now. The new_parallels.json is a test project and I put it there so everyone can see it. This will be the new structure (or something like it) for the parallels.json file but it isn't being used as of yet.

PR #3641, opened 2026-05-22
Process difficulty uses collection

jhanarato:
Wrote tests and refactored as part of issue #3618


jhanarato:
Also note that [our difficulties file](https://github.com/suttacentral/sc-data/blob/main/additional-info/difficulties.json) only contains levels for DN and MN suttas and was last updated 9 years ago.

Issue #3625, opened 2026-05-07
Replace server-side `make` targets with Python-native tasks via `poe`


agilgur5:
## Summary

Replace server-side `make` targets -- i.e. those that are primarily `uv run` -- with Python-native tasks via [`poe`](https://github.com/nat-n/poethepoet)

## Motivation

- `poe` tasks can be configured inside the `pyproject.toml`
  - so devDeps and their corresponding tasks would all be co-located
    - as would any server-side tasks that have dedicated scripts 
  - This would match the client-side `package.json#scripts`
  - It's also easier to configure for task purposes (i.e. it isn't an entire build system like `make`, and is much more ergonomic due to being simpler and newer)
    - For example, I mentioned `poe` for configuring parallel lint checks in https://github.com/suttacentral/suttacentral/pull/3597#discussion_r2989874098.
       - Parallel tasks can be done in `make`, but is easier said than done (it is more all-or-nothing style, see also some prior work I did in partial parallel/sequential tasks in https://github.com/argoproj/argo-workflows/pull/12583)
- `poe` tasks are Python-native, so you can run them in a native Python (not containerized) environment with plain `uv run`
  - this builds on top of prior work to modernize the Python tooling (#3557) and make the Python env work natively without Docker 

## Implementation

With `poe`, we could just shorten the server-side `make` targets to something like:
```makefile
server:
  @docker exec -t sc-flask uv run $@
```

In this example, `make server test` would be equivalent to `uv run test`.

Task dependencies (if any) and specific task configuration could then be moved to the `pyproject.toml`.

Containerized Python tasks run via `make`, which runs `docker exec` + `uv run`, while native Python would run with plain `uv run`

jhanarato:
Sounds good to me!

PR #3597, opened 2026-03-25
ci: lint Python dependencies

agilgur5:
## Summary

Add a CI/lint check for unused + undeclared Python dependencies with [`fawltydeps`](https://github.com/tweag/FawltyDeps) to prevent the problem of unused and undeclared deps from popping up again
- also add a check for synced lockfile with `uv lock --check`

 This is what I've been using to double-check my work during dep clean-up (#3591, #3589, #3588, #3587, #3578) 

~This is a draft PR for now as it relies on there being no extraneous deps for the CI check to pass (it is currently _correctly_ [failing](https://github.com/suttacentral/suttacentral/actions/runs/23522595329/job/68468818206?pr=3597)). There are also some dependency conflicts with some unused and transitive deps, so it will merge conflict with those PRs too.~
- ✅ Those PRs have been merged and this rebased on top, so this now ready to review. ~Will merge conflict with #3596 right now though~

## Details

- add `fawltydeps` as a devDep in `pyproject.toml`
  - configure ignores for deps like CLIs (like `uwsgi`, `fawltydeps` itself) that are not directly imported in Python code
    - c.f. https://github.com/tweag/FawltyDeps/blob/6b57917cc4824ebae253303dd004d990eb569ba4/docs/configuration.md
  - configure ignore for self-reference of our `sc_flask` package

- add `make lint-server-deps` command that runs `fawltydeps` and `uv lock --check`
  - and placeholders for `make lint` and `make lint-server`
    - (i.e. to include `make lint-client` (#3572) and `ruff`)
- add `make lint` to CI

## Verification

The new CI check passes

PR #3656, opened 2026-05-26
ci: configure Python linting with `ruff`

agilgur5:
## Summary

Add `ruff` as a dev dep to lint ~and format~ Python code
  - chose `ruff` as it's backed by [the same org as `uv`](https://github.com/astral-sh/ruff/blob/f2fb134161ed5842388b46e65de4435117601880/README.md?plain=1#L60) and is very performant
    - literally had results instantly on a fresh run

## Details

- use a bit more than default config for now
- add a lint check in the `Makefile` for now

**ci: auto-fix Python code via `ruff`**
- run `uv run ruff check --fix` and commit the changes
  - mostly unused imports and unused `f`-strings (were just plain strings)
- add a `lint-server-fix` target to the `Makefile`

**ci: manually fix remaining Python lint errors**
- mostly unused imports and dead code
  - plus a missing import, an accidentally duplicated class/test name, etc

**ci: configure additional lint rules for ruff**
- add a structure for adding more rules
- add two simple rules for now that the codebase is already compliant with:
  - check for stray [debugger statements](https://docs.astral.sh/ruff/rules/debugger/)
  - check for shadowed built-ins ([example](https://docs.astral.sh/ruff/rules/builtin-variable-shadowing/))

## Verification

- CI passes; `uv run ruff check` exits 0 with no errors

## Future Work

- [ ] add [more lint rules](https://docs.astral.sh/ruff/rules)
  - partly done [in a separate branch](https://github.com/agilgur5/suttacentral/commit/63e35eafe4735df510a6fa2d658645b7c171d108) as there's a lot to check and fix
    - those all require changes to the source code, so will be better as individual PRs 
- [ ] add [formatting](https://docs.astral.sh/ruff/formatter/)
  - partly done [in a separate branch](https://github.com/agilgur5/suttacentral/commit/eb7e97dd3251bd484c1b3849ad88081725ba63c5) as there are a lot of automated changes

### DIFF
--- a/server/src/sc_flask/manage.py
+++ b/server/src/sc_flask/manage.py
@@ -37,7 +37,6 @@ def list_routes():
 
     output = []
     for rule in app.url_map.iter_rules():
-        options = {arg: "[{0}]".format(arg) for arg in rule.arguments}
         methods = ','.join(rule.methods)
         line = urllib.parse.unquote("{:50s} {:20s} {}".format(rule.endpoint, methods, rule))
         output.append(line)
--- a/server/src/search/texts.py
+++ b/server/src/search/texts.py
@@ -340,16 +340,6 @@ def import_texts_to_arangodb():
     )
     languages = list(db.aql.execute(query))
 
-    order = [
-        "en", "pli", "lzh", "san", "pra", "xct", "pgd", "de", "zh", "af",
-        "ar", "bn", "ca", "cs", "es", "fa", "fi", "fr", "gu",
-        "haw", "he", "hi", "hr", "hu", "id", "it", "jpn", "kan", "kho",
-        "ko", "la", "lt", "mr", "my", "nl", "no", "pl", "pt",
-        "ro", "ru", "si", "sk", "sl", "sld", "sr", "sv", "ta", "th",
-        "uig", "vi", "xto"
-    ]
-    # languages = sorted(languages, key=lambda x: order.index(x["uid"]))
-
     for lang in tqdm(languages):
         loader = TextLoader(lang)
         loader.import_all_text_to_db()
--- a/server/src/search/texts.py
+++ b/server/src/search/texts.py
@@ -7,8 +7,6 @@ from tqdm import tqdm
 
 from common.arangodb import get_db
 from common.queries import (
-    CURRENT_MTIMES,
-    CURRENT_BILARA_MTIMES,
     TEXTS_BY_LANG_FOR_SEARCH,
     BILARA_TEXT_BY_LANG_FOR_SEARCH,
     TEXT_REFERENCES
--- a/Makefile
+++ b/Makefile
@@ -85,6 +85,7 @@ lint: lint-server
 
 .PHONY: lint-server
 lint-server: lint-server-deps
+	@docker exec -t sc-flask uv run ruff check
 
 # check for a synced lockfile and no unused or undeclared deps
 .PHONY: lint-server-deps
--- a/Makefile
+++ b/Makefile
@@ -87,6 +87,10 @@ lint: lint-server
 lint-server: lint-server-deps
 	@docker exec -t sc-flask uv run ruff check
 
+.PHONY: lint-server-fix
+lint-server-fix:
+	@docker exec -t sc-flask uv run ruff check --fix
+
 # check for a synced lockfile and no unused or undeclared deps
 .PHONY: lint-server-deps
 lint-server-deps:
--- a/README.md
+++ b/README.md
@@ -168,7 +168,8 @@ class InitialMigration(Migration):
 2. `uv run python -m sc_flask.manage list_routes` - Lists all available routes/URLs.
 
 ### 1.10 Style guidelines
-* Follow [PEP8](https://www.python.org/dev/peps/pep-0008/) for Python code.
+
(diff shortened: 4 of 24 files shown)


agilgur5:
Heh, the dep linter I set-up in #3597 is [already showing some results](https://github.com/suttacentral/suttacentral/actions/runs/26457968413/job/77897343727?pr=3656):
```
These dependencies appear to be unused (i.e. not imported):
- 'ruff'
```

Need to add an ignore for that as it is a CLI

server/src/sc_flask/manage.py
```diff
@@ -37,7 +37,6 @@ def list_routes():
 
     output = []
     for rule in app.url_map.iter_rules():
-        options = {arg: "[{0}]".format(arg) for arg in rule.arguments}
```
agilgur5, line 40: `options` is unused, but I hadn't checked the history for why that is (i.e. if its usage were removed in an earlier commit but the variable was forgotten about)

server/src/search/texts.py
```diff
@@ -342,14 +340,6 @@ def import_texts_to_arangodb():
...
 
-    order = [
-        "en", "pli", "lzh", "san", "pra", "xct", "pgd", "de", "zh", "af",
-        "ar", "bn", "ca", "cs", "es", "fa", "fi", "fr", "gu",
-        "haw", "he", "hi", "hr", "hu", "id", "it", "jpn", "kan", "kho",
-        "ko", "la", "lt", "mr", "my", "nl", "no", "pl", "pt",
-        "ro", "ru", "si", "sk", "sl", "sld", "sr", "sv", "ta", "th",
-        "uig", "vi", "xto"
-    ]
     # languages = sorted(languages, key=lambda x: order.index(x["uid"]))
```
agilgur5: `order` was only used by the commented out `languages` variable here.
should probably remove the commented out `languages` variable as well

I'm guessing the sorting was no longer necessary / not necessary / the DB can do it?

agilgur5 (commented):
left some small notes on lines I looked at for a bit. the rest were relatively straightforward changes, but these raised an eyebrow

jhanarato:
> Heh, the dep linter I set-up in #3597 is [already showing some results](https://github.com/suttacentral/suttacentral/actions/runs/26457968413/job/77897343727?pr=3656):
> 
> ```
> These dependencies appear to be unused (i.e. not imported):
> - 'ruff'
> ```
> 
> Need to add an ignore for that as it is a CLI

I had the same problem with `coverage`. I ended up removing it to pass CI

server/src/sc_flask/manage.py
jhanarato: I notice the line above:

https://github.com/suttacentral/suttacentral/blob/d3392912edf65e28a45181a0b0335ecdc7fb66fd/server/src/sc_flask/manage.py#L36

Why is this import not at the top of the file? It certainly confuses PyCharm:

> Cannot find reference 'parse' in 'urllib'

Looking at the history, this function was updated many times without removing the unused `options`.

Going all the way back... it gets introduced here ad41cec3521f1e244c755c4e99298e36c481ed17

Even then it was unused. This is the same dev that wrote the parallels code - fairly typical for them I'm afraid.

server/src/search/texts.py
jhanarato: Apparently the DB does it. Commenting out `languages` was the last commit made to this file, before I moved it anyway.

server/src/search/texts.py
```diff
@@ -342,14 +340,6 @@ def import_texts_to_arangodb():
     )
     languages = list(db.aql.execute(query))
```
jhanarato, line 341: This is the way we get the languages now.

server/src/sc_flask/manage.py
agilgur5: > Why is this import not at the top of the file?

Most of the functions do dynamic imports in the `manage.py` file. Given that it is used as a CLI, I would guess the intent was to reduce the runtime imports (& therefore start-up time) when using an unrelated function.
When used as a server, you'll end up importing most things eventually anyway, but not quite so as a CLI.
But that may well be a premature optimization; I'm not sure how big a difference it makes. Some of the CLI usage might also be replaceable with `poe` (#3625)

The other common usage of dynamic imports is to help with cyclic imports, but I'm not sure that's relevant in this case.

> It certainly confuses PyCharm:
> 
> ```
> Cannot find reference 'parse' in 'urllib'
> ```

I'm not sure that's related to the location of the import. Python can do dynamic imports, so PyCharm should be able to understand it.
It might actually be old usage from a different version of `urllib` or something? Not sure when this function was last run

server/src/search/texts.py
agilgur5: > Apparently the DB does it.

See @jhanarato's [later comment](https://github.com/suttacentral/suttacentral/pull/3656#discussion_r3385095918) on the AQL query

agilgur5: > should probably remove the commented out `languages` variable as well

I've removed this in the [latest version of this commit](https://github.com/suttacentral/suttacentral/pull/3656/commits/eac3dc4e074f1bf254c89a8faab4641f30557c01)

ihongda (approved):

PR #3661, opened 2026-06-03
refactor(deps): use SemVer ranges in `pyproject.toml`

agilgur5:
## Summary

Use SemVer ranges for all Python deps where possible

This is a follow-up to https://github.com/suttacentral/suttacentral/pull/3557#discussion_r2839214522

## Details

- updating a dependency to a SemVer patch or minor version _should_ be non-breaking, so we don't have to pin to a specific patch version
  - now that we have a lockfile (`uv.lock`), all versions are _still_ pinned to a patch
  - this just changes the "allowable" versions of deps and makes it easier to auto-update within the range. for example:
    - if there's a CVE in a dep, we can auto-update the patch seamlessly now
    - if we need to bump a dep due to another dep requirement, it can be automatically bumped in the lockfile if it is within range, without changing the main manifest
      - `uv add` etc will do that automatically, whereas before it would error, since it had to match the version _exactly_, not just be SemVer-compatible
    - etc
    
## Notes to Reviewers

- `regex` cannot use a SemVer range as it uses CalVer, not SemVer, and so doesn't have a notion of compatibility between versions
- Python itself shouldn't use any minor version because, per [PEP 387](https://peps.python.org/pep-0387/), it doesn't strictly follow SemVer and can have breaking changes within a minor version after a deprecation cycle

## Verification

Double-checked that these ranges would be compatible by hand for some cases with https://nok.github.io/pip-dev/ / https://github.com/nok/pip-dev (an interactive version specifier visualization)

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -3,43 +3,43 @@ name = "sc-flask"
 version = "0.1.0"
 description = "The backend python implementation for SuttaCentral.net"
 readme = "README.md"
-requires-python = "==3.11.3"
+requires-python = "~=3.11.3"
 dependencies = [
-    "aksharamukha==2.3",
-    "algoliasearch==3.0.0",
-    "beautifulsoup4==4.11.1",
-    "click==8.1.3",
-    "decorator==5.1.1",
-    "faker==15.3.4",
-    "flasgger==0.9.5",
-    "flask==2.2.5",
-    "flask-caching==2.0.1",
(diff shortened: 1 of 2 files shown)


jhanarato (approved):
I'm happy to have the dependencies bumped by `uv`. I guess this means we'll rely less on Dependabot.

I've rebuilt and run my smoke tests. CI can take care of the unit tests.

Issue #3655, opened 2026-05-26
Loading code for new_parallels.json


ayya-vimala:
This is a suggested loading code for the new file for parallels: sc-data/new_parallels.json
The loading code has to replace the function of the same name in arangoload.py. When it is decided to use this json format, please test the code thoroughly because I can't actually test it myself because I have no access to the database and cannot run it locally.

```py
def generate_relationship_edges(
        change_tracker, relationship_dir, additional_info_dir, db
):
    relationship_file = relationship_dir / Path('new_parallels.json')

    if not change_tracker.is_any_file_new_or_changed([relationship_file]):
        return

    print('Generating Parallels')
    relationship_data = json_load(relationship_file)

    uid_matcher = get_uid_matcher(db)

    remarks_data = json_load(additional_info_dir / 'notes.json')

    remarks = defaultdict(dict)
    for remark in remarks_data:
        uids = remark['relations']
        remark_text = remark['remark']
        remarks[frozenset(uids)] = remark_text

    def process_uids(from_uid, to_uids, r_type, is_resembling, ll_edges):
        """Generate edges from one UID to a list of target UIDs."""
        m = regex.search('[0-9]+$', from_uid)
        from_nr = int(m[0]) if m else 0

        true_from_uids = uid_matcher.get_matching_uids(from_uid)
        if not true_from_uids and ' ' not in from_uid:
            logging.error(f'Relationship from uid could not be matched: {from_uid} (dropped)')
            return

        for to_uid in to_uids:
            if to_uid == from_uid:
                continue
            true_to_uids = uid_matcher.get_matching_uids(to_uid)
            if not true_to_uids:
                logging.info(f'Relationship to uid could not be matched: {to_uid} (appears as orphan)')
                true_to_uids = ['orphan']

            for true_from_uid in true_from_uids:
                for true_to_uid in true_to_uids:
                    remark = remarks.get(frozenset([true_from_uid, true_to_uid]), None)
                    ll_edges.append({
                        '_from': true_from_uid,
                        '_to': true_to_uid,
                        'from': from_uid,
                        'number': from_nr,
                        'to': to_uid.split('-')[0],
                        'type': r_type,
                        'resembling': is_resembling,
                        'remark': remark,
                    })

    def process_retell_mention(first_uid, other_uids, r_type, ll_edges):
        """Generate bidirectional edges for retells and mentions."""
        m = regex.search('[0-9]+$', first_uid)
        from_nr = int(m[0]) if m else 0

[...]

PR #3663, opened 2026-06-09
Refactor relationships loading algorithm

jhanarato:
After @ayya-vimala dropped some code into issue #3655, I decided to have a go at refactoring the algorithm that creates the edges in the `relationship` collection. Along the way I found a few bugs, and given how radically the code has changed, there's a chance that I've introduced one or two.

That said, I wrote extensive tests before starting the refactoring, wrote several more along the way, and tidied them up at the end of the refactoring.

It was a lot of fun, and somewhat exhausting.

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -52,5 +52,9 @@ ignore_unused = [
     "uwsgi", # used as CLI in `Makefile`, `entrypoint.sh`
 ]
 
+ignore_undeclared = [
+    "fakes" # is declared outside a package in data_loader tests
+]
+
 [tool.pytest.ini_options]
 testpaths = ["tests"]
--- a/server/src/data_loader/relationships.py
+++ b/server/src/data_loader/relationships.py
@@ -1,0 +1,164 @@
+import logging
+from collections import defaultdict
+from itertools import product
+from pathlib import Path
+
+import regex
+from arango.database import StandardDatabase
+from tqdm import tqdm
+
+from common.uid_matcher import UidMatcher
+from common.utils import chunks
+from data_loader.ports import FileChangeTracker
+from data_loader.util import json_load
+
+
+def get_uid_matcher(db: StandardDatabase) -> UidMatcher:
+    all_uids = set(
+        db.aql.execute(
+            '''
+            FOR doc IN super_nav_details
+                RETURN doc.uid
+            '''
+        )
+    )
+
+    return UidMatcher(all_uids)
+
+
+def generate_relationship_edges(
+        change_tracker: FileChangeTracker, relationship_dir: Path, additional_info_dir: Path, db: StandardDatabase
+) -> None:
+    relationship_files = [relationship_dir / Path('parallels.json')]
+
+    if not change_tracker.is_any_file_new_or_changed(relationship_files):
+        return
+
+    print('Generating Parallels')
+    relationship_data = []
+    for relationship_file in relationship_files:
+        relationship_data.extend(json_load(relationship_file))
+
+    uid_matcher = get_uid_matcher(db)
+
+    remarks_data = json_load(additional_info_dir / 'notes.json')
+
+    remarks = defaultdict(dict)
+
+    for remark in remarks_data:
+        uids = remark['relations']
+        remark_text = remark['remark']
+        remarks[frozenset(uids)] = remark_text
+
+    ll_edges = []
+    for entry in tqdm(relationship_data):
+        entry.pop('remarks', None)
+        for r_type, uids in entry.items():
+            if r_type == 'retells':
+                r_type = 'retelling'
+            elif r_type == 'mentions':
+                r_type = 'mention'
+            elif r_type == 'parallels':
+                r_type = 'full'
+
+            if r_type == 'full':
+                full = [uid for uid in uids if not uid.startswith('~')]
+                partial = [uid for uid in uids if uid.startswith('~')]
+                for from_uid in full:
+                    m = regex.search('[0-9]+$', from_uid)
+                    if m:
+                        from_nr = int(m[0])
+                    else:
+                        from_nr = 0
+                    true_from_uids = uid_matcher.get_matching_uids(from_uid)
+                    if not true_from_uids and ' ' not in from_uid:
+                        logging.error(
+                            f'Relationship from uid could not be matched: {from_uid} (dropped)'
+                        )
+                        continue
+
+                    for to_uids, is_resembling in ((full, False), (partial, True)):
+                        for to_uid in to_uids:
+                            if to_uid == from_uid:
+                                continue
+                            true_to_uids = uid_matcher.get_matching_uids(to_uid)
+                            if not true_to_uids:
+                                logging.info(
+                                    f'Relationship to uid could not be matched: {to_uid} (appears as orphan)'
+                                )
+                                true_to_uids = ['orphan']
+
+                            for true_from_uid in true_from_uids:
+                                for true_to_uid in true_to_uids:
+                                    remark = remarks.get(
+                                        frozenset([true_from_uid, true_to_uid]), None
+                                    )
+                                    ll_edges.append(
+                                        {
+                                            '_from': true_from_uid,
+                                            '_to': true_to_uid,
+                                            'from': from_uid,
+                                            'number': from_nr,
+                                            'to': to_uid.lstrip('~').split('-')[0],
+                                            'type': r_type,
+                                            'resembling': is_resembling,
+                                            'remark': remark,
+                                        }
+                                    )
+            else:
+                first_uid = uids[0]
+                m = regex.search('[0-9]+$', first_uid)
+                if m:
+                    from_nr = int(m[0])
+                else:
+                    from_nr = 0
+                true_first_uids = uid_matcher.get_matching_uids(first_uid)
+                for true_first_uid, to_uid in product(true_first_uids, uids[1:]):
+                    true_from_uids = uid_matcher.get_matching_uids(to_uid)
+                    if not true_from_uids and ' ' not in from_uid:
+                        logging.error(
+                            f'Relationship from uid could not be matched: {from_uid} (dropped)'
+                        )
+                        continue
+                    for true_from_uid in true_from_uids:
+                        remark = remarks.get(
+                            frozenset([true_from_uid, true_first_uid]), None
+                        )
+                        ll_edges.append(
+                            {
+                                '_from': true_first_uid,
+                                '_to': true_from_uid,
+                                'from': first_uid.lstrip('~'),
+                                'to': to_uid,
+                                'number': from_nr,
+                                'type': r_type,
+                                'resembling': any(
+                                    x.startswith('~') for x in [first_uid, from_uid]
+                                ),
+                                'remark': remark,
+                            }
+                        )
+                        m = regex.search('[0-9]+$', to_uid)
+                        if m:
+                            to_nr = int(m[0])
+                        else:
+                            to_nr = 0
+                        ll_edges.append(
+                            {
+                                '_from': true_from_uid,
+                                '_to': true_first_uid,
+                                'from': to_uid,
+                                'to': first_uid.lstrip('~'),
+                                'number': to_nr,
+                                'type': r_type,
+                                'resembling': any(
+                                    x.startswith('~') for x in [first_uid, from_uid]
+                                ),
+                                'remark': remark,
+                            }
+                        )
+
+    # Because there are many edges (nearly 400k at last count) chunk the import
+    db['relationship'].truncate()
+    for chunk in chunks(ll_edges, 10000):
+        db['relationship'].import_bulk_logged(chunk, from_prefix='super_nav_details', to_prefix='super_nav_details')
--- a/server/src/data_loader/relationships.py
+++ b/server/src/data_loader/relationships.py
@@ -147,12 +147,10 @@ def handle_other_parallels(from_uid, ll_edges: list[Any], r_type: str | Any, rem
                     '_from': true_first_uid,
                     '_to': true_from_uid,
                     'from': first_uid.lstrip('~'),
-                    'to': to_uid,
+                    'to': to_uid.lstrip('~'),
                     'number': from_nr,
                     'type': r_type,
-                    'resembling': any(
-                        x.startswith('~') for x in [first_uid, from_uid]
-                    ),
+                    'resembling': any(x.startswith('~') for x in [first_uid, to_uid]),
                     'remark': remark,
                 }
             )
@@ -161,13 +159,11 @@ def handle_other_parallels(from_uid, ll_edges: list[Any], r_type: str | Any, rem
                 {
                     '_from': true_from_uid,
                     '_to': true_first_uid,
-                    'from': to_uid,
+                    'from': to_uid.lstrip('~'),
                     'to': first_uid.lstrip('~'),
                     'number': to_nr,
                     'type': r_type,
-                    'resembling': any(
-                        x.startswith('~') for x in [first_uid, from_uid]
-                    ),
+                    'resembling': any(x.startswith('~') for x in [first_uid, to_uid]),
                     'remark': remark,
                 }
             )
--- a/server/src/data_loader/relationships.py
+++ b/server/src/data_loader/relationships.py
@@ -331,24 +331,27 @@ def load_relationships(change_tracker: FileChangeTracker, relationship_dir: Path
     Edge.load_remarks(notes_data)
 
     unmatched = Unmatched()
-    entries = all_entries(relationship_data)
+    raw = raw_entries(relationship_data)
+    entries = all_entries(raw)
     import_entries(db, entries, unmatched)
     unmatched.log()
 
 
-def import_entries(db: StandardDatabase, entries: list[Entry], unmatched: Unmatched) -> None:
+def import_entries(db: StandardDatabase, entries: Iterator[Entry], unmatched: Unmatched) -> None:
     db['relationship'].truncate()
-    edges = entries_to_edges(tqdm(entries), unmatched)
+    edges = entries_to_edges(entries, unmatched)
     for chunk in edge_chunks(edges):
         write_chunk(db, chunk)
 
 
-def all_entries(relationship_data: list[dict]) -> list[Entry]:
-    return [
-        Entry(entry_type, encodings)
-        for entry_data in relationship_data
-        for entry_type, encodings in entry_data.items()
-    ]
+def raw_entries(relationship_data: list[dict]) -> Iterator[tuple[str, list[str]]]:
+    for entry in tqdm(relationship_data):
+        for entry_type, encodings in entry.items():
+            yield entry_type, encodings
+
+
+def all_entries(entries: Iterable[tuple[str, list[str]]]) -> Iterator[Entry]:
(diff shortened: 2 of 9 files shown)


jhanarato:
And it is massively slow! Hmm... what have I done.

Yeah, haven't actually run the ETL for days. The profiler is your friend, here's the incantation:

```
uv run --env-file=env/.local.dev.env python -m cProfile -o profiler_output.prof -m sc_flask.manage load_data
```

server/src/data_loader/relationships.py
```diff
@@ -0,0 +1,246 @@
...
+        'parallels': 'full',
+    }
+
+    for entry_data in relationship_data:
+        entry_data.pop('remarks', None)
+        for entry_type, encodings in entry_data.items():
+            yield Entry(
+                entry_type=entry_type,
+                relationship_type=relationship_types[entry_type],
+                encodings=[Encoding(encoding) for encoding in encodings]
```
jhanarato: This line is causing the ETL to grind to a halt.

jhanarato:
Ahh... here's the culprit:

```
profiler_output.prof% stats relationships 5
Tue Jun  9 18:13:42 2026    profiler_output.prof

         218306746 function calls (218170871 primitive calls) in 125.843 seconds

   Ordered by: internal time
   List reduced from 6104 to 26 due to restriction <'relationships'>
   List reduced from 26 to 5 due to restriction <5>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    1.979    1.979   72.545   72.545 /home/jr/Code/suttacentral/server/src/data_loader/relationships.py:117(load_relationships)
      200    0.008    0.000   70.336    0.352 /home/jr/Code/suttacentral/server/src/data_loader/relationships.py:96(<listcomp>)
     4830    0.007    0.000    0.057    0.000 /home/jr/Code/suttacentral/server/src/data_loader/relationships.py:161(create_parallel_edges)
     2722    0.005    0.000    0.072    0.000 /home/jr/Code/suttacentral/server/src/data_loader/relationships.py:146(parallels_edges)
        1    0.004    0.004    0.100    0.100 /home/jr/Code/suttacentral/server/src/data_loader/relationships.py:63(all_uids)
```

jhanarato:
Here's the benchmark to beat:

```
relationships_main.prof% stats generate_relationship_edges
Tue Jun  9 18:29:16 2026    relationships_main.prof

         399568411 function calls (399153773 primitive calls) in 244.041 seconds

   Ordered by: internal time
   List reduced from 6300 to 1 due to restriction <'generate_relationship_edges'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.610    0.610    9.132    9.132 /home/jr/Code/suttacentral/server/src/data_loader/arangoload.py:104(generate_relationship_edges)
```

server/src/data_loader/relationships.py
```diff
@@ -0,0 +1,246 @@
...
+from common.utils import chunks
+from data_loader.ports import FileChangeTracker
+from data_loader.util import json_load
+
+
+class Encoding:
+    all_uids: set[str] = set()
+
+    def __init__(self, encoding: str):
+        self._matcher = UidMatcher(self.all_uids)
```
jhanarato: This is probably the source of our performance issues. Instead of storing the UIDs, we should store the matcher as a class variable.

jhanarato:
Easy fix! 
```
relationships_refactored.prof% sort tottime
relationships_refactored.prof% stats load_relationships
Tue Jun  9 19:38:49 2026    relationships_refactored.prof

         411517775 function calls (411103131 primitive calls) in 239.376 seconds

   Ordered by: internal time
   List reduced from 6314 to 1 due to restriction <'load_relationships'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.009    0.009   14.592   14.592 /home/jr/Code/suttacentral/server/src/data_loader/relationships.py:116(load_relationships)
```

OK, about 50% slower, but not a big change relative to the ETL run.

server/src/data_loader/relationships.py
jhanarato: Fixed!

server/tests/data_loader/test_relationships.py
```diff
@@ -0,0 +1,344 @@
...
+
+import pytest
+from arango.collection import StandardCollection
+from arango.database import StandardDatabase
+
+from common.arangodb import get_db
+from common.utils import current_app
+from data_loader.ports import FileChangeTracker
+from data_loader.relationships import generate_relationship_edges
+from fakes import FakeFileChangeTracker
```
jhanarato, line 16: This gives us an error in CI due to `fawltydeps`. Perhaps put in `conftest.py`?

jhanarato: Or we can add `__init_.py` to our test tree.

jhanarato:
I'm going to write a characterization test: 

For `main` and `refactor_relationships`, dump the sorted `relationship` collection with `_key`, `_id` and `_rev` removed. 

The results should be identical.

jhanarato:
OK, my characterization test indicates I'm missing 120 edges out of over 400K:

```
❯ wc -l relationships_*
   421159 relationships_main.json
   421039 relationships_refactored.json
   842198 total
   ```

I'll take a look.

**Comment by jhanarato** [comment]
OK, maybe a couple of issues:

https://github.com/suttacentral/sc-data/blob/7855ada307e9679b373bbcf8152f0c09ec072b81/relationship/parallels.json#L65252-L65257

The `main` branch sets `resembling` to `False` while the `refactor_relationships` sets it to `True`.

https://github.com/suttacentral/sc-data/blob/7855ada307e9679b373bbcf8152f0c09ec072b81/relationship/parallels.json#L70096-L70101

Two of the resulting edges are missing in the refactored code.

**Comment by jhanarato** [comment]
This has additional edges with the refactored code:

https://github.com/suttacentral/sc-data/blob/7855ada307e9679b373bbcf8152f0c09ec072b81/relationship/parallels.json#L20089-L20164

**Comment by jhanarato** [comment]
We can now see the full list of changes to edge data from `main` to `refactor_relationships` in this commit:

a431d574e36c256c30d68fe19b03bdc582bc2a3c

**Comment by jhanarato** [comment]
There's a discrepancy here:

image

Mentions with UID expansion are not expanded in my refactored code:

**Comment by jhanarato** [comment]
> Mentions with UID expansion are not expanded in my refactored code

This turned out to be an issue where both to and from encodings need to be expanded. For the latter, I was trying to consume an iterator more than once. 

Problem solved and I'm  back to refactoring.

**Comment by jhanarato** [comment]
Hey @ayya-vimala,

As I understand it, the encoding `pli-tv-pvr7#97.1-#98.1` means the sequence of paragraphs `97.1` to `98.1` in the root text with UID `pli-tv-pvr7`. Is that correct?

Thanks again.

**Comment by ayya-vimala** [comment]
> As I understand it, the encoding pli-tv-pvr7#97.1-#98.1 means the sequence of paragraphs 97.1 to 98.1 in the root text with UID pli-tv-pvr7. Is that correct?

Yes, indeed

pli-tv-pvr7 is the actual text-name as it is defined on SC: you can add `https://suttacentral.net/` before it and make a link that takes you to the suttaplex card: https://suttacentral.net/pli-tv-pvr7

Then anything that starts with a # is an actual ID within the text and something like #id1-#id2 is a range of several paragraphs with these ids.

**Comment by jhanarato** [comment]
Much obliged!

> pli-tv-pvr7 is the actual text-name as it is defined on SC: you can add `https://suttacentral.net/` before it and make a link that takes you to the suttaplex card: https://suttacentral.net/pli-tv-pvr7

Yeah, that's a "true uid" as the old code had it.

> Then anything that starts with a # is an actual ID within the text and something like #id1-#id2 is a range of several paragraphs with these ids.

I suspect this is handled in the front end. This is what turns up in the API:

```
"pli-tv-pvr7#97.1-#100.1": [
    {
      "enumber": 1,
      "to": {
        "to": "an7.75-78",
        "volpages": "PTS AN 4.141",
        "alt_volpages": null,
        "acronym": "AN 7.78",
        "uid": "an7.78",
        "root_lang": "pli",
        "original_title": "Catutthavinayadharasutta",
        "translated_title": null,
        "type": "mention",
        "from": "super_nav_details/pli-tv-pvr7",
        "biblio": null,
        "translations": [
          {
          ...
```

So it's not being broken down in anyway between the `parallels.json` file and the API. The full string is preserved.

**Comment by jhanarato** [comment]
Apologies for the crossover in conversation with issue #3655

**Comment by jhanarato** [comment]
And some time later... we're ready for review!

I spent a lot more time on this than expected. It does mostly the same thing as it did before, except:

- Will load the relationships if either `parallels.json` or `notes.json` change. Previously `notes.json` would not trigger a load.
- Handles entries that are both mentions and resembling.
- `tqdm()` works properly.
- It is somewhat slower at about 17 seconds compared to 9 seconds for the original code.

I don't think the latter is a deal breaker. Adding 8 seconds to a 2 minute process is nothing to worry about, so long as we don't see any instability due to increased CPU usage. I'll spend a bit of time looking into this with the profiler. Having proper `tqdm()` output should make any future performance regressions or improvements obvious.

I created a characterization test that dumps the contents of the `relationship` collection to a CSV file. It's many megabytes, but great if you just want to test that nothing has changed.

The rains retreat is coming up fast, so it would be great to get this reviewed and into production before then.

**Comment by jhanarato** [comment]
Profiler output for main & refactored:

```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.663    0.663    9.721    9.721 arangoload.py:104(generate_relationship_edges)
```

```
ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   37.173   37.173 relationships.py:310(load_relationships)
```

So considerably slower. The overall time for both is affected by the profiling. I get 17 seconds via `tqdm()` when run normally.

**Comment by jhanarato** [comment]
I just checked the stats via my stage printer. Not good...

The current code runs in 8.7 seconds, with 1.5 seconds of CPU time.

The new code takes 17.6 seconds, but the CPU time is up to 10.2 seconds.

That's unacceptable given our stability issues. 

Given my time constraints, this is unlikely to be solved before the rains retreat. It doesn't actually solve any current problems, though it is a fine reference implementation. Perhaps I'll find time dig in with the profiler eventually.

File: server/src/data_loader/relationships.py [hunk 4]
```diff
@@ -0,0 +1,376 @@
...
+from common import utils
+from data_loader.ports import FileChangeTracker
+from data_loader.util import json_load
+
+
+class Encoding:
+    _matcher: UidMatcher = UidMatcher(set())
+
+    @classmethod
+    def load_uids(cls, uids: Iterable[str]) -> None:
```
jhanarato, line 23: This is here for performance reasons. Creating a `UidMatcher` every time was extremely slow.

server/src/data_loader/relationships.py
```diff
@@ -0,0 +1,376 @@
...
+        return {
+            '_from': self._uids.from_,
+            '_to': self._uids.to,
+            'from': self._from(),
+            'to': self._to(),
+            'number': self._number(),
+            'type': str(self._edge_type),
+            'resembling': self._resembling(),
+            'remark': self._remark(),
+        }
```
jhanarato, line 190: We spend a lot of time in this method:

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   439087    0.603    0.000    6.181    0.000 relationships.py:172(as_dict)
```

jhanarato:
Well... sorting by `cumtime` we get this:

```
after.prof% stats relationships
Fri Jul 10 10:12:54 2026    after.prof

         470731399 function calls (470308827 primitive calls) in 263.094 seconds

   Ordered by: cumulative time
   List reduced from 6101 to 84 due to restriction <'relationships'>

   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
        1    0.000    0.000   37.173   37.173 relationships.py:310(load_relationships)
        1    0.037    0.037   36.901   36.901 relationships.py:331(import_entries)
     8084    0.018    0.000   25.524    0.003 relationships.py:371(write_chunk)
    16209    0.015    0.000   11.237    0.001 relationships.py:365(edge_chunks)
```

Why are we calling `write_chunk()` so many times?

We produce about 430,000 edges and our chunk size is 10,000. That would be 43 chunks right?

server/src/data_loader/relationships.py
```diff
@@ -0,0 +1,376 @@
...
+    entries = all_entries(relationship_data)
+    import_entries(db, entries, unmatched)
+    unmatched.log()
+
+
+def import_entries(db: StandardDatabase, entries: list[Entry], unmatched: Unmatched) -> None:
+    db['relationship'].truncate()
+    for entry in tqdm(entries):
+        edges = edges_for_entry(entry, unmatched)
+        for chunk in edge_chunks(edges):
```
jhanarato: There's a bug here. We end up writing a chunk for every entry. Instead of around 40 calls to `write_chunk()` we have over 8000

jhanarato:
Yeah, there's a bug there. I should have time to come up with a fix in the next few days.

jhanarato:
Chunking now works properly.

Clock time: 12.1s,  CPU time: 4.65s

Perhaps clock time is a reasonable measure. The time not spent in the Python is sure to be used by ArangoDB.

But I'll see what else I can optimise.

agilgur5:
> The time not spent in the Python is sure to be used by ArangoDB.

I'm pretty sure the non-CPU time is used by ArangoDB (which for the Python is I/O time). It's been pretty consistently around ~7s in all of your past measurements:

> Clock time: 12.1s, CPU time: 4.65s



> The current code runs in 8.7 seconds, with 1.5 seconds of CPU time.
> 
> The new code takes 17.6 seconds, but the CPU time is up to 10.2 seconds.

jhanarato:
> I'm pretty sure the non-CPU time is used by ArangoDB (which for the Python is I/O time).

I'm measuring clock and CPU time for the Python. I'd imagine ArangoDB will be working the CPU, but it's not visible in the profiling. The Python code is definitely using up more CPU time, but I guess we should count both ArangoDB time and Python.

Parsing the `parallels.json` file is still pretty CPU heavy, not just IO. I'll continue profiling to see if there's any low hanging fruit.

agilgur5:
> I'd imagine ArangoDB will be working the CPU, but it's not visible in the profiling

From Python's perspective, it's making a request out to ArangoDB and getting a response back. They happen to be running on the same CPU, but Python doesn't know that.

You could run an OS or kernel profiler to see both on the same CPU, but I don't think that's particularly useful in this case as you won't have interpreter-level visibility

jhanarato:
> You could run an OS or kernel profiler to see both on the same CPU, but I don't think that's particularly useful in this case as you won't have interpreter-level visibility

We're now bulk loading the same documents in batches of 10K as we did before. I assume the performance of ArangoDB is the same as it was before.

jhanarato:
I had a look at the profiling for matching uids:

Before refactoring:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
   476790    0.155    0.000    0.659    0.000 uid_matcher.py:79(get_matching_uids)
    79396    0.077    0.000    0.268    0.000 order.py:32(get_matching_uids)
```

After refactoring:
```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
  1333890    0.342    0.000    2.178    0.000 relationships.py:38(matching_uids)
  1333890    0.440    0.000    1.836    0.000 uid_matcher.py:79(get_matching_uids)
    79672    0.075    0.000    0.263    0.000 order.py:32(get_matching_uids)
```

That's about 2.8 X calls to `UidMatcher.get_matching_uids()`. You can see the difficulty in profiling the original code. There was just one big function. However, the `UidMatcher` hasn't changed so it's a good proxy. 

And there's an obvious optimization. We call `Encoding.matching_uids()` in `has_matching_uids()`. We can store the uids in an instance attribute when we initialise `Encoding` objects.

```
   ncalls  tottime  percall  cumtime  percall filename:lineno(function)
    79672    0.076    0.000    0.263    0.000 order.py:32(get_matching_uids)
    49685    0.020    0.000    0.125    0.000 uid_matcher.py:79(get_matching_uids)
  1333890    0.111    0.000    0.111    0.000 relationships.py:39(matching_uids)
  ```

That's a tenth of the calls now.

Clock time is 10.9 seconds, CPU time 3.7 seconds (python only).

**Comment by jhanarato** [comment]
After moving the number extraction to an Encoding instance attribute:

Clock time 10.1 seconds, CPU time 2.3 seconds (python only).

**Comment by jhanarato** [comment]
This is good enough, performance wise.

All we need to do now is fix the linter error. I'm not sure how to do that. @agilgur5 ?

**Comment by agilgur5** [comment]
Copying the `fawltydeps` linter error [from CI](https://github.com/suttacentral/suttacentral/actions/runs/29548077317/job/87784595204?pr=3663):

> ```
> These imports appear to be undeclared dependencies:
> - 'fakes'
> ```

Can you try `from .fakes` instead? Since `base_dir` is `src/`, `from fakes` would be interpreted as relative to it (or in the `venv`), but it's in `tests/data_loader/`. Versus `.fakes` is relative to the current file

File: server/tests/data_loader/test_relationships.py [hunk 7]
```diff
@@ -0,0 +1,830 @@
...
+import pytest
+from arango.collection import StandardCollection
+from arango.database import StandardDatabase
+
+from common.arangodb import get_db
+from common.utils import current_app
+from data_loader.ports import FileChangeTracker
+from data_loader.relationships import load_relationships, Encoding, Edge, Entry, \
+    EntryType, EdgeType, OtherEdges, Remarks, EdgeUids, EdgeEncodings, Unmatched, ParallelsEdges, entries_to_edges
+from fakes import FakeFileChangeTracker
```
agilgur5, line 16:
```suggestion
from .fakes import FakeFileChangeTracker
```

server/tests/data_loader/test_fakes.py
```diff
@@ -0,0 +1,34 @@
+from pathlib import Path
+
+from fakes import FakeFileChangeTracker
```
agilgur5, line 3:
```suggestion
from .fakes import FakeFileChangeTracker
```

server/tests/data_loader/test_relationships.py
jhanarato: That solves the `fawltydeps` issue, but now my tests won't run:

```
test_fakes.py:3: in <module>
    from .fakes import FakeFileChangeTracker
E   ImportError: attempted relative import with no known parent package
```

jhanarato: Can we just add an ignore for now?

jhanarato: I see I could just edit the `pyproject.toml` file. But is there a "right" way to handle this?

server/pyproject.toml
```diff
@@ -52,5 +52,9 @@ ignore_unused = [
     "uwsgi", # used as CLI in `Makefile`, `entrypoint.sh`
 ]
 
+ignore_undeclared = [
```
jhanarato, line 55: Is this ok with you for now?

If so, that's the last hurdle and we can merge before I go on retreat.

server/tests/data_loader/test_relationships.py
jhanarato: Hey @agilgur5, over here! :wave:

Issue #3664, opened 2026-06-10
`alt_acronym` is not loaded and displayed


jhanarato:
@ayya-vimala asked me to take a look at this.

We define the `alt_volpage` in the `text_extra_info.json` file:

```json
{
    "uid": "sn11.25",
    "acronym": "SN 11.25",
    "alt_acronym": "SN 271",
    "volpage": "PTS (1st ed) SN i 240",
    "alt_volpage": "PTS (2nd ed) SN i 514",
    "alt_name": null,
    "biblio_uid": null
  },
```

But when we [load the suttaplex](https://suttacentral.net/sn11.25) the second volpage is `PTS (2nd ed) 1.515–1.516`

Apparently the original `alt_volpage` was included on the legacy site.

jhanarato:
It seems this pull request sets the `alt_volpage` to `null` in ArangoDB: #2207

The suttaplex card was updated in the PR, so there was some sort of change happening there.

@ihongda?

ayya-vimala:
It seems that the `alt_volpage` is always `null` for all non-pali texts. For the pali-texts, they are re-calculated from the bilara reference data.

ihongda:
@jhanarato @ayya-vimala  Got it, let me check it.

ihongda:
The value of `alt_volpage` will be recalculated and updated in `update_text_extra_info()`， The data source relied upon for the update is the reference data in bilara-data, for example:

<img width="824" height="367" alt="Image" src="https://github.com/user-attachments/assets/8ab328fc-9e1c-4101-8f82-fbd752572332" />

ayya-vimala:
Thank you @ihongda  - I just spoke with Bhante @Sujato and there are a few things that need to change about this:

**alt_volpage** 
Should become the number as loaded from text_extra_info.json + the last of the `pts-vp-pli2ed` numbers in Bilara references, rather than the first + last of the `pts-vp-pli2ed` numbers. The reason is that if you use the first `pts-vp-pli2ed` number in Bilrara, you have already missed the first part of the sutta because that reference number may be on the previous sutta. So if you use the one from text_extra_info.json as the first one, you've got the right numbers in that sutta.

**alt_acronym**
The alt_acronym needs to be loaded into the Arango DB from text_extra_info.json and displayed on the "nerdy row" next to the acronym itself. There is an alt_acronym for SN / AN and ITI (the latter I will upload in the next few days).

**search**
[...]

PR #3665, opened 2026-06-12
fix: improve handling of alternative volume page references.

ihongda:
Related issue: #3664

### DIFF
--- a/server/src/common/queries.py
+++ b/server/src/common/queries.py
@@ -1636,7 +1636,12 @@ UPDATE_TEXT_EXTRA_INFO_ALT_VOLPAGE = '''
 UPSERT { uid: @uid }
 INSERT { uid: @uid, acronym: null, alt_acronym: null, volpage: null, alt_volpage: @ref, alt_name: null, biblio_uid: null }
 UPDATE {
-    alt_volpage: @ref
+    alt_volpage: CONCAT_SEPARATOR(
+        ',',
+        OLD.alt_volpage != null && OLD.alt_volpage != ''
+            ? APPEND(SPLIT(OLD.alt_volpage, ','), SPLIT(@ref, ','), true)
+            : SPLIT(@ref, ',')
+    )
 } IN text_extra_info
 '''
 
--- a/server/src/data_loader/arangoload.py
+++ b/server/src/data_loader/arangoload.py
@@ -382,12 +382,12 @@ def update_text_extra_info():
     db = arangodb.get_db()
     bilara_references = list(db.aql.execute(BILARA_REFERENCES))
     for reference in tqdm(bilara_references):
-        refs = json_load(reference['file_path'])
+        reference_map = json_load(reference['file_path'])
         pts_refs_1st = []
         pts_refs_2nd = []
-        for uid, ref in refs.items():
-            refs = get_pts_ref(ref)
-            for pts_ref in refs:
+        for uid, ref in reference_map.items():
+            pts_refs = get_pts_ref(ref)
+            for pts_ref in pts_refs:
                 if pts_ref.find('pts-vp-pli2ed') != -1:
                     pts_refs_2nd.append(
                         pts_ref.replace('pts-vp-pli2ed', 'PTS (2nd ed) ')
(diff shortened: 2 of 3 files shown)


ihongda:
Hi Bhante @jhanarato, 
image
 
There is a test error here; does it need to be fixed?

jhanarato:
> Hi Bhante @jhanarato, image
> 
> There is a test error here; does it need to be fixed?

Hmm... I'll take a look

jhanarato:
Can you show me the full assertion failure message?

Oh, sorry, that's a screenshot of the CI run. It truncates the assert which is unhelpful.

@agilgur5 - is there anything we've done recently that might have caused this? It's a pretty simple test. Create some directories via the `tmp_path` fixture and check `language_directories()` returns them.

agilgur5:
> @agilgur5 - is there anything we've done recently that might have caused this?

CI on `main` [passes](https://github.com/suttacentral/suttacentral/actions/runs/27252130642/job/80478666731) and this branch is off latest `main`. But CI seems to have failed [on the previous commit](https://github.com/suttacentral/suttacentral/actions/runs/27249633619/job/80471060358) on this same test.
Neither change has anything to do with this test.

So I would suspect that it's a flaky test for one reason or another. There's a retry button in the GHA job which you can use to test that theory in CI

jhanarato:
> So I would suspect that it's a flaky test for one reason or another. There's a retry button in the GHA job which you can use to test that theory in CI

Good idea! Note there's an option to enable debug logging:

image

Nope, it's reproducible.

Could it be related to #3661?

jhanarato:
I just created a dummy PR #3666 and get the same issue. Now to reproduce it locally.

Here's my hunch. It relates to `pathlib`? I can imagine a security patch changing that behavior.

jhanarato:
I ran this locally with no issues:

```cli
docker compose -f docker-compose.yml -f docker-compose.dev.yml build --no-cache sc-flask
make run-dev
make test-server-ci
```

jhanarato:
Oh I see, @ihongda, 

If you scroll up in the CI, you'll see the full message:

```
>       assert language_directories(html_dir=tmp_path) == directories
E       AssertionError: assert [PosixPath('/...rectory0/en')] == [PosixPath('/...ectory0/lzh')]
E         At index 0 diff: PosixPath('/tmp/pytest-of-root/pytest-0/test_with_language_directory0/lzh') != PosixPath('/tmp/pytest-of-root/pytest-0/test_with_language_directory0/en')
E         Use -v to get more diff

tests/data_loader/test_unsegmented.py:369: AssertionError

```

jhanarato:
Aha! I see it now:

```
PosixPath('/tmp/pytest-of-root/pytest-0/test_with_language_directory0/lzh')
PosixPath('/tmp/pytest-of-root/pytest-0/test_with_language_directory0/en')
```

I'm not sorting the directories. 

Here's the fix:

https://github.com/suttacentral/suttacentral/pull/3666/changes

jhanarato:
Nice work @ihongda !

agilgur5:
> Here's my hunch. It relates to `pathlib`? I can imagine a security patch changing that behavior.

> I'm not sorting the directories.

Huh, does that mean that `Path.glob` returns directories in a non-deterministic order? As sometimes it's sorted and other times it isn't 
https://github.com/suttacentral/suttacentral/blob/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/server/src/data_loader/unsegmented.py#L154-L155

**EDIT**: Yep, indeed, [the docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) do mention the ordering:

> Note: The paths are returned in no particular order. If you need a specific order, sort the results.

jhanarato:
> Huh, does that mean that `Path.glob` returns directories in a non-deterministic order? As sometimes it's sorted and other times it isn't

Yeah, weird eh? Might be related to the semver changes, but maybe not. I couldn't reproduce the bug on my own machine. Who knows, the github runner may have changed.

agilgur5:
You may have missed my edit [above](https://github.com/suttacentral/suttacentral/pull/3665#issuecomment-4697753674):

> **EDIT**: Yep, indeed, [the docs](https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob) do mention the ordering:
>
> > Note: The paths are returned in no particular order. If you need a specific order, sort the results.

So that is documented behavior

Issue #2013, opened 2021-02-24
Patimokkha texts: a new model


sujato:
## the problem

If we look in Vinaya, there are two main kinds of text. 

- A full Vinaya analysis such as we find in Pali. These include a kind of text called Vibhanga, which have rules + explanation. They include `vb` in the UID.
- A plain set of rules (patimokkha, `pm` ). 

The Vibhangas work okay, but the patimokkhas don't appear at all.

https://staging.suttacentral.net/pitaka/vinaya/pli-tv-vi/pli-tv-bu-pm

There's no root text. These root texts are in fact included in root/pli/vri. They are not part of the `ms` edition.

It's the same with Chinese patimokkha texts:

https://staging.suttacentral.net/pitaka/vinaya/san-mg-vi

## the bad old solution

Since the Patimokkhas are plain lists of rules, we do not divide them into separate pages per rule as with the Vibhangas. Instead, we simply give the whole patimokkha on a single page.

The problem is that, because there are so many rule parallels, we cannot list all the parallels in a single suttaplex.

The old site uses a clumsy solution.

- give a suttaplex card for the whole patimokkha text. 
- then give separate cards for each rule. 

Unfortunately, in the "whole text" card, it says there are no parallels, which is incorrect, then in the parallel cards it says there is no text, which is also incorrect.

## a better way: separate text from parallels

- Clicking on the patimokkha takes you to the main patimokkha card as per here: https://suttacentral.net/pli-tv-bu-pm
- However, this does not have the "parallels" section, just the texts.
- Underneath the suttaplex card is a button: "Show rule parallels and details".
- If you click it, it loads the suttaplex cards for all rules, together with blurbs and parallels. 
- However, these cards have no mention of the texts or translations.

Mockups:

default view

![PATIMOKKHA_DEFAULT](https://user-images.githubusercontent.com/6112010/108925014-f489bb00-768f-11eb-861a-3c9db767b32f.png)

detail view

![PATIMOKKHA_Details](https://user-images.githubusercontent.com/6112010/108925034-fce1f600-768f-11eb-867e-6eda0ad3e3ba.png)

PR #3675, opened 2026-07-14
fix(client): `pm-2` is also Patimokkha

### LINKED ISSUE #433: Bhikkhu Pātimokkha (2nd) in Sarvastivada Vinaya is empty
Is there supposed to be any root text for the rules under https://suttacentral.net/lzh-sarv-bu-pm-2? I checked and I couldn't find any root text. If there is none, then Bhikkhu Patimokkha (2nd) should be completely removed.

Tagging @ayya-vimala

agilgur5:
## Summary

Check for `-pm-2` endings as well in `isPatimokkha` functions
- Fixes https://github.com/suttacentral/sc-data/issues/433
  - See my root cause analysis in https://github.com/suttacentral/sc-data/issues/433#issuecomment-4972719377 et al

## Details

- these functions only checked if the ID ended in `-pm`, but sometimes there are 2, such as with [`lzh-sarv-bu-pm-2.html`](https://github.com/suttacentral/sc-data/blob/254154c7ed0a10ab24307be36dd8a0d3934604fb/html_text/lzh/vinaya/lzh-sarv-bu-pm-2.html)
  - modify them to also check for ending in `-pm-2`

- In https://github.com/suttacentral/sc-data/issues/433, the result of this check was that a suttaplex card was not displaying for `pm-2` despite the data being near identical between `pm` and `pm-2`

## References

- The logic originates from #2013 / #2271

## Verification

This now matches what PM1 looks like, screenshots:
<table>
    <tr>
      <th>Before</th>
      <th>After</th>
   </tr>
   <tr>
      <td>Screenshot 2026-07-19 at 9 38 01 PM</td>
      <td>Screenshot 2026-07-19 at 9 03 56 PM</td>
    </tr>
</table>

### DIFF
--- a/client/elements/navigation/sc-navigation.js
+++ b/client/elements/navigation/sc-navigation.js
@@ -301,7 +301,7 @@ export class SCNavigation extends LitLocalized(LitElement) {
   }
 
   _isPatimokkha(uid) {
-    return uid.endsWith('-pm');
+    return uid.endsWith('-pm') || uid.endsWith('-pm-2');
   }
 
   _computeMenuApiUrl(uid) {
--- a/client/elements/suttaplex/sc-suttaplex-list.js
+++ b/client/elements/suttaplex/sc-suttaplex-list.js
@@ -121,11 +121,11 @@ class SCSuttaplexList extends LitLocalized(LitElement) {
   }
 
   isPatimokkha() {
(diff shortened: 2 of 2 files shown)


ihongda:
Thanks @agilgur5

PR #3680, opened 2026-07-19
refactor(client): use more variables around `search:resultsFor`

agilgur5:
## Summary

~The localization key `search:resultsFor` itself was showing up in the search UI momentarily instead of the localized text.~
Refactors some code around `search:resultsFor` to make it easier and less ambiguous to read.

This was a change I made while [attempting to root cause it](https://discourse.suttacentral.net/t/suttacentral-search-bugs/32425/156?u=agilgur5) a bug reported [on the forum](https://discourse.suttacentral.net/t/suttacentral-search-bugs/32425/151?u=agilgur5) by Ven. @delightedbydhamma. This change failed to fix it, but did help root cause I suppose, see the comments below

## Details

- refactor the code a bit around `${this.localize('search:resultsFor')}` to use more variables and avoid awkward/ambiguous formatting of whitespace

## Verification

- Tested locally, search runs the same as before

### DIFF
--- a/client/elements/sc-page-search.js
+++ b/client/elements/sc-page-search.js
@@ -1244,13 +1244,11 @@ export class SCPageSearch extends LitLocalized(LitElement) {
   #createMetaData() {
     const description = this.localize('search:metaDescriptionText');
     const searchResultsText = this.localize('search:searchResultsText');
-    const pageTitle = `${this.#calculateResultCount(this.resultCount)} ${this.localize(
-      'search:resultsFor'
-    )} ${this.searchQuery}`;
+    const resultsFor = this.localize('search:resultsFor');
 
-    const toolbarTitle = `${this.#calculateResultCount(this.resultCount)} ${this.localize(
-      'search:resultsFor'
-    )} <strong class="highlightTitle">${this.searchQuery}</strong>`;
+    const countResultsFor = `${this.#calculateResultCount(this.resultCount)} ${resultsFor}`;
+    const pageTitle = `${countResultsFor} ${this.searchQuery}`;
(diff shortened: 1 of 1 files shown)


agilgur5:
> `search:resultsFor` itself was showing up in the search UI momentarily

Hmm, I didn't think about this somehow, but if it's momentary, that means it probably does get localized correctly _eventually_. Which might explain the `search:search` issue in the bug report as well.
The whitespace may well be ignored because `search:resultsFor` itself is still quoted, so it is equivalent to, for example, `${this.localize(  'search:resultsFor'  )}`

In which case, this probably won't fix it, but is still a good refactor IMO.

If testing confirms that, then the reason why both keys are showing up instead of their localized text is probably because the localization data has not been loaded yet? Though I thought the rendering would pause for that... odd why this is the only page with this issue 🤔

agilgur5:
> If testing confirms that

Yea this doesn't fix it, so I'll rewrite the description as a tiny refactor instead I suppose

> then the reason why both keys are showing up instead of their localized text is probably because the localization data has not been loaded yet? Though I thought the rendering would pause for that... odd why this is the only page with this issue 🤔

So the localized text does not load until the search request is complete. On slower requests this is visible for some time.

I'm suspecting this is perhaps because of [the `until` directive](https://github.com/suttacentral/suttacentral/blob/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/client/elements/sc-page-search.js#L403) perhaps? Search is the [only place](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+until+language%3AJavaScript&type=code&l=JavaScript) it is used.
It seems like `until` is supposed to behave similar to React's Suspense. This component might need a bit of refactoring so that the localization can load before the search request is complete.

ihongda:
Thanks @agilgur5

PR #3681, opened 2026-07-20
chore(client): auto-update element EN localization files

agilgur5:
## Summary

Commit the results of the auto-update of files in `client/localization/elements/`

## Details

- these were automatically updated while I was doing some set-up and testing of the front-end
  - [during `make load-data`](https://github.com/suttacentral/suttacentral/blob/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/server/src/data_loader/arangoload.py#L663), I believe
  
## Notes for Reviewers

Does this need to be committed to this codebase? Can it be `gitignore`'d? Especially if it is not updated frequently (or automatically by CI)
Not sure if there is a compelling reason to have a copy of these files, but it is generally good practice to only have sources in "source control" 😅


jhanarato:
Oh right. Yeah, that always bugged me. The files are modified every time you run `make load-data`

So these files were last modified by @ihongda in 2022, but Blake committed them three times in October 2021. 

Many more are created by the ETL, but there's a `.gitignore` file there:

https://github.com/suttacentral/suttacentral/blob/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/client/localization/elements/.gitignore#L1-L4

Why English files are allowed and others aren't, I have no idea. That reason may no longer be valid.

I recommend removing the English JSON files from git. Check that everything still works as expected. Obviously.

agilgur5:
> So these files were last modified by @ihongda in 2022, but Blake committed them three times in October 2021.

There's more times they were committed according to [the history](https://github.com/suttacentral/suttacentral/commits/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/client/localization/elements)

jhanarato:
> > So these files were last modified by @ihongda in 2022, but Blake committed them three times in October 2021.
> 
> There's more times they were committed according to [the history](https://github.com/suttacentral/suttacentral/commits/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/client/localization/elements)

My bad, I was looking at a single file:

https://github.com/suttacentral/suttacentral/commits/main/client/localization/elements/a-new-beginning_en.json

jhanarato:
So I've got about 280 files in `client/localization/elements` versus 37 in git. They would be the ones ending in `en.json`.

Are the rest generated from the English?

jhanarato:
Here's the ETL code:

https://github.com/suttacentral/suttacentral/blob/main/server/src/data_loader/copy_localization.py

I guess maybe we copy everything except the English from `bilara-data`?

Yeah, there are a bunch of files that match the glob `'*/*/site/*.json'`, even the English

https://github.com/suttacentral/bilara-data/tree/published/root/en/site

https://github.com/suttacentral/bilara-data/tree/published/translation/de/site

agilgur5:
> Yeah, there are a bunch of files that match the glob `'*/*/site/*.json'`, even the English
> 
> https://github.com/suttacentral/bilara-data/tree/published/root/en/site

Yea I had a bit of trouble finding the source English in `bilara-data` too for [an unrelated localization error](https://discourse.suttacentral.net/t/suttacentral-search-bugs/32425/169?u=agilgur5) (see my edit history there) until I found the relevant ETL. Then I realized they were in `root/en/site` and not `translation/en/site`. Which makes sense, since, in the case of the interface, the root language is English 😅

jhanarato:
OK, I'm a little confused as to what this PR entails.

There are some files to commit, or not commit...?

agilgur5:
Well the PR commits the files based on the current pattern. In doing so, I asked the question of why they are committed.

If we don't need to commit them, I'm happy to close this PR and open one removing and `gitignore`ing them instead.
I don't know the answer to that myself (otherwise I wouldn't have asked)

jhanarato:
> Well the PR commits the files based on the current pattern. In doing so, I asked the question of why they are committed.
> 
> If we don't need to commit them, I'm happy to close this PR and open one removing and `gitignore`ing them instead. I don't know the answer to that myself (otherwise I wouldn't have asked)

Our hypothesis, if I understand correctly, is that if we remove all these files in this directory from version control:

https://github.com/suttacentral/suttacentral/tree/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/client/localization/elements

They would be recreated from the `bilara-data` repo when the ETL runs.

If that's right, let me know and I'll test.

jhanarato:
OK, I went ahead and experimented:

```cli
make run-dev-rebuild
make delete-database
make migrate
make load-data
ls -A1 client/localization/elements/ > files_before.txt
git checkout HEAD client/localization/elements/
git rm client/localization/elements/*en.json
git commit -m "Remove localization files that should not have been checked in."
sudo rm -rf client/localization/elements/*.json
sudo rm -rf client/localization/elements/build/
make delete-database
make migrate
make load-data
ls -A1 client/localization/elements/ > files_after.txt
diff files_before.txt files_after.txt
```

And two files seem to be missing:

```diff 
37d36
< a-new-beginning_en.json
247d245
< static-map_en.json
```

At this point I haven't checked the contents of `client/localization/elements/build` which is created along the way by the `copy_localization()` function.

jhanarato:
Having checked: the build directory contains the same contents.

agilgur5:
> And two files seem to be missing:
> 
> ```diff
> 37d36
> < a-new-beginning_en.json
> 247d245
> < static-map_en.json
> ```

These both seem to be outdated and no longer used:
- [`static-map_en.json`](https://github.com/suttacentral/suttacentral/blob/b2cb0f91eed45e42c234e39ce17cdb87a2965167/client/localization/elements/static-map_en.json) seems to have been replaced by [`map_en.json`](https://github.com/suttacentral/suttacentral/blob/b2cb0f91eed45e42c234e39ce17cdb87a2965167/client/localization/elements/map_en.json)
  - the `StaticMap` component [currently uses `map_*.json`](https://github.com/suttacentral/suttacentral/blob/b2cb0f91eed45e42c234e39ce17cdb87a2965167/client/elements/static/sc-static-map.js#L21) as well. nothing currently uses `static-map_*.json`
- [`a_new_beginning_en.json`](https://github.com/suttacentral/suttacentral/blob/b2cb0f91eed45e42c234e39ce17cdb87a2965167/client/localization/elements/a-new-beginning_en.json#L2) is, afaict, about the "SC Next" project around ~2018, so ~8 years ago.
  - [nothing](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+a_new_beginning&type=code) currently uses it
  
So these are actually an artifact of the directory never being cleaned. If it was used as an ephemeral directory (i.e. `gitignore`'d), that would be more explicit that it could be cleaned without consequence.

So I think all signs point to this should be `.gitignore`'d and I'm not sure why it isn't. The `gitignore` lines originate from 415772576920fc9fc914d4e983d04bee2d31818c, but that commit doesn't explain why the `_en` ones are special-cased.

So I'm going to close this PR and instead open one to remove all these files and put them back in the `gitignore` unless @ihongda thinks there's a reason not to

agilgur5:
> So I'm going to close this PR and instead open one to remove all these files and put them back in the `gitignore`

✅ Completed in #3717

PR #3684, opened 2026-07-20
refactor(client): use `compose watch` to sync front-end files

agilgur5:
## Summary

During development, move from mounted volumes to `compose watch` to efficiently sync front-end files

## Details

- sync pretty much all files (except ones in `.dockerignore` like [`node_modules/`](https://github.com/suttacentral/suttacentral/blob/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/client/.dockerignore#L4)) and only rebuild on `package.json` changes
  - similar to [the example](https://docs.docker.com/compose/how-tos/file-watch/#example-1) in the docs

- I also was having trouble getting FS events propagated to the container when I changed a file during dev when using `volumes` (i.e. could not trigger live reload)
  - using `action: sync` instead entirely avoids that problem
  
## References

This matches prior changes to the back-end in https://github.com/suttacentral/suttacentral/commit/eaa794dbab6ed73219e391133b9ba650a2272433 et al / #3575

## Verification

I've been using this configuration for a bit, including for several of my recent front-end PRs like #3683 and #3675

### DIFF
--- a/docker-compose.dev.yml
+++ b/docker-compose.dev.yml
@@ -32,7 +32,6 @@ services:
           path: ./server/uv.lock
 
 
-
   sc-nginx:
     volumes:
       - ./client/:/opt/sc/static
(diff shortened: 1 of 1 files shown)


jhanarato (approved):
Nice work, @agilgur5!

I've not checked out and built this myself, but it looks good to me.

ihongda:
Thanks @agilgur5

PR #1974, opened 2021-02-10
Fixed main.js being pushed and not used by the browser

robertmusil:

Issue #1310, opened 2019-01-24
Use LitElement, remove Polymer 


sujato:
## Update

Track this via the Milestone.

## Status as of Nov 2020

The text pages and many other elements have been rewritten in LitElement, however several elements remain to be changed. We should make a list of remaining elements.

I'll be making a list and adding items to the remove polymer milestone.

## LitElement replaces Polymer

Polymer, together with its system of libraries and components, is being replaced by LitElement. We must go through the entire code base and remove all dependencies on Polymer and legacy elements and ensure they are replaced with modern alternatives.

The general policy should be to produce code that is:

- valid and semantic
- fast and light
- reliable on modern browsers
- based on web standards

Where possible, use just HTML and CSS. Use modern vanilla JS where necessary. Import web components where they make sense. At all costs, avoid heavy and unnecessary dependencies. Evaluate each change and only add the minimum code necessary. Aim to produce clean, semantic HTML.

## MWC elements

The Polymer team is updating from Polymer/paper to Material Web Components (mwc) based on LitElement.

There are several versions of this, the one we want is "Material Components Web Components"

https://github.com/material-components/material-components-web-components

This is the obvious choice for updating, and in many cases it may be the correct one. Note, however, that the official MWC elements have been somewhat compromised by their reliance on "old-school" upstream CSS in BEM, etc. The polymer team is working to improve this, but it means that as of now, MWC components are significantly larger than they need be.

An optimized version of the MWC components is being developed, but is not yet published.

https://captaincodeman.github.io/mega-material/ 

These are super-small!

Thus we should take a flexible approach to replacing elements on a case-by-case basis:

- Use MWC
- Use other third-party libraries 
- Build our own

**Review all imported web components and completely remove dependency on legacy Polymer components and other outdated components.**

There's a good list of component libraries here.

https://open-wc.org/faq/component-libraries.html 

Not all components have been directly replaced by mwc. In some cases the components are no longer necessary, as the web platform has evolved. For example:

- `neon-animation`: [deprecated](https://www.polymer-project.org/blog/2017-07-12-an-update-on-neon-animation), use CSS keyframe animations, CSS transitions, or the Web Animations API. The latter is polyfilled with `web-animations-js`, but given our limited animation requirements, we should see if this can be removed as well.
[...]

PR #3688, opened 2026-07-21
clean(deps): remove unused polyfills

agilgur5:
## Summary

Remove a few unused / unimported polyfills

## Details

- remove polyfills from dependencies in `package*.json` manifests, as well as copied files in Webpack config, the commented out NGINX conf, and `web-test-runner` config

### specific polyfill details
- `@webcomponents/webcomponentsjs` and `web-animations-js` were [not](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral%20webcomponentsjs&type=code) [imported](https://github.com/search?q=repo%3Asuttacentral%2Fsuttacentral+web-animations-js&type=code) anywhere and so were unused
  - (nor could I find them in independently loaded script tags) 
  - `@webcomponents/webcomponentsjs` _was_ loaded in a `<script>` tag in the NGINX conf, but that line was commented out 5.5 years ago in https://github.com/suttacentral/suttacentral/pull/1974 (the second half seemed to be commented out [on accident](https://github.com/suttacentral/suttacentral/pull/1974#discussion_r3624844311)?)
    - so it hasn't been used since at least 2021, and the lack of polyfills appears to not have been a major problem with users
      - which I'd expect, as all the features are baseline available in all modern browsers:
        - Lit supports to [at least 2021](https://lit.dev/docs/tools/requirements/), [`customElements`](https://developer.mozilla.org/en-US/docs/Web/API/Window/customElements) baseline available since 2020, [`<template>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/template) since 2015, [`<slot>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/slot) since 2020, [`ShadowRoot`](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot) since 2020, [`parentNode`](https://developer.mozilla.org/en-US/docs/Web/API/Node/parentNode) since 2015
  - I could not find when `web-animations-js` was actually last used, but it was supposed to be reviewed for removal during the Polymer -> Lit migration per https://github.com/suttacentral/suttacentral/issues/1310 (5-7 years ago)

- `@web/dev-server-legacy` is [used for](https://modern-web.dev/docs/dev-server/plugins/legacy/) testing on old browsers, but we don't currently test in CI and no developer runs an old browser, so this is not used
  - and per the lack of polyfill usage above, it doesn't seem like we have users of very old, long EoL browsers (e.g. IE11), so we don't need to test on them anyway
- this also remove an NPM warning from [a deprecated dependency](https://github.com/suttacentral/suttacentral/blob/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/client/package-lock.json#L3614) of `@web/dev-server-legacy`:
    ```
    npm WARN deprecated @babel/plugin-proposal-dynamic-import@7.18.6: This proposal has been merged to the ECMAScript standard and thus this plugin is no longer maintained. Please use @babel/plugin-transform-dynamic-import instead.
    ```
  
## References

- I suspected this was not necessary in a DM to Ven. Jhanarato back in April ([private link](https://discourse.suttacentral.net/chat/c/-/68/8281)), finally got around to double checking that 😅 

## Verification

- `npm test` still passes


jhanarato (approved):
Good to see more cruft removed. Looks good to my untrained eye.

ihongda:
Thanks @agilgur5

ihongda (approved):

PR #3691, opened 2026-07-23
refactor(test): standardize using `@open-wc/testing` package

agilgur5:
## Summary

Switch direct usage of `@esm-bundle/chai` et al to use of wrapper `@open-wc/testing`

## Details

- should use that wrapper instead of `@esm-bundle/chai` directly or `lit` directly
  - c.f. [its docs](https://github.com/open-wc/open-wc/blob/8a89d51ac5db27f09e3a006d2771af8ba7e492c3/docs/docs/testing/testing-package.md#testing-helpers)
  - (otherwise could use directly and no need for wrapper, but choose one and stick with it)

- while at it:
  - remove unused `LitElement` import
  - shorten `sinon` import
  - fix some formatting of imports:
    - external imports before internal ones
    - single quotes as per [`prettier` config](https://github.com/suttacentral/suttacentral/blob/ac30f9c7dc07da9bafd2bf67027c5cbd33c5ee81/client/package.json#L26)
    - space between braces

## Reference

#3572 and future PRs will automate the formatting
  
## Verification

- `npm test` still passes

### DIFF
--- a/client/package-lock.json
+++ b/client/package-lock.json
@@ -35,7 +35,6 @@
         "@babel/eslint-plugin": "7.23.5",
         "@babel/plugin-transform-runtime": "7.22.5",
         "@babel/preset-env": "7.23.6",
-        "@esm-bundle/chai": "4.3.4",
         "@open-wc/testing": "4.0.0",
         "@rollup/plugin-babel": "^6.0.3",
         "@web/dev-server-legacy": "2.1.0",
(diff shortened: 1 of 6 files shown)


jhanarato:
Seems like a reasonable thing to do. Am I right that these tests don't run in CI yet?

agilgur5:
> Am I right that these tests don't run in CI yet?

Indeed, that's been on [my to-do list](https://github.com/suttacentral/suttacentral/pull/3549#discussion_r2778027843) since I noticed too. There are some small complications in that they need a browser and a local version of the API to run properly (ideally test/fake data would be used for most tests though), so there's a bit of configuration I needed to check for that.

ihongda (approved):

PR #1583, opened 2020-02-28
Test new text page.

ihongda:
1. Rewritten sc-text-page-selector.js with litElement.
2. Add new component: sc-bottom-sheet.js
3. Add new component: sc-bilara-segmented-text.js
4. Modified sc-stepper.js styles.

PR #3714, opened 2026-08-10
refactor: use `reduxActions` instead of `this.actions` methods

agilgur5:
## Summary

In several files, delete `actions` method and replace it with existing code in `reduxActions`

## Details

The actions are static methods that don't depend on the instance, so can be shared across all components
  - the store is global too, so semantically, it doesn't make sense to have them as instance methods
  - it is also fairly common Redux pattern to have shared ["action creators"](https://redux.js.org/tutorials/fundamentals/part-7-standard-patterns#action-creators)
  
**refactor: move duplicate actions to `reduxActions`**

- `changeSuttaMetaText`, `changeSuttaPublicationInfo`, and `setStaticPagesToolbarDisplayState` are used in multiple components
  - so remove duplicate instance `actions` and move them to shared `reduxActions`
  
**clean: remove leftover unused actions**

- `TopSheetViews` has two `actions` that were created but never used: `toggleTextualInfo` and `downloadParagraphs`
  - they originate from https://github.com/suttacentral/suttacentral/commit/a67510135d0b81a37b0139309b0f87ce35203a94 (https://github.com/suttacentral/suttacentral/pull/1634#discussion_r3754125836), which created the component with unused functions
    - these seemed to have been copy+pasted from an older `SettingsMenu` component (which was removed in https://github.com/suttacentral/suttacentral/commit/68c4ebfd8e0b2cf410b4b3397dde74d3e8fb7ed7), but that one actually used these functions, while this component does not

- `TopSheetViews` also has an action that was replaced but never removed: `setReferenceDisplayType`
  - this was replaced by `setDisplayedReferences` in https://github.com/suttacentral/suttacentral/commit/f925cf3f7a2fb0a8d0bc6ea99a48f54bc197110e (https://github.com/suttacentral/suttacentral/pull/1983#discussion_r3754158674), but `setReferenceDisplayType` was not removed despite being no longer used

- `TextBilara` has an action that was created but not used as well: `chooseSegmentedSuttaTextView`
  - it was created in https://github.com/suttacentral/suttacentral/commit/f35516de7c97c48840fec01c30ed87c0d2060a6a (https://github.com/suttacentral/suttacentral/pull/1583#discussion_r3754135369), where it is not used
    
**refactor: rename setReferenceDisplayType to setDisplayedReferences**

- `setReferenceDisplayType` was the old name before there were multiple references, afaict, while `setDisplayedReferences` is the new name for the array type
  - it wasn't renamed everywhere though, so we ended up with a duplicate action that had two different names
  - remove the instance-local one in `TopSheetViews` and use the shared one in `reduxActions`
    - the instance-local one was added in https://github.com/suttacentral/suttacentral/commit/f925cf3f7a2fb0a8d0bc6ea99a48f54bc197110e, while the global one was added in https://github.com/suttacentral/suttacentral/commit/b2ac9b2959aaff8c6d9a5e27cf2365cd90d8b70f
    
## Notes to Reviewers

- If you search `actions()`, there's a few more instance methods left. I left those for now as the actions either operate on a single component or one other component, and so perhaps should be instance-local, [plus stored with `localStorage`](https://github.com/suttacentral/suttacentral/blob/b2cb0f91eed45e42c234e39ce17cdb87a2965167/client/redux-store.js#L304)
  - Some existing global actions should perhaps be converted to instance-local storage too
- #3713 cleans up some other unused actions

### DIFF
--- a/client/elements/addons/sc-redux-actions.js
+++ b/client/elements/addons/sc-redux-actions.js
@@ -8,7 +8,7 @@ export class reduxActions {
     });
   }
 
-  static setReferenceDisplayType(displayedReferences) {
+  static setDisplayedReferences(displayedReferences) {
     store.dispatch({
       type: 'SET_REFERENCE_DISPLAY_TYPE_ARRAY',
       displayedReferences,
--- a/client/elements/addons/sc-redux-actions.js
+++ b/client/elements/addons/sc-redux-actions.js
@@ -235,4 +235,25 @@ export class reduxActions {
       showIllustrations,
     });
   }
+
+  static changeSuttaMetaText(metaText) {
+    store.dispatch({
+      type: 'CHANGE_SUTTA_META_TEXT',
+      metaText,
+    });
+  }
+
+  static changeSuttaPublicationInfo(publicationInfo) {
+    store.dispatch({
+      type: 'CHANGE_SUTTA_PUBLICATION_INFO',
+      suttaPublicationInfo: publicationInfo,
+    });
+  }
+
+  static setStaticPagesToolbarDisplayState(toolbarDisplayState) {
+    store.dispatch({
+      type: 'CHANGE_STATIC_PAGES_TOOLBAR_DISPLAY_STATE',
+      staticPagesToolbarDisplayState: toolbarDisplayState,
+    });
+  }
 }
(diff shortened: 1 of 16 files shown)


ihongda:
Thanks @agilgur5 This is a problem I've been wanting to address.

ihongda (approved):

agilgur5:
> Thanks @agilgur5 This is a problem I've been wanting to address.

Yea it looked like it had been halfway done, so this gets it to 90% done with the exception of the ones in the "Notes to Reviewers" section that perhaps need to be refactored differently

PR #2309, opened 2021-10-23
Locale templates

blake-sc:

PR #3716, opened 2026-08-11
clean: remove unused imports in localization scripts

agilgur5:
## Summary

Fix some lint issues in Python scripts in `client/localization/`
- these were incidental findings from when I accidentally ran `ruff` (#3656) in the top-level directory instead of in `server/`

## Details

- the linter found some unused imports, which originate from when these scripts were created in c6e7a08fda1ea3e3aed859c5b4bb30b67de66c48 (https://github.com/suttacentral/suttacentral/pull/2309#pullrequestreview-4909251928)
  - and I fixed some whitespace issues while I was at it as well (trailing whitespace mostly)

### DIFF
--- a/client/localization/common.py
+++ b/client/localization/common.py
@@ -1,5 +1,4 @@
 import re
-import logging
 from pathlib import Path
 
 static_elements_dir = Path('../elements/static/')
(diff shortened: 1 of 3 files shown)


ihongda:
Thanks for your hard working! @agilgur5

PR #3729, opened 2026-08-23
fix: improve metadata of Edition pages

agilgur5:
## Summary

Ensure `Edition` and `EditionMatter` pages have a unique title and description

#### fix: `EditionMatter` pages should have proper metadata

- per my own [forum bug report](https://discourse.suttacentral.net/t/suttacentral-bug-reports/27763/340?u=agilgur5), these pages had a title of `publicationEditionMatter—SuttaCentral` and the site-wide default `meta` `description`
  - this generic metadata originates from [the `PageSelector` component](https://github.com/suttacentral/suttacentral/blob/585f9dea6258d43325b20da3a34d52e62586f262/client/elements/sc-page-selector.js#L608)
  - they should have their own metadata, similar to the fix for a prior bug report: https://github.com/suttacentral/suttacentral/commit/a685b5d3e6959016ade235203ba70536720f5951 (#3492)
    - this is nearly the same metadata as the `Edition` page, but with ` - ${this.matter}` following it to differentiate and define the page

- use `#updateMetaData` to be properly private and more semantically correct than `create`, [following `SuttaplexList`](https://github.com/suttacentral/suttacentral/blob/585f9dea6258d43325b20da3a34d52e62586f262/client/elements/suttaplex/sc-suttaplex-list.js#L189) and a recent PR of mine (#3718)
      
#### fix: use the Edition `blurb` for `description` metadata

- Editions have blurbs, so that can be used for the metadata to give a unique description
  - instead of duplicating the title into the description

- also use `#updateMetaData` instead of `_createMetaData` per previous commit and my other PR

#### fix: use the `EditionMatter` `h1` for `description` metadata

- pull out the text of the `h1` (that [each Edition front matter file has](https://github.com/suttacentral/bilara-data/blob/548765d7eb515e3dc47eb92f3f0057ef3c0456b4/_publication/en/brahmali/editions/vinaya/pli-tv-vi/matter/foreword.html#L2)) to use as a unique description
  - instead of duplicating the title in the description
  
## Verification

Tested locally and made sure the `title` and `meta` `description` updated properly for the Vinaya main page and front matter pages

## Future Work
- [ ] `${this.matter}` is an abbreviation/path/ID, so not ideal, but a good first start IMO
- [ ] the `EditionMatter` description could be longer, potentially pulling from the first `p` instead, but that is a lot less consistent than the heading
  - for now, I think this the approach in this PR is a good bit better than before

### DIFF
--- a/client/elements/publication/sc-publication-edition-matter.js
+++ b/client/elements/publication/sc-publication-edition-matter.js
@@ -132,10 +132,12 @@ export class SCPublicationEditionMatter extends LitLocalized(LitElement) {
   }
 
   #updateMetaData() {
-    const description = `${this.editionDetail?.[0].root_name} — ${this.editionInfo?.publication?.creator_name} - ${this.matter}`;
+    const title = `${this.editionDetail?.[0].root_name} — ${this.editionInfo?.publication?.creator_name} - ${this.matter}`;
+    const doc = new DOMParser().parseFromString(this.matterContent, 'text/html');
+    const description = doc.querySelector('h1').textContent;
     dispatchCustomEvent(document, 'metadata', {
-      pageTitle: description,
-      title: description,
+      pageTitle: title,
+      title,
       description,
     });
   }
--- a/client/elements/publication/sc-publication-edition-matter.js
+++ b/client/elements/publication/sc-publication-edition-matter.js
@@ -8,6 +8,7 @@ import { SCPublicationStyles } from '../styles/sc-publication-styles';
 import { typographyCommonStyles } from '../styles/sc-typography-common-styles';
 import { typographyStaticStyles } from '../styles/sc-typography-static-styles';
 import { store } from '../../redux-store';
+import { dispatchCustomEvent } from '../../utils/customEvent';
 import { allEditions } from './sc-publication-common';
 
 export class SCPublicationEditionMatter extends LitLocalized(LitElement) {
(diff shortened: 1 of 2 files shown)


agilgur5:
> this merge conflicts with #3711 , so is a draft PR for now

Now that that PR has been merged, I have rebased this on top and fixed the merge conflicts. It is ready for review now

client/elements/publication/sc-publication-edition-matter.js
```diff
@@ -130,6 +131,17 @@ export class SCPublicationEditionMatter extends LitLocalized(LitElement) {
     }
   }
 
+  #updateMetaData() {
+    const title = `${this.editionDetail?.[0].root_name} — ${this.editionInfo?.publication?.creator_name} - ${this.matter}`;
+    const doc = new DOMParser().parseFromString(this.matterContent, 'text/html');
```
agilgur5, line 136: Ack, there's a bug on this line, as discovered by Ven. @sabbamitta [on the forum](https://discourse.suttacentral.net/t/suttacentral-bug-reports/27763/342?u=agilgur5).

I had an _implicit_ merge conflict with another PR of mine #3725 on this line and didn't handle it [in my rebase](https://github.com/suttacentral/suttacentral/compare/2e555f41a816563ea0eb29591bac0c78118c44e3..fb547d5127e0e6bd250eac15359086c717b3a6cd#diff-b2e9da80cc0919b78735d9018e916c6ee179c35d2885b9712eaeee49adf5e928) (I only handled the explicit ones). I was perhaps being a little too clever in separating all these as independent PRs and should've stacked some instead 😕 

My bad, will have a fix out shortly:
```suggestion
    const doc = new DOMParser().parseFromString(matterContent, 'text/html');
```

(and pass `matterContent` into the function etc)

agilgur5: Fixed in https://github.com/suttacentral/suttacentral/pull/3748

PR #3730, opened 2026-08-24
fix: do not generate `<span>` tags when the variant is blank.

ihongda:
This issue was caused by the recent addition of aligned segment IDs to some JSON files.


agilgur5:
For reference, here's the corresponding [forum bug report](https://discourse.suttacentral.net/t/suttas-sprinkled-with-empty-notes-on-variant/45040?u=agilgur5) and one of the commits of [blank segments in `bilara-data`](https://github.com/suttacentral/bilara-data/commit/fb224bd6d5323d75585c59833a000ad25a34f547) that is causing the regression

PR #3741, opened 2026-09-09
refactor: use `this` consistently in `TopSheetViews`

agilgur5:
## Summary

Ensure `this` is used consistently instead of a self-referencing `querySelector`

## Details

- `_onPaliScriptChanged` ran a `querySelector` on itself and then used a variable instead of `this`
  - everything else uses `this`, so this function was the only mismatch
  - it also uses would-be private functions that have an `_`; if those were converted to the newer private syntax (`#`), they wouldn't work outside of `this`

  - Same as #3739, this originates from e42d0f978062a9f01193ccfcfb0fbb4996465089 (#1878)
    - I'm guessing this may have been due to a timing issue with native HTML state, such as with `select`s, as described in [this SO answer](https://stackoverflow.com/a/55888908/3431180) and [this old Lit issue](https://github.com/lit/lit-element/issues/805#issuecomment-605554546), but `this` works fine for me when testing
    
## Verification

Re-tested the dropdown locally and it updated the script correctly every time I changed it


ihongda:
Thanks @agilgur5

PR #3742, opened 2026-09-09
fix(styles): don't show empty translations when side notes enabled

agilgur5:
## Summary

Remove chunks of empty vertical space when there are omitted translations in a no Pali + side notes view

## Details

- this is a follow-up to ef5cd93 (#3687), where I missed the fact that different stylesheets sometimes apply on this page
  - no Pali = `plainStyles`, while [no Pali + side notes = `plainPlusStyles`](https://github.com/suttacentral/suttacentral/blob/59e3cef3a87644e9f162c815d29d2a35461d04cc/client/elements/text/sc-text-bilara.js#L101)
  - as a result, I missed the case of translation only + side notes (partly because I usually read with asterisk tooltip notes)

- note that the CSS is a bit more computationally expensive here as I had to apply it to `.segment`, as changing any of the empty children didn't fix the empty block
  - I believe this is partly due to the use of `display: grid` (while no notes uses `display: block`)
  - in the side notes case, it also has to apply to more than just verse
  
## Verification
Tested locally, screenshots:
<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
Screenshot 2026-09-09 at 5 54 34 PM
    </td>
    <td>
Screenshot 2026-09-09 at 5 53 56 PM
    </td>
  </tr>
</table>

I did also double-check that views with Pali worked fine, but they use a different stylesheet anyway


ihongda:
Thanks @agilgur5

PR #3745, opened 2026-09-10
refactor: reduce nesting by inverting `if` statements

agilgur5:
## Summary
- less nesting = less [code complexity](https://www.ibm.com/think/topics/code-complexity) (cognitive complexity, to be specific)
- instead of writing an `if` statement followed by a large block of code, invert the conditional, return early, and then proceed linearly with the rest of the code

## Process details
- I went through a majority of the conditionals in the codebase, particularly those in the front-end, but not quite all of them.
  - I mostly targeted larger blocks and blocks with simple conditionals (where I wouldn't have to think too hard about how to invert them)
    - so skipped some smaller blocks, most one-liners, and complex conditionals
  - I also skipped some conditionals where there was an existing style to follow, for instance, in `stateChanged` and `updated` methods that follow a similar pattern in most components

## Notes to Reviewers
- the [diff](https://github.com/suttacentral/suttacentral/pull/3745/changes?w=1) is substantially easier to read if you [toggle "hide whitespace"](https://docs.github.com/en/pull-requests/reference/branches#comparing-branches-in-pull-requests)
- Some linters (#3656, #3572) have support for automatically checking complexity, though usually in a blunt check of "more than X amount of nesting", rather than a "is this nesting necessary?" (although some can do both)

## Future Work
- [ ] it might make sense to add a [`.git-blame-ignore-revs`](https://docs.github.com/en/repositories/working-with-files/using-files/viewing-and-understanding-files#ignore-commits-in-the-blame-view) for this commit after it is merged, as this is a purely stylistic change


ihongda:
Thanks @agilgur5 
After this refactoring, the code is more readable. 👍🏼

client/elements/addons/sc-top-sheet-publication-bilara.js
```diff
@@ -160,30 +160,31 @@ export class SCTopSheetPublicationBilara extends SCTopSheetCommon {
...
-      this.editionNumber = this.getEditionAttribute(data, 'edition_number');
-      this.publicationDate = this.getEditionAttribute(data, 'publication_date');
-      this.publisher = this.getEditionAttribute(data, 'publisher');
-      this.editionURL = this.getEditionAttribute(data, 'edition_url');
-      this.publicationType = this.getEditionAttribute(data, 'publication_type');
-      this.numberOfVolumes = data.number_of_volumes;
-      this.licenseType = this.getLicenseAttribute(data, 'license_type');
-      this.licenseAbbreviation = this.getLicenseAttribute(data, 'license_abbreviation');
-      this.licenseStatement = this.getLicenseAttribute(data, 'license_statement');
+    if (!data?.is_published === true) {
```
agilgur5, line 163: I wrote this a little weirdly (this was a simplification to optional chaining + an invert), could be more straightforward as:


```suggestion
    if (data?.is_published !== true) {
```

client/elements/sc-page-selector.js
```diff
@@ -380,62 +380,69 @@ export class SCPageSelector extends LitLocalized(LitElement) {
...
+      const contextToolbar = scSiteLayout?.querySelector('#context_toolbar');
+      const newScActionItems = document.createElement('sc-action-items');
+      newScActionItems.id = 'action_items';
+      contextToolbar.appendChild(newScActionItems);
+      this._setActionItemsDisplayState();
+    })
+    .catch(err => {
+      console.error(err);
+    });
+}
```
agilgur5, line 407: woops, this closing brace got unindented accidentally

PR #3746, opened 2026-09-10
refactor: simplify logic for showing toasts for references

agilgur5:
## Summary

Simplify a bit of toast logic for references in `TopSheetViews`'

## Details

- the only difference is the text shown, so the logic can be simplified a bit and readability improved
  - this could be further simplified into a single conditional, but I think that would hamper readability a bit

- also fix style of missing space after `if`: `if(` -> `if (`
  - consistent with the rest of the codebase


ihongda:
Thanks @agilgur5

PR #3747, opened 2026-09-10
fix: show fallback lang when interface translation is blank

agilgur5:
## Summary

Don't show raw localization keys when an interface translation has an empty string

## Details

- as [reported on the forum](https://discourse.suttacentral.net/t/partially-translated-pages-show-raw-localization-keys/45242?u=agilgur5) by Ven. @wcang, the interface started showing raw localization keys when it was partially translated
  - similar to 41573627fbdb949daafde9112a73d07406c4ff3a (#3730) and #3743, this has the same root cause of blank segments added to `bilara-data` like https://github.com/suttacentral/bilara-data/commit/fb224bd6d5323d75585c59833a000ad25a34f547
- this fix restores the previous behavior before the addition of blank / empty string segments to `bilara-data`, that interface translations would fallback to English for any blank/untranslated segments

## Verification

Tested against [the Vinaya Guide in Chinese](https://suttacentral.net/vinaya-guide-brahmali?lang=zh), as per the bug report, which was only [partially translated](https://github.com/suttacentral/bilara-data/blob/11c9d708978cde8ba61096d8a75f7ddfb846f639/translation/zh/site/vinaya_translation-zh-site.json#L24). The [interface translation](https://github.com/suttacentral/bilara-data/blob/11c9d708978cde8ba61096d8a75f7ddfb846f639/translation/zh/site/interface_translation-zh-site.json#L157) is also only partly complete, as visible in the screenshots below:

<table>
  <tr>
    <th>Before</th>
    <th>After</th>
  </tr>
  <tr>
    <td>
Screenshot 2026-09-10 at 7 59 12 PM
    </td>
    <td>
Screenshot 2026-09-10 at 7 59 38 PM
    </td>
  </tr>
</table>


wcang:
Thank you, @agilgur5. You're the best!

ihongda:
Thanks @agilgur5

